PR #88, opened 2020-09-23
Cycle Detection in Welldefinedness of Constants

### LINKED ISSUE #44: Cyclic definition of constants
> Created by bitbucket user **Wytse Oortwijn** on 2020-09-10 12:45

The following simple Go\(bra\) program results in an `java.lang.IllegalStateException: Cycle detected in attribute evaluation 'store' at LOOP` stack trace:

```go
package pkg
const CYCLE = CYCLE
```

‌

ArquintL:
Fixes #44


src/main/scala/viper/gobra/frontend/info/implementation/typing/IdTyping.scala
```diff
@@ -28,12 +29,26 @@ trait IdTyping extends BaseTyping { this: TypeInfoImpl =>
     }
   }
 
-  private[typing] def wellDefActualRegular(entity: ActualRegular, id: PIdnNode): ValidityMessages = entity match {
+  /**
+    * Returns true if n or any of its children (in the current package) have an entity that corresponds to one in the set e
```
Felalolf, line 33: "any of its children ... has an entity that is contained in the set e"?

src/main/scala/viper/gobra/frontend/info/implementation/typing/IdTyping.scala
```diff
@@ -28,12 +29,26 @@ trait IdTyping extends BaseTyping { this: TypeInfoImpl =>
...
+    */
+  private def isNotCyclic(n: PNode, e: Set[Entity]): ValidityMessages = LocalMessages(message(n, s"got cyclic structure starting at $n", n match {
+    case n: PIdnNode => entity(n) match {
+      case r if e.contains(r) => true
+      case SingleConstant(_, _, _, _, _, ctx) if this != ctx => false // we do not follow the evaluation into different packages
```
Felalolf, line 38: Why do we not follow the evaluation into another package? I guess that it is not necessary because the other package cannot point back into the current package because imports are not allowed to have a cycle.

Felalolf (approved):
You should fix the one function description and then it is fine.

src/main/scala/viper/gobra/frontend/info/implementation/typing/IdTyping.scala
ArquintL: yes exactly, I've extended the comment

Issue #91, opened 2020-09-30
Support for Go's integer types


jcp19:
At the moment, Gobra is only capable of handling integers of the `int` type. There is no support for `int8`, `int16`, `int32` and `int64` and its unsigned variants. Besides, Gobra assumes that integers are unbounded instead of being limited by the size of its type.

jcp19:
Added support for all integer types at the Gobra language level and type checker. I still have not propagated these changes to the internal language and encoding but I will do that in a future commit

PR #94, opened 2020-10-01
Add int types of different sizes and its unsigned variables at the Gobra language level

jcp19:
This PR tackles the issue #91. It extends Gobra's language to support the types `int`, `int8`, `int16`, `int32`, `int64` and its synonyms and unsigned counterparts as well as as the type `uintptr`. This required extensive changes to the type-checker and forced me to introduce a type `IntT(UntypedConst)` which allows us to reason about untyped constants at the type-checker level. 

Besides, I extended the checker for well defined expressions to disallow assignments whose RHS is a constant expression that leads to an overflow such as
```go
var x int8 = 128 // overflow --> the max value of int8 is 127
```
(these kinds of assignments are also rejected by the Go compiler)

I updated all the tests and files that relied on the single `IntT` type and added new tests for the features I introduced (if you believe that there is an interesting test case that I may have missed, please let me know 😉).

Please note that I did not make any changes in the internal language. I am hoping to do that in a separate PR.


src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -873,7 +873,21 @@ object Parser {
...
+        // signed integer types
+        "rune" ^^^ PInt32Type() |
+        "int8" ^^^ PInt8Type() |
+        "int16" ^^^ PInt16Type() |
+        "int32" ^^^ PInt32Type() |
+        "int64" ^^^ PInt64Type() |
+        "int" ^^^ PIntType() | // 'int' must come after all 'intX' parsers, otherwise parsing will stop at 'int'
+        // unsigned integer types
+        "byte" ^^^ PUInt8Type() |
+        "uint8" ^^^ PUInt8Type() |
```
Felalolf, line 885: Since both are parsed as the same type, the type checker is not able to distinguish between byte and uint8. This is only ok, if these types are indistinguishable in their behaviour. Is this the case here? Similarly, for unes and int32, etc.

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
```diff
@@ -159,6 +166,41 @@ trait Assignability extends BaseProperty { this: TypeInfoImpl =>
...
+        case UntypedConst => true
+        case x =>
+          val bounds = getBoundsOfIntKind(x)
+          bounds.lower <= value && value <= bounds.upper
+      }
+
+      case _ => ???
+    }
+
+  private val getBoundsOfIntKind: IntegerKind => Bounds = {
```
Felalolf, line 189: The bounds seem a bit redundant since every kind X is mapped to Bounds.X.

src/main/scala/viper/gobra/ast/frontend/PrettyPrinter.scala
```diff
@@ -437,6 +437,16 @@ class DefaultPrettyPrinter extends PrettyPrinter with kiama.output.PrettyPrinter
     case PNamedOperand(id) => showId(id)
     case PBoolType() => "bool"
     case PIntType() => "int"
+    case PInt8Type() => "int8"
+    case PInt16Type() => "int16"
+    case PInt32Type() => "int32"
```
Felalolf, line 442: For example here, the int32 could have been a rune, as well.

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -326,11 +326,13 @@ object PLiteral {
   )
 }
 
+sealed trait PNumExpression
```
Felalolf, line 329: Please, add a description.

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -423,15 +425,15 @@ case class PGreater(left: PExpression, right: PExpression) extends PBinaryExp
 
 case class PAtLeast(left: PExpression, right: PExpression) extends PBinaryExp
 
-case class PAdd(left: PExpression, right: PExpression) extends PBinaryExp
+case class PAdd(left: PExpression, right: PExpression) extends PBinaryExp with PNumExpression
```
Felalolf, line 428: I just want to point out that we have statements with imlicit arithmetic operation, e.g. x += 1

src/main/scala/viper/gobra/frontend/info/implementation/property/TypeMerging.scala
```diff
@@ -18,7 +18,17 @@ trait TypeMerging extends BaseProperty { this: TypeInfoImpl =>
     // currently, only merging of l and r being identical types is supported
     // possible future improvement: if l is assignable to r, then return r instead of None (or vice versa)
     (l, r) match {
-      case (Single(lst), Single(rst)) => if (identicalTypes(lst, rst)) Some(lst) else None
+      case (Single(lst), Single(rst)) =>
+        if (identicalTypes(lst, rst)) Some(lst) else {
+          (lst, rst) match {
+            case (a, IntT(UntypedConst)) if underlyingType(a).isInstanceOf[IntT] => Some(a)
+            case (IntT(UntypedConst), b) if underlyingType(b).isInstanceOf[IntT] => Some(b)
+            case (SequenceT(l), SequenceT(r)) => for {
```
Felalolf, line 26: What is with the other data structure kinds, e.g. sets, multisets, array, structs, ...

src/main/scala/viper/gobra/util/Bounds.scala
```diff
@@ -0,0 +1,20 @@
...
+  val Int64: Bounds = Bounds(-9223372036854775808L, 9223372036854775807L)
+
+  val UInt: Bounds = Bounds(0, 4294967295L) // TODO: allow for changes according to the int size of the machine arch (either 32 or 64 bit)
+  val UInt8: Bounds = Bounds(0, 255)
+  val UInt16: Bounds = Bounds(0, 65535)
+  val UInt32: Bounds = Bounds(0, 4294967295L)
+  val UInt64: Bounds = Bounds(0, 4294967295L * 4294967295L) // Cannot represent the upper bound as a literal
+  val UIntPtr: Bounds = Bounds(0, 4294967295L + 4294967295L) //TODO: change according to spec
+  // val UInt64: Bounds = Bounds(0, 18446744073709551615L)
+  // val UIntPtr: Bounds = Bounds(0, 18446744073709551615L)
```
Felalolf, line 19: You can write the bounds, when you wrap them into a bigint https://stackoverflow.com/questions/59353423/big-int-in-scala

src/main/scala/viper/gobra/util/Bounds.scala
```diff
@@ -0,0 +1,20 @@
+package viper.gobra.util
+
+case class Bounds(lower: BigInt, upper: BigInt)
+
+case object Bounds {
+  val Int: Bounds = Bounds(-2147483648, 2147483647) // TODO: allow for changes according to the int size of the machine arch (either 32 or 64 bit)
```
Felalolf, line 6: write an issue for this, so that it is not forgotten.

src/test/resources/regressions/features/int_sizes/int-assign-fail1.gobra
```diff
@@ -0,0 +1,24 @@
+package ints
```
Felalolf, line 1: Add licenses to all new files.

src/test/scala/viper/gobra/ExprTypingUnitTests.scala
```diff
@@ -18,7 +18,10 @@ class ExprTypingUnitTests extends FunSuite with Matchers with Inside {
 
   test("TypeChecker: should classify an integer literal as integer") {
     frontend.exprType(PIntLit(42))() should matchPattern {
-      case Type.IntT =>
+      // the type should be Type.Int instead of Type.UntypedConst due to the implementation of
+      // front.end(exprType) which in practice evaluates the type of the literal in a
```
Felalolf, line 22: is the function called "front.end". This seems like a spelling error.

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
```diff
@@ -43,16 +44,22 @@ trait Assignability extends BaseProperty { this: TypeInfoImpl =>
     case (Single(lst), Single(rst)) => (lst, rst) match {
 
         // for go's types according to go's specification (mostly)
+      case (IntT(UntypedConst), r) if isIntegerType(underlyingType(r)) => true
+      case (l, IntT(UntypedConst)) if isIntegerType(underlyingType(l)) => true // not in spec
```
Felalolf, line 48: The comment is confusing. "Not part of Go spec, but necessary for the definition of comparability".

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
```diff
@@ -159,6 +166,41 @@ trait Assignability extends BaseProperty { this: TypeInfoImpl =>
...
+  private def intValInBounds(value: BigInt, typ: Type): Boolean =
+    underlyingType(typ) match {
+      case IntT(t) => t match {
+        case UntypedConst => true
+        case x =>
+          val bounds = getBoundsOfIntKind(x)
+          bounds.lower <= value && value <= bounds.upper
+      }
+
+      case _ => ???
```
Felalolf, line 186: This should throw an error, right? I would use ??? only if something is supposed to be there eventually.

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
Felalolf: You could add the bounds to the type kind already.

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
```diff
@@ -43,16 +44,22 @@ trait Assignability extends BaseProperty { this: TypeInfoImpl =>
     case (Single(lst), Single(rst)) => (lst, rst) match {
 
         // for go's types according to go's specification (mostly)
+      case (IntT(UntypedConst), r) if isIntegerType(underlyingType(r)) => true
+      case (l, IntT(UntypedConst)) if isIntegerType(underlyingType(l)) => true // not in spec
       case (l, r) if identicalTypes(l, r) => true
-      case (l, r) if !(l.isInstanceOf[DeclaredT] && r.isInstanceOf[DeclaredT])
-        && identicalTypes(underlyingType(l), underlyingType(r)) => true
+      // This rule seems to be ignored in the Go compiler
```
Felalolf, line 50: What do you mean by that?

src/main/scala/viper/gobra/frontend/info/implementation/typing/TypeTyping.scala
```diff
@@ -101,6 +111,11 @@ trait TypeTyping extends BaseTyping { this: TypeInfoImpl =>
     case t: PType => typeType(t)
   }
 
+  lazy val isIntegerType: Type => Boolean = {
+    case IntT(_) => true
+    case _ => false
+  }
+
```
Felalolf, line 118: This is the same as just an instanceOf[IntT]

Felalolf: Also, it should be function and not a lazy field.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
```diff
@@ -141,12 +141,20 @@ trait GhostExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
-        case PRangeSequence(_, _) => SequenceT(IntT)
-        case PSequenceAppend(left, _) => exprType(left)
+        case PRangeSequence(_, _) => SequenceT(IntT(UntypedConst))
+
+        case PSequenceAppend(left, right) =>
+          val lType = exprType(left)
+          val rType = exprType(right)
+          typeMerge(lType, rType) match {
+            case Some(seq@SequenceT(_)) => seq
+            case _ => lType
```
Felalolf, line 156: In this case, you should maybe throw a violation.

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
```diff
@@ -43,14 +49,17 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
...
 
     case n@PAssignmentWithOp(right, op, left) =>
       isExpr(right).out ++ isExpr(left).out ++
         assignable.errors(left)(n) ++ compatibleWithAssOp.errors(exprType(left), op)(n) ++
-        assignableTo.errors(exprType(right), exprType(left))(n)
+        assignableTo.errors(exprType(right), exprType(left))(n) ++
+        assignWithinBounds.errors(exprType(left), right)(n)
 
     case n@PShortVarDecl(rights, lefts, _) =>
+      // TODO: add overflow checks
```
Felalolf, line 62: Could it be better to add the overflow check at the arithmetic operations. For example, when you have an addition with literals and constants, then you check whether there is an overflow. That way, you do not have to do it for all assignments.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -312,7 +314,12 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
          _: PLess | _: PAtMost | _: PGreater | _: PAtLeast =>
       BooleanT
 
-    case _: PAdd | _: PSub | _: PMul | _: PMod | _: PDiv | _: PLength => IntT
+    case _: PIntLit | _: PAdd | _: PSub | _: PMul | _: PMod | _: PDiv =>
+      val exprNum = expr.asInstanceOf[PNumExpression]
```
Felalolf, line 318: Can it not infer that on its own, i.e. did you try to remove the as instance of?

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -321,14 +328,42 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
     case e => violation(s"unexpected expression $e")
   }
 
+  private def getDefaultType(expr: PNumExpression): Option[Type] = {
```
Felalolf, line 331: Please, add a method describtion.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -321,14 +328,42 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
+    }
+
+    val parent = parents(0)
+
+    parent match {
+      case PShortVarDecl(_, _, _) =>
+          Some(IntT(Int))
+
+      case _ => None
+    }
```
Felalolf, line 347:
```
expr match {
  case tree.parent(p) => p match { 
    case _: PShortVarDecl => Some(IntT(Int))
    case _ => None
  }
}
```

Felalolf:
You should also update the PR with the newest version of master.

src/main/scala/viper/gobra/frontend/Parser.scala
jcp19: Yes, the Go spec defines byte and rune as aliases to uint8 and int32 respectively (https://golang.org/ref/spec#Numeric_types). As an additional check, I verified that
```go
func main() {
	var a byte = uint8(1)
	var b uint8 = byte(2)
	fmt.Printf("%d %d\n", a, b)

	var c int32 = rune(1)
	var d rune = int32(1)
	fmt.Printf("%d %d\n", c, d)
}
```
compiles without type errors and warnings.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
jcp19: The reason why I did not throw a violation is that Gobra would terminate with an exception whereas if I let it proceed, the well-defidedness checker would detect the error and send a message and, this way, it would provide useful information to the user instead of seeing a stack trace. Would it be better if I returned `UnknownType` instead of   `lType`?

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
jcp19: I do not think that that would be better because:
1. you would need to add more checks for the type-checker to detect context-defined types (e.g. `var x int8 = 127 + 1` would require you to know when you are type-checking the expression `127 + 1` that it is of type `int8`. In principle, you would need to add cases for all kinds of assignments anyway.
2. if you only checked arithmetic operations, how would you check that a literal assignment overflows? e.g.
```go
var x int8 = 128 // this should overflow and the compiler treats it as an error
```

src/main/scala/viper/gobra/ast/frontend/Ast.scala
jcp19: It seems to me that this does not affect how we type-check integer expressions. Do you see any problem that this may pose?

src/main/scala/viper/gobra/util/Bounds.scala
jcp19: This was a stupid mistake by me that should not have been committed in the first place, I left here as a placed older and forgot to change. Thank you for the stack overflow question though :)

jcp19: Changed in commit 8cb139e

src/test/resources/regressions/features/int_sizes/int-assign-fail1.gobra
jcp19: I added licenses to all new test files in commit 8cb139e. I did not add a license to the Bounds file because I'm planning to remove it in another commit before the PR gets merged.

src/test/scala/viper/gobra/ExprTypingUnitTests.scala
jcp19: Fixed in 8cb139e

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
jcp19: Fixed in 8cb139e

src/main/scala/viper/gobra/frontend/info/implementation/typing/TypeTyping.scala
jcp19: removed `isIntegerType` in 8cb139e

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
jcp19: Changed in 8cb139e

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
jcp19: No, it leads to a Type Mismatch because expr is of type `PActualExpression`.

jcp19: Replaced this part of code by
```scala
case exprNum: PNumExpression =>
      val typ = intExprType(exprNum)
      if (typ == IntT(UntypedConst)) getDefaultType(exprNum).getOrElse(typ) else typ
```

src/main/scala/viper/gobra/ast/frontend/Ast.scala
jcp19: Solved in fec51ca

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
jcp19: Yes, you are correct

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
jcp19: Removed Bounds.scala file altogether in bceb60f

src/main/scala/viper/gobra/ast/frontend/PrettyPrinter.scala
jcp19: Fixed in cd4d443

src/main/scala/viper/gobra/util/Bounds.scala
jcp19: Issue #95

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
jcp19: Fixed in bceb60f

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
jcp19: The Go language spec states that
```
A value x is assignable to a variable of type T ("x is assignable to T") if one of the following conditions applies:
...
- x's type V and T have identical underlying types and at least one of V or T is not a defined type.
...
```

It is my understanding that, given such rule, the go compiler would accept something like
```go
package main

type IntType int

func main() {
	var x int = 1
	var y IntType = x
}
```
However, the compiler rejects this program with the message
```
./test.go:7:6: cannot use x (type int) as type IntType in assignment
```

src/main/scala/viper/gobra/frontend/info/implementation/property/TypeMerging.scala
jcp19: I'm not sure if it makes sense to define this function for structs given that they are statically defined and any struct literal must have the fields of the declared type. I think that it makes sense for all other types that you mentioned.

jcp19: Check 84706b9

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
jcp19: Added in 203d2a2

src/main/scala/viper/gobra/ast/frontend/Ast.scala
Felalolf: I just wanted to point it out. Since constants cannot be reassigned, it probably does not matter.

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
Felalolf: Then please make the comment more explicit. "Albeit stated in the language specification, the Go compiler does not accept assignments where ...."

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
Felalolf: Usually, the pattern is that the well-def functions check if a node can be typed and the typing function then returns the type. In partcular, the typing function is only evaluated if the well-def function returned that the function is safe to type. Following that pattern, the typing functions should throw an exception if something problematic occurs, since the well-def function should have checked and handled this,

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
Felalolf: If you have a function that gives you the type of an expressions, then you can check literals and arightmetic operations by checking that their result (if constant) is within the bounds of the expected type. Checking at assignments seems a bit problematic, since there can be other places where additions happen, for example you can have something like `v := foo(127 + 1)` where `foo` expects an int8.

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
jcp19: Fixed in b1738cf

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
jcp19: You are right, we cannot detect that case in the current state

jcp19:
> You should also update the PR with the newest version of master.

Done in fad74cb

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
jcp19: Solved in 76be76a

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
Felalolf: Do you intend to fix this in this PR or in another PR?

Felalolf: If not, then please create an issue for this.

jcp19: I was just writing a fix for this :) Please take a look at this commit 5f400d7

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -235,8 +239,14 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
           case (_: PEquals | _: PUnequals, l, r) => comparableTypes.errors(l, r)(n)
           case (_: PAnd | _: POr, l, r) => assignableTo.errors(l, AssertionT)(n) ++ assignableTo.errors(r, AssertionT)(n)
-          case (_: PLess | _: PAtMost | _: PGreater | _: PAtLeast | _: PAdd | _: PSub | _: PMul | _: PMod | _: PDiv, l, r) =>
+          case (_: PLess | _: PAtMost | _: PGreater | _: PAtLeast, l, r) =>
             assignableTo.errors(l, IntT(UntypedConst))(n) ++ assignableTo.errors(r, IntT(UntypedConst))(n)
+          case (_: PAdd | _: PSub | _: PMul | _: PMod | _: PDiv, l, r) =>
+            assignableTo.errors(l, IntT(UntypedConst))(n) ++ assignableTo.errors(r, IntT(UntypedConst))(n) ++ {
+              val constVal = intConstantEval(n)
+              val typCtx = getTypeFromContext(n.asInstanceOf[PNumExpression])
+              if (constVal.isDefined && typCtx.isDefined) assignableWithinBounds.errors(typCtx.get, n)(n) else noMessages
```
Felalolf, line 248: A minor thing. You should first check wether constVal is defined and then comput the expected type (to avoid unnecessary computations).

Felalolf: (Just as scala practice, do whathever you prefer) You could write the following:
```
val res = for {
  constVal <- intConstantEval(n)
  typeCtx <- getTypeFromContext(n.asInstanceOf[PNumExpression])
} yield assignableWithinBounds.errors(typCtx.get, n)(n)
res.getOrElse(noMessages)
```

Felalolf (approved):
I wrote one potential improvement. Do that and then it is fine.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
jcp19: I realised that I don't even need to compute the value of the expression here, as this is made in `assignableWithinBounds`. As such, I could probably compress the code even further by writing
```scala
getTypeFromContext(n.asInstanceOf[PNumExpression]).map(assignableWithinBounds.errors(_, n)(n)).getOrElse(noMessages)
```
but this is hard to read so I will adapt your suggestion

PR #98, opened 2020-10-06
allowing arbitrary postconditions in pure functions

stammon:


Felalolf (approved):
Everything looks good to me.

Issue #97, opened 2020-10-06
Parse error if name of struct name has prefix of existing type


stammon:
If you have a struct whose name starts with a name of an existing type (such as int), the parser fails.

example:
``` go
package pkg;

type intPair struct {
    a int;
    b int;
};

func test(t intPair) (res int) {
    return t.a + t.b
}
```

When renaming intPair to twoInts, the parse error vanishes

jcp19:
This bug was fixed by PR #99 and is already merged in master.

PR #99, opened 2020-10-06
Fixed parsing error mentioned on issue #97

jcp19:
I changed the parser so that it does not eagerly parse type names and stops too early. For example, the parser now correctly parses the type identifier `intPair` instead of only accepting the prefix `int` and leaving `Pair` to be consumed.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Parser.scala
+++ b/src/main/scala/viper/gobra/frontend/Parser.scala
@@ -902,7 +902,7 @@ object Parser {
         exactWord("uint64") ^^^ PUInt64Type() |
         exactWord("uintptr") ^^^ PUIntPtr()
 
-    private def exactWord(s: String): Regex = (s ++ "\\b").r
+    private def exactWord(s: String): Regex = ("\\b" ++ s ++ "\\b").r
 
     lazy val qualifiedType: Parser[PDot] =
       declaredType ~ ("." ~> idnUse) ^^ PDot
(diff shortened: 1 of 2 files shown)


src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -883,22 +885,24 @@ object Parser {
...
-        "uint" ^^^ PUIntType()
+        exactWord("byte") ^^^ PByte() |
+        exactWord("uint") ^^^ PUIntType() |
+        exactWord("uint8") ^^^ PUInt8Type() |
+        exactWord("uint16") ^^^ PUInt16Type() |
+        exactWord("uint32") ^^^ PUInt32Type() |
+        exactWord("uint64") ^^^ PUInt64Type() |
+        exactWord("uintptr") ^^^ PUIntPtr()
+
+    private def exactWord(s: String): Regex = (s ++ "\\b").r
```
ArquintL, line 905: Why do you only use `\b` at the end of `s`? Doesn‘t it work if you add it on both sides?

jcp19: Yes, it does work if I add it on both sides.

jcp19: I was trying to minimize the regex to be matched on but i do not think that it makes a difference in this particular case. I will change that

ArquintL: It might be true that in our parser "blaint" will never be parsed as "bla" "int" but nevertheless (and given the function name) I prefer checking for word boundaries on both sides

ArquintL (approved):

PR #103, opened 2020-10-13
Basic support for option types

wytseoortwijn:
Adds some basic support for option types.


src/main/scala/viper/gobra/translator/implementations/components/OptionImpl.scala
```diff
@@ -0,0 +1,220 @@
...
+package viper.gobra.translator.implementations.components
+
+import viper.gobra.translator.interfaces.Collector
+import viper.gobra.translator.interfaces.components.Options
+import viper.silver.{ast => vpr}
+
+class OptionImpl extends Options {
+  private val domainName : String = "Option"
+  private val typeVar : vpr.TypeVar = vpr.TypeVar("T")
+  private var generated : Boolean = false
```
ArquintL, line 16: I like the var name in OptionToSeqImpl better:
```
  /**
     * Determines whether the Viper domain
     * should be generated upon finalisation.
     */
   private var generateDomain : Boolean = false
```

ArquintL (approved):

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -1100,6 +1105,15 @@ object Parser {
...
     }
 
+    lazy val optionNone : Parser[POptionNone] =
+      "none" ~> ("[" ~> typ <~ "]") ^^ POptionNone
+
+    lazy val optionSome : Parser[POptionSome] =
+      "some" ~> ("(" ~> expression <~ ")") ^^ POptionSome
+
+    lazy val optionGet : Parser[POptionGet] =
+      "option" ~> ("(" ~> expression <~ ")") ^^ POptionGet
```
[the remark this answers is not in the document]
wytseoortwijn: No, you are right in that "get" is much more standard and intuitive.

However, I didn't want to make "get" a reserved keyword, since it is probably used a lot in 'real-world' Go programs. So I opted for "option" instead, since it already is a reserved keyword.

Do you think it is better to use "get"?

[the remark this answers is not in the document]
Felalolf: Some langauges use "the", I believe.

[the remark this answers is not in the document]
ArquintL: I agree that it could be pretty annoying to have `get` as a reserved keyword. Do we have a general strategy so far? Simply don't care and force programmers to respect all reserved keywords or go for some special prefix to decrease likelihood (or even better use an invalid prefix in go)?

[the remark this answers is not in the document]
wytseoortwijn: I don't think there is a general strategy at the moment.

One possibility is to make `gobra` a reserved keyword, and use something like `gobra:get`.

jcp19 (approved):

src/main/scala/viper/gobra/frontend/Parser.scala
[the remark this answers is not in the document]
jcp19: That approach seems reasonable, if we ever need to add extra operations to the language we can always use the `gobra` keyword to prefix them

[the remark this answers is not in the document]
ArquintL: I like this idea as well. Another extension of that would be to write a "fake" import that would allow you to specify the qualifier (`gobra`, `.` or any other identifier)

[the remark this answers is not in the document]
wytseoortwijn: Yes, I like the "fake import" idea.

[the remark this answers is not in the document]
Felalolf: Me too.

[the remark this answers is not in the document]
Felalolf: There is a small problem there. Since Go does not support polymorphism, so the import does really have to be fake, since the functions we provide could not be defined in Go.

src/main/scala/viper/gobra/translator/implementations/components/OptionImpl.scala
wytseoortwijn: I agree. Just updated it.

src/main/scala/viper/gobra/translator/implementations/components/OptionImpl.scala
```diff
@@ -0,0 +1,225 @@
...
+    */
+  private lazy val option_ex_axiom : vpr.DomainAxiom = {
+    val oDecl = vpr.LocalVarDecl("o", vpr.DomainType(domainName, Map[vpr.TypeVar, vpr.Type]())(Seq(typeVar)))()
+    val eDecl = vpr.LocalVarDecl("e", typeVar)()
+
+    vpr.NamedDomainAxiom(
+      name = "option_ex",
+      exp = vpr.Forall(
+        Seq(oDecl),
+        Seq(),
```
Felalolf, line 79: Please, add a trigger.

src/main/scala/viper/gobra/translator/implementations/components/OptionImpl.scala
```diff
@@ -0,0 +1,225 @@
...
+  private lazy val optnone_some_axiom : vpr.DomainAxiom = {
+    val eDecl = vpr.LocalVarDecl("e", typeVar)()
+    val left = some(eDecl.localVar)()
+    val right = none(typeVar)()
+
+    vpr.NamedDomainAxiom(
+      name = "optnone_some",
+      exp = vpr.Forall(
+        Seq(eDecl),
+        Seq(),
```
Felalolf, line 110: Please, add a trigger.

src/main/scala/viper/gobra/translator/implementations/components/OptionImpl.scala
```diff
@@ -0,0 +1,225 @@
...
+    val oDecl = vpr.LocalVarDecl("o", vpr.DomainType(domainName, Map[vpr.TypeVar, vpr.Type]())(Seq(typeVar)))()
+    val eDecl = vpr.LocalVarDecl("e", typeVar)()
+    val left = get(oDecl.localVar, typeVar)()
+    val right = some(eDecl.localVar)()
+
+    vpr.NamedDomainAxiom(
+      name = "optsome_over_get",
+      exp = vpr.Forall(
+        Seq(oDecl, eDecl),
+        Seq(),
```
Felalolf, line 135: This is the last time that I will write it, but all quantified expressions should get a trigger. When you want to use the default trigger, then instantiate it explicitly. We should always know what our triggers are.

src/main/scala/viper/gobra/translator/implementations/components/OptionToSeqImpl.scala
```diff
@@ -0,0 +1,125 @@
...
+    Seq(vpr.LocalVarDecl("o", options.typ(typeVar))()),
+    vpr.SeqType(typeVar)
+  )(domainName = domainName)
+
+  /**
+    * Definition of the "opt2seq_none" axiom:
+    *
+    * {{{
+    * axiom opt2seq_none {
+    *   forall o : Option[T] :: opt2seq(o) == Seq() <==> o == optnone()
```
Felalolf, line 43: The axiomatization seems weird to me. I would expect something along the lines of:
`opt2seq(optnone()) == Seq()  && forall e :: opt2seq(optsome(e)) == Seq(e)`.
A separate axiom should specify that an option is either some or none. I think we had some previous work on how ADT are encoded into Viper.

Felalolf: I think that we usually use some "tag" utility function to distinguish between the elements.

Felalolf (changes requested):
Triggers are missing, I have some questions regarding the encoding, and the files GobraStrategy.scala and Nodes.scala have to be adapted. Lastly, the option getter should be changed from "option" to either "the" or "get". Regarding the proposed handling of name clashes, someone should create an issue.

src/main/scala/viper/gobra/translator/implementations/components/OptionImpl.scala
wytseoortwijn: That sounds reasonable. I updated (most of) the axioms.

src/main/scala/viper/gobra/translator/implementations/components/OptionToSeqImpl.scala
wytseoortwijn: I updated the axiom according to your suggestion. Regarding "tag" functions; do you have an example or pointer?

src/main/scala/viper/gobra/translator/implementations/components/OptionImpl.scala
wytseoortwijn: Done

src/main/scala/viper/gobra/translator/implementations/components/OptionImpl.scala
wytseoortwijn: Do you have a suggestion for a trigger for
```
axiom {
  forall o : Option[T] :: o == optnone() || exists e : T :: o == optsome(e)
}
```
or otherwise an alternative axiom?

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1476,16 +1477,16 @@ object Desugar {
           dop <- go(op)
         } yield dop.typ match {
           case _: in.SequenceT => dop
-          case _: in.ArrayT => in.SequenceConversion(dop)(src)
-          case t => violation(s"expected a sequence or exclusive array type, but got $t")
+          case _ => in.SequenceConversion(dop)(src)
```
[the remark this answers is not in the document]
wytseoortwijn: Fixed

src/main/scala/viper/gobra/translator/implementations/components/OptionImpl.scala
wytseoortwijn: Nevermind, this axiom is superfluous

src/main/scala/viper/gobra/translator/implementations/components/OptionToSeqImpl.scala
Felalolf: I will send you a mail with two sources in a minute.

Felalolf:
Please write when I should have a look at the PR again.

src/main/scala/viper/gobra/translator/implementations/components/OptionImpl.scala
```diff
@@ -104,7 +77,8 @@ class OptionImpl extends Options {
   /**
     * {{{
     * axiom optsome_over_get {
-    *   forall o : Option[T], e : T :: optget(o) == e <==> o == optsome(e)
+    *   forall o : Option[T], e : T :: { optget(o), optsome(e) }
+    *     (optget(o) == e) <==> (o == optsome(e))
```
Felalolf, line 81: This axiom is unsound (incombination with other axioms). 
My - longer - explanation:
1) The axiom implies that get is injective without any restriction
2) If get is complete on type option, then get is not injective, because `get(none) == get(some(get(none)))` and `none != some(...)`

I also double checked with Martin and his - shorter - explanation is: substitute `o` with `none` and `e` with `get(none)`, then the axiom yields `none == some(get(none))` which again is invalid.

src/main/scala/viper/gobra/frontend/Parser.scala
Felalolf, line 1115: can you change the name before it is merged in? You could use the, get, option-get, etc

Felalolf (changes requested):

src/main/scala/viper/gobra/translator/implementations/components/OptionImpl.scala
wytseoortwijn: Yes, that's indeed true; thanks for spotting that!

I updated the axioms to be similar to the ADT encoding example on http://viper.ethz.ch/examples/encoding-adts.html.

(I could make it slightly less elaborate if that's better, by just using a simple `tag` function to replace the constructor type business.)

src/main/scala/viper/gobra/frontend/Parser.scala
wytseoortwijn: Yes, fixed it. It now uses "get".

src/main/scala/viper/gobra/translator/implementations/components/OptionImpl.scala
```diff
@@ -0,0 +1,439 @@
...
+    *   }
+    *
+    *   axiom {
+    *     forall o : Option[T] :: { optIsNone(o) } optType(o) == optTypeNone() <==> optIsNone(o)
+    *   }
+    *
+    *   axiom {
+    *     forall o : Option[T] :: { optIsSome(o) } optType(o) == optTypeSome() <==> optIsSome(o)
+    *   }
+    * }
```
Felalolf, line 328: The domain is indeed quite large now.
How about something like the following:

```
domain Option[T] {

  function some(e: T): Option[T]
  function unit(): Option[T]
  function get(o: Option[T]): T
  function isUnit(o: Option[T]): Bool

  axiom {
    forall e: T :: { some(e) } get(some(e)) == e && isUnit(some(e)) == false
  }

  axiom {
    forall o: Option[T] :: { get(o) } isUnit(o) == false ==> o == some(get(o))
  }

  axiom {
    isUnit((unit(): Option[T])) == true
  }

  axiom {
    forall o: Option[T] :: { isUnit(o) } o == unit() || exists e: T :: {some(e)} o == some(e)
  }
}
```

With some tests:
```
domain Dummy[T] {
  function dummy(x: T): Bool
  axiom {
    forall x: T :: { dummy(x) } dummy(x)
  }
}


method test1() {
  var x: Option[Int] := some(4)
  assert x != unit()
  assert get(x) == 4
}

method test2(x: Option[Int]) {
  if (x != unit()) {
    assert dummy(get(x)) // hard to avoid
    assert exists e: Int :: some(e) == x
  }
}
```

src/main/scala/viper/gobra/translator/implementations/components/OptionImpl.scala
```diff
@@ -0,0 +1,439 @@
...
+    *
+    *   // Destructor
+    *
+    *   function optGet(o : Option[T]) : T
+    *
+    *   // Constructor Types
+    *
+    *   function optType(o : Option[T]) : Int
+    *   unique function optTypeNone() : Int
+    *   unique function optTypeSome() : Int
```
Felalolf, line 293: I was not even aware of this viper feature. Why do you not just substitute values such as 0 and 1?

Felalolf (approved):

Issue #43, opened 2020-07-30
Recursion in type definitions


viper-admin:
> Created by bitbucket user **Wytse Oortwijn** on 2020-07-30 08:53

Gobra crashes with a `StackOverflowError` when trying to deal with invalid recursion in type definitions, like for example:

```
type Wrong struct {
  w Wrong
}
```

Would be nice to have an extra check for this in the type checker or so.

PR #104, opened 2020-10-14
Fix issue 43

jcp19:
Add type system level check for invalid recursive type definitions as described in issue #43

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
@@ -33,7 +33,7 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
     case n: PTypeDecl => isType(n.right).out ++ (n.right match {
       case s: PStructType =>
         val typ = n.left.name
-        message(n, s"invalid recursive type $typ", invalidRecursiveType(n.left.name, s))
+        message(n, s"invalid recursive type $typ", cyclicStructDef(n.left.name, s))
       case _ => noMessages
     })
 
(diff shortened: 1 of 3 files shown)


src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
```diff
@@ -120,4 +125,19 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
 
     case s => violation(s"$s was not handled")
   }
+
+  /**
+    * Checks whether a struct whose type has name 'name' is cyclically defined (in terms of itself or other cyclic structures).
+    * `visitedTypes` keeps track of the types that were already discovered and checked, used for detecting cyclic definition chains
+    */
+  private def cyclicStructDef(struct: PStructType, visitedTypes: Set[String] = Set.empty) : Boolean = {
```
ArquintL, line 133: `Set.empty` seems to be a bad default value as the declared name should initially be contained in the set

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
```diff
@@ -120,4 +125,19 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
 
     case s => violation(s"$s was not handled")
   }
+
+  /**
+    * Checks whether a struct whose type has name 'name' is cyclically defined (in terms of itself or other cyclic structures).
```
ArquintL, line 130: `name` does not appear as any parameter. Also regarding the `visitedTypes`'s default value, it might make sense to change the parameters to be of type `PStructType` and `PIdnDef` and the recurse on some inner function

jcp19: I agree with your suggestions. I implemented them in a new commit

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
jcp19: Not necessarily if you are checking a structure which has no assigned name (e.g. a struct literal)

jcp19: Even in such a case, a struct may be invalid if it has a field of a type which is itself an invalid recursive struct

ArquintL: Makes sense but shouldn't this check then be performed in `wellDefActualType` with `name` being `None`?

jcp19: Yes, I also realized that and I am changing it :)

jcp19: Just added in latest commit

ArquintL (approved):

Felalolf (approved):

PR #106, opened 2020-10-16
Automatic Copyright Header Checking

ArquintL:
Adds copyright header checking and adds some copyright headers


Felalolf (approved):
Still a good idea

PR #107, opened 2020-10-21
Improved SBT Caching

ArquintL:
Updates the caching behavior of SBT in CI by forcing a cache update after every run. This should minimize the number of files that need to be recompiled in the next run as the most recent cache version will be used. This change has already been implemented in Gobra IDE and is now coming to Gobra as well.


ArquintL:
Based on feedback by @jcp19, the caching of compiled scala files has been removed as NoSuchMethodError exceptions have occurred multiple times at runtime. This indicates that sbt incorrectly computes files that have to be recompiled.
The primary cache key has been updated to reflect this change and to avoid reusing old caches

Felalolf (approved):

Issue #105, opened 2020-10-16
Add overflow check support


jcp19:
Add optional overflow checks to Gobra that report every arithmetic operation that may cause an overflow.

Issue #95, opened 2020-10-03
Add flag to specify the default size of int's


jcp19:
The Go language specification defines the `int` and `uint` types as having either 32 bits or 64, as long as they have the same size (https://golang.org/ref/spec#Numeric_types).

At the moment, Gobra assumes that they have 32 bits but we should add a flag that changes the size of `int` to 64bit.

jcp19:
Besides, we can also add this property to a configuration file

jcp19:
It would also be useful to be able to specify the size of the numeric constants - the go spec says that they may have arbitrary precision but it only requires that constants have at least 256 bits.

From the go specification
>  Implementation restriction: Although numeric constants have arbitrary precision in the language, a compiler may implement them using an internal representation with limited precision. That said, every implementation must:
> - Represent integer constants with at least 256 bits.
> ...
> - Give an error if unable to represent an integer constant precisely.

PR #108, opened 2020-10-24
Add overflow checks to Gobra

jcp19:
Changelog
-----
This PR adds the following features to Gobra:
- optional overflow checks to Gobra when running with the `-o` flag (issue #105)
  - custom error messages for overflow errors (with the help of @ArquintL)
- option to change the default size of `int` and `uint` from 32 bit to 64 with the `--int64` flag (issue #95)
- a step in Gobra's pipeline to perform-to-program transformations
- extend integer typing information to the internal language (issue #91)

Overflow Checking
-----
Adding overflow checks to a Gobra program is implemented through a program-to-program transformation at the internal language level. This transformation performs 2 steps:
1. For every (non-pure) method and function, add assume statements at the beginning of the corresponding body to ensure that each argument is within the bounds of the respective type. For each statement in the block that contains an expression, check if that expression may lead to an overflow.
2. For every pure method and function, add the bounds of each argument's value to the pre-condition of the method and add an assertion that the return value (which must be unique) is within the bounds of the return type.

As an example, the following expression
```go
ensures ret == u + 1
func incrementU8(u uint8) (ret uint8) {
	ret = u + 1
	return
}
```
is transformed into
```go
ensures ret == u + 1
func incrementU8(u uint8) (ret uint8) {
	assert 0 <= u + 1 && u + 1 <= 255
        ret = u + 1
	return
}
```
which fails leading to the error
```
[info] <42:8> Expression may cause integer overflow. 
[info] Expression u + 1 may cause integer overflow
```

The encoding of overflow checks is sensitive to the size of the `int` and `uint` size, that can be changed by the user with the `--int64` flag. For example, the following code would lead to an overflow error in 32-bit mode but it would be fine in 64-bit mode.
```go
requires i <= 2147483647
ensures res == i + 1
func overflowDefaultInt(i int) (res int) {
	return i + 1
}
```

The current implementation in practice generates overflow checks for every sub-expression of an expression. The reason for this is that there may be expressions that may never overflow but that may rely on overflow in its sub-expressions such as
```go
func subExp1(x uint8) uint8 {
	return (x * x) / (x + 1)
}
```
In such an expression, `x * x` and `x + 1` may both lead to overflows even though the result of the operation as a whole will be within bounds of uint8.

Tests
----
I added two test suites for overflow checking, one that runs in 32bit mode and the other that runs in 64bit mode


jcp19:
I am experiencing the same problem with caching that I was having before. For some reason, Github action's sbt cache stores an outdated version of the `Config` constructor which causes the tests to abort. In my machine, the tests run successfully. The tests also pass if I disable the workflow's cache.

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
```diff
@@ -0,0 +1,142 @@
...
+    // Same as pure functions
+    case m@PureMethod(receiver, name, args, results, pres, posts, body) => body match {
+      case Some(expr) =>
+        val newPre = pres ++ getPureBlockPre(args)
+        val newPost = posts ++ Vector(getPureBlockPost(expr, results))
+        PureMethod(receiver, name, args, results, newPre, newPost, body)(m.info)
+      case None => m
+    }
+
+    // TODO: should we add overflow checks to predicates?
```
ArquintL, line 52: Let's discuss this today in the Gobra meeting

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
```diff
@@ -0,0 +1,142 @@
...
+    assertionExprInBounds(body, results(0).typ)(addAnnotation(body.info))
+  }
+
+  private def stmtTransform(stmt: Stmt): Stmt = stmt match {
+    case b@Block(decls, stmts) => Block(decls, stmts map stmtTransform)(b.info)
+
+    case s@Seqn(stmts) => Seqn(stmts map stmtTransform)(s.info)
+
+    case i@If(cond, thn, els) =>
+      val condInfo = addAnnotation(cond.info)
```
ArquintL, line 91: I don't understand what `addAnnotation` does / how it works. I would have assumed that you add some back-tracking information for the newly generated assertions but why is it called on the original condition?

src/main/scala/viper/gobra/ast/internal/transform/InternalTransform.scala
```diff
@@ -0,0 +1,28 @@
...
+
+  /**
+    * Program-to-program transformation
+    */
+  def transform(p: Program): Program
+
+  /**
+    * Lifts a member-to-member transformation to a program-to-program transformation
+    */
+  def transformMembers(memberTrans: Member => Member): Program => Program = {
```
ArquintL, line 25: Nice!

src/main/scala/viper/gobra/reporting/Source.scala
```diff
@@ -17,6 +17,9 @@ object Source {
 
   case class Origin(pos: SourcePosition, tag: String)
 
+  abstract class Annotation()
+  class AnnotatedOrigin(val origin: Origin, val annotation: Annotation) extends Origin(origin.pos, origin.tag)
```
ArquintL, line 21: could this be a case class?

src/main/scala/viper/gobra/reporting/Source.scala
```diff
@@ -39,6 +42,9 @@ object Source {
       override lazy val origin: Option[Origin] = Some(src)
       override def vprMeta(node: internal.Node): (vpr.Position, vpr.Info, vpr.ErrorTrafo) =
         (vpr.TranslatedPosition(src.pos), Verifier.Info(pnode, node, src), vpr.NoTrafos)
+
+      def annotateOrigin(annotation: Annotation): Single =
+        new Single(pnode, new AnnotatedOrigin(src, annotation))
```
ArquintL, line 47: `new` keyword shouldn't be necessary (at least the first one)

src/test/resources/overflow_checks/int64/overflow.gobra
```diff
@@ -0,0 +1,19 @@
...
+// Should overflow when running in 32 bit mode, not in 64
+requires i <= 2147483647
+ensures res == i + 1
+func overflowInt32(i int) (res int) {
+	return i + 1
+}
+
+// Operation may overflow, no bound checks are performed
+ensures ret == u + 1
+func incrementU8(u uint) (ret uint) {
```
ArquintL, line 15: why is this function called `incrementU8`?
In addition, I do not understand the above comment

src/test/scala/viper/gobra/GobraOverflowCheckTests.scala
```diff
@@ -0,0 +1,75 @@
...
+import viper.gobra.frontend.Config
+import viper.gobra.reporting.NoopReporter
+import viper.gobra.reporting.VerifierResult.{Failure, Success}
+import viper.silver.testing.{AbstractOutput, AnnotatedTestInput, ProjectInfo, SystemUnderTest}
+import viper.silver.utility.TimingUtils
+
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+
+sealed abstract class GobraOverflowCheckTests extends GobraTests {
```
ArquintL, line 21: Would it be possible to replace this class by inline command line options?
I.e. move `resources/overflow_checks` to `resources/regressions/features/overflow_checks` and add the following line at the beginning of the files in the `int64` subdir:
```
// ##(--overflow --int64)
```
resp. for files in `int32` subdir:
```
// ##(--overflow)
```

ArquintL (changes requested):

src/main/scala/viper/gobra/reporting/DefaultErrorBackTranslator.scala
```diff
@@ -60,7 +62,13 @@ object DefaultErrorBackTranslator {
...
       InsufficientPermissionError(info)
     case vprrea.AssertionFalse(Source(info)) =>
-      AssertionFalseError(info)
+      info.origin match {
+        case origin: AnnotatedOrigin => origin.annotation match {
+          case OverflowCheckAnnotation => OverflowErrorReason(info)
+          case _ => ???
+        }
+        case _ => AssertionFalseError(info)
+      }
```
Felalolf, line 71: Instead of adding case splits here, could we instead introduce a separate step before in which the annotated errors are translated.

src/main/scala/viper/gobra/reporting/VerifierError.scala
```diff
@@ -180,6 +186,11 @@ case class SeqIndexNegativeError(node: Source.Verifier.Info, index: Source.Verif
   override def message: String = s"Index ${index.origin.tag.trim} into ${node.origin.tag.trim} might be negative"
 }
 
+case class OverflowErrorReason(node: Source.Verifier.Info) extends VerificationErrorReason {
+  override def id: String = "integer_overflow_error"
+  override def message: String = s"Expression ${node.origin.tag} may cause integer overflow"
+}
+
```
Felalolf, line 193: Is this actually required. I would assume that OverflowError suffices.

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -138,6 +140,21 @@ class Gobra extends GoVerifier with GoIdeVerifier {
     }
   }
 
+  /**
+    * Applies transformations to programs in the internal language. Currently, only adds overflow checks but it can
+    * be easily extended to perform more transformations
+    */
+  private def performInternalTransformations(program: Program, config: Config): Either[Vector[VerifierError], Program] = {
+    var result = program
```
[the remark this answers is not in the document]
jcp19: Sure, I will change it :)

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -557,6 +610,13 @@ sealed abstract class BinaryExpr(val operator: String) extends Expr {
   def right: Expr
 }
 
+sealed abstract class BinaryIntExpr(override val operator: String) extends BinaryExpr(operator) with IntOperation {
+  override val typ: Type = (left.typ, right.typ) match {
+    case (IntT(_, kind1), IntT(_, kind2)) => IntT(Addressability.rValue, TypeBounds.merge(kind1, kind2))
```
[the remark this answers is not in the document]
jcp19: In principle, a well typed program should never lead to a situation where we have two unmergeable kinds here. If nonetheless such a case arises in practice, the `merge` function will throw a violation as this clearly breaks our assumptions.

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
```diff
@@ -0,0 +1,142 @@
...
+/**
+  * Adds overflow checks to programs written in Gobra's internal language
+  */
+object OverflowChecksTransform extends InternalTransform {
+  override def name(): String = "add_integer_overflow_checks"
+
+  override def transform(p: Program): Program = transformMembers(memberTrans)(p)
+
+  private def memberTrans(member: Member): Member = member match {
+    // Adds overflow checks per statement that contains an expression and assume statements in the beginning
```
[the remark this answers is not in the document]
jcp19: I agree

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -39,7 +38,9 @@ case class Config(
                  shouldTypeCheck: Boolean = true,
                  shouldDesugar: Boolean = true,
                  shouldViperEncode: Boolean = true,
-                 shouldVerify: Boolean = true
+                 checkOverflows: Boolean = false,
+                 shouldVerify: Boolean = true,
+                 int64bit: Boolean = false // picks the size of int's, 32 if false, 64 bit otherwise
```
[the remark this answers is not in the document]
jcp19: Not according to Go's specification (https://golang.org/ref/spec#Numeric_types) which states that integers may be either 32bit or 64bit. I would not mind changing the flag to something else but given the current Go spec, I don't see the point in doing that now.

src/main/scala/viper/gobra/ast/internal/Program.scala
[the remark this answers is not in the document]
ArquintL: Ah so these errors get caught during type checking

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
jcp19: I indeed add an `OverflowCheckAnnotation` label to a node's `info` field. When an error is reported, I check whether the `info` field is marked with such an annotation and in such a case, I transform its error message from an assertion error to an overflow error. In this particular case, I'm adding an annotation to the branch condition because I am generating an overflow check for that expression and I want to be able to point to this expression if an overflow is detected here.

ArquintL: Ah so this works because the generated assertion stmt and the node that causes this generation share the same `info`? 
Another consequence is that you couldn't distinguish which of the following two assertion failed and you would both translated to an overflow error:
```
assert 4 + 5 < MAX_INT // generated overflow check
assert 4 + 5 == 9 // original assertion
```
But this case is not an issue as (as discussed yesterday) no overflow checks should be produced for expressions in assertions, right?

src/test/scala/viper/gobra/GobraOverflowCheckTests.scala
jcp19: That is a great suggestion, I will change it!

src/test/resources/overflow_checks/int64/overflow.gobra
jcp19: Just a copy past error. I'm changing the comment to "Operation may lead to overflow, no bound checks are performed on the argument u before incrementing it" which I believe is clearer

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
jcp19: It is able to distinguish between the two cases because, on the first assertion, the assert statement would be "marked" with an `OverflowCheckAnnotation` and the origin information of the expression node that caused the generation of the assertion whereas the second one would not. As an demo, the program
```go
// Should overflow when running in 32 bit mode, not in 64
requires i <= 2147483647
ensures res == i + 1
func overflowInt32(i int) (res int) {
	assert 5 + 1 == 5
	return i + 1
}
```
leads to the error
```
[info] 	<28:2> Assert might fail. 
[info] Assertion 5 + 1 == 5 might not hold
```
but if we comment the line ```assert 5 + 1 == 5``` then it produces the error
```
[info] 	<29:9> Expression may cause integer overflow. 
[info] Expression i + 1 may cause integer overflow
```

jcp19: Regarding the second observation about not producing overflow checks for assertions, this is achieved by mapping Assert, Assume, Inhale and Exhale statements to themselves in the program transformation that adds Overflow Checks

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
jcp19: Added a small comment about this on 46a41771680dbcbaf8764875afa5e8a9d084bd01

src/main/scala/viper/gobra/reporting/Source.scala
jcp19: I don't think it can because it extends `Origin` which is itself a case class and case-to-case inheritance is prohibited. I could however create an `AbstractOrigin` class that is extended by both `Origin` and `AnnotatedOrigin`. Would this be a better option?

src/main/scala/viper/gobra/reporting/Source.scala
jcp19: Removed the first one

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -211,6 +214,56 @@ case class MemoryPredicateAccess(arg: Expr)(val info: Source.Parser.Info) extend
 
 sealed trait Expr extends Node with Typed
 
+object Expr {
+  def getProperSubExpressions(expr: Expr): Set[Expr] = expr match {
```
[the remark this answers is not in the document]
jcp19: Sorry, I missed this entirely

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
```diff
@@ -0,0 +1,142 @@
...
+    * Computes the post-conditions to be added to pure functions and methods to check for overflows, i.e.
+    * that the expression result is within the bounds of its type
+    */
+  private def getPureBlockPost(body: Expr, results: Vector[Parameter.Out]): Assertion = {
+    // relies on the current assumption that pure functions and methods must have exactly one result argument
+    if (results.length != 1) violation("Pure functions and methods must have exactly one result argument")
+    assertionExprInBounds(body, results(0).typ)(addAnnotation(body.info))
+  }
+
+  private def stmtTransform(stmt: Stmt): Stmt = stmt match {
```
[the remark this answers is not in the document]
jcp19: I also thought about this and, in the end, I chose to go with adding overflow checks to statements because it is not always possible to add an overflow check in the scope that an expression appears. For example, when you have a pure function, you cannot add an assertion to its body to check that the result of its expression is within bounds because the body can only contain one expression. As such, I chose to add overflow checks in the scopes that I see fit (such as specific statements (e.g. If statements, while loops, assignments) and also, pure methods and functions (by adding the overflow checks to their contracts) and regular methods and functions. So, if the expression that you provided appears in such a context such as an assignment
```
x := foo(x + 202020020202020)
```
then it will be detected as `x + 202020020202020` is a subexpression of ` foo(x + 202020020202020)`. Otherwise, it wont be detected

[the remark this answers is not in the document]
jcp19: With respect to what was said yesterday at the meeting, I realized that we don't have an explicit ghost modifier for statements (that I am aware of) and as such, this may lead to unexpected results when we increment a ghost variable, as this will be treated as a regular assignment and overflow checks will be generated. Should we add one?

src/main/scala/viper/gobra/reporting/Source.scala
Felalolf: yes

src/main/scala/viper/gobra/frontend/Config.scala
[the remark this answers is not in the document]
Felalolf: Then please add a comment about that to the line of code.

src/main/scala/viper/gobra/reporting/DefaultErrorBackTranslator.scala
jcp19: Sure, that seems way cleaner!

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -40,7 +40,10 @@ case class Config(
                  shouldViperEncode: Boolean = true,
                  checkOverflows: Boolean = false,
                  shouldVerify: Boolean = true,
-                 int64bit: Boolean = false // picks the size of int's, 32 if false, 64 bit otherwise
+                 // The go language specification states that int and uint variables can have either 32bit or 64, as long
+                 // as they have the same size. This flag allows users to pick the size of int's and uints's: 32 if false,
+                 // 64 bit otherwise.
+                 int64bit: Boolean = false
```
Aurel300, line 46: From my cursory glance at the spec I understand it to mean that the size of `int`s and `uint`s is actually based on the CPU architecture/word size. Would it be reasonable to try to detect this at runtime?

As a separate point, would a 64-bit default not be better, given how common 64-bit architectures are in personal computers now?

jcp19: I would not try to detect this at runtime (at least, not by default) because you may be performing verification on a different machine than the one that will run the code (e.g. in case you want to perform verification on a more powerful machine). I agree with your second point

src/main/scala/viper/gobra/reporting/VerifierError.scala
jcp19: I would argue that it is needed in order to have meaningful error messages. If this class was not here, we would most likely use the `AssertionFalseError` which would lead to messages such as `Assertion X might not hold`, where X is the condition that causes the overflow error

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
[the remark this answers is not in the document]
ArquintL: There are ghost stmts in Gobra, you can simply put `ghost` in front of a stmt (see `ghostStatement` in the parser).
For now I would create overflow checks for ghost stmts as well as it seems very related to the discussion of mathematical types

ArquintL (approved):

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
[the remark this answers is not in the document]
jcp19: But not at the level of the internal language right? I may have missed something

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
ArquintL: > It is able to distinguish between the two cases because
Are you sure? I meant a program that only contains the assertion stmt `assert 4 + 5 == 9` which would then result in generating the other assert stmt (to check for a potential overflow). My understanding so far is that you call `addAnnotation` on the original assert stmt and that this works because both assert stmts share the same `info` object. If my thoughts are correct, then an 'assertion might not hold' error is transformed to an overflow error for both assert stmts

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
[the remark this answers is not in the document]
ArquintL: No I think you're right. It looks like this information is lost during desugaring

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
jcp19: The info field is not necessarily shared. When I call `addAnotation`, I create a new `Info` value that contains the previous one but is annotated with a label to identify it as an overflow check. I then use it as the info field only on expressions that check for overflows

jcp19: The info field is not necessarily shared. When I call `addAnotation`, I create a new Info value that contains the previous one but is annotated with a label to identify it as an overflow check. I then use it as the info field only on expressions that check for overflows

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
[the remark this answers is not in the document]
Felalolf: I am confused, to my knowledge, we want to generate the bound checks also for ghost code, so there should not be a problem.

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
ArquintL: Ah! I totally missed that `addAnnotation` returns a new info object. I thought you're attaching some information to the existing info object

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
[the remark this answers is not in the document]
ArquintL: Yes that was my take away message from the meeting as well

src/main/scala/viper/gobra/frontend/Config.scala
ArquintL: @jcp19 I agree with you as well. Will you adapt the default value of `int64bit`?

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
```diff
@@ -0,0 +1,146 @@
...
+    * Computes the post-conditions to be added to pure functions and methods to check for overflows, i.e.
+    * that the expression result is within the bounds of its type
+    */
+  private def getPureBlockPosts(body: Expr, results: Vector[Parameter.Out]): Vector[Assertion] = {
+    // relies on the current assumption that pure functions and methods must have exactly one result argument
+    if (results.length != 1) violation("Pure functions and methods must have exactly one result argument")
+    Vector(assertionExprInBounds(body, results(0).typ)(addAnnotation(body.info)))
+  }
+
+  private def stmtTransform(stmt: Stmt): Stmt = stmt match {
```
Felalolf, line 89: The other discussion got kind of hijacked, so I am continuing it here.
Only checking some statement does not really look like a concistent approach to me. How do you know that all statements are covered? What is with calls without an assignemnt, e.g. `x, y := 20200202020, 20020200202; foo(x + y)`? What is with make statements and return statements? What is when we add a new statement?

Also, is it ok to not cover pure functions? You could add an assert expression to the internal representation to handle these cases.

Lastly, and this one is more complicated, what is with conditional code, e.g. `x, y := MAXINT, MAXINT; b := false; z := b ? x + y : 4`?

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
```diff
@@ -0,0 +1,146 @@
...
+    case b@Block(decls, stmts) => Block(decls, stmts map stmtTransform)(b.info)
+
+    case s@Seqn(stmts) => Seqn(stmts map stmtTransform)(s.info)
```
Felalolf, line 92: I think the rewriting framework here is a bit flawed, but we can change that later. 
At an earlier point in time, Gobra reused Viper's rewriting framework. Unfortunately, due to changes to Viper this was not possible anymore (Before the framework was independend of the Viper AST. Then reflections were added thatenforce that the ast must have the Viper meta data ...). Back then, I objected to this change heavily, because now (still?) we would have to copy the entire rewriting framework and change how the meta data is handled. It might be worth to check if this is still the case and if not, then we should reuse Viper's framework again.

Felalolf: Ah, I think that the rewriting framework is flawed because there is a lot of redundant code like the two cases that I marked. If more compound statements are added, then you need to add these lines for each of them.

src/main/scala/viper/gobra/reporting/DefaultErrorBackTranslator.scala
```diff
@@ -112,9 +125,23 @@ class DefaultErrorBackTranslator(
...
-      AssertError(info) dueTo translate(reason)
+      info.origin match {
+        case origin: AnnotatedOrigin =>
+          origin.annotation match {
+            case OverflowCheckAnnotation => OverflowError(info) dueTo translate(reason)
+            case _ => ???
+          }
+        case _ =>
+          AssertError(info) dueTo translate(reason)
+      }
```
Felalolf, line 136: The split into the default translater and a default translator for annotated code has not happened yet, right? Or am I missing something?

Felalolf (changes requested):

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
```diff
@@ -0,0 +1,146 @@
...
+
+    case x => x
+  }
+
+  /**
+    * Adds overflow checks to the bodies of a block. If block is parametrized,
+    * `computeNewBody` adds assumptions about the parameters' bounds
+    */
+  private def computeNewBody(args: Vector[Parameter.In], body: Block): Block = {
+    val assumeArgBounds = args map { case p@Parameter.In(_, typ) => Assume(assertionExprInBounds(p, typ)(p.info))(p.info) }
```
Felalolf, line 66: I mentioned this a bunch of times, but think that just adding the information that the parameter are within their bounds is too weak too show many properties. But again, we can change this later, when we see that stronger assumptions are necessary. It is an ok first step.

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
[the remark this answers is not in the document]
jcp19: Sorry, my bad

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
jcp19: I will change the name `addAnnotation` to `createAnnotatedInfo` to make it clearer

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
jcp19: 1. You are right, I am going to add checks for the other kinds of statements that are missing.

2. I am covering them already by adding the overflow checks to the post-condition. I did not add them to the body though because it consists of an Option[Expr]

3. This triggers an Overflow Error which is in practice unreachable. I agree that this seems overly conservative but it could also be argued that this expression contains a  sub-expression that causes an overflow. In either case, I can improve this in a future PR

src/main/scala/viper/gobra/frontend/Config.scala
jcp19: Yes, changed default int size to 64bit

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
jcp19: Sorry for the late reply to your comment. Should I bring this up in the next Viper meeting? In the case that we cannot re-use the rewriting framework from Viper, I will improve this (maybe in a future PR).

Felalolf: That is a good idea, I might ask this at the meeting if the situation fits.

src/main/scala/viper/gobra/reporting/DefaultErrorBackTranslator.scala
jcp19: I added it to the `defaultReasonTransformer` function but I overlooked the case split in `defaultErrorTransformer` function. Changed that in last commit

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
jcp19: For completeness sake, I'm pasting this message here
> in a first look and assuming that it is implemented in the traits Rewritable (https://github.com/viperproject/silver/blob/master/src/main/scala/viper/silver/ast/utility/rewriter/Rewritable.scala) and Strategy (https://github.com/viperproject/silver/blob/master/src/main/scala/viper/silver/ast/utility/rewriter/Strategy.scala), it seems that the framework is dependent on the Viper AST and I am not sure if we can reuse that in a non-hacky way

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
jcp19: So, I updated the program transformer to explicitly match every kind of statement and to generate overflow checks for all "non-logical" statements (i.e. for all statements excluding `Inhale`, `Exhale`, `Assert`, `Assume`, `Fold` and `Unfold`, and also `Return` as it does not contain any expression). If we ever add a new statement, the compiler will let us know that we have a non-exhaustive pattern matching.

I also made some simplifications and detected and fixed a bug in cases like this one
```go
type rect struct {
    width, height int
}

func (r rect) getWidth() int {
	return r.width // was throwing an overflow error even though it does not make sense
}
```

The reason for this is that Gobra was not able to prove that the field was within bounds, even though the type system ensures that it is. I thought about adding axioms to the structures asserting that every integer field access is within bounds however this was not so easy to do given that structs are translated to polymorphic tuples at the Viper level. What I ended up doing is to prove that no overflow occurs in an expression, assuming that the struct fields it accesses are within bounds.

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    val computeAssertions = (exprs: Set[Expr]) => exprs.map{x => genAssertionExpr(x, x.typ)}.foldRight(trueLit)((x,y) => And(x,y)(info))
+
+    // assumptions for the values that are considered within bounds
+    val assumptions = computeAssertions(valuesWithinBounds)
+
+    // Assertions that need to be verified assuming the expressions in `assumptions`
+    val obligations = ExprAssertion(computeAssertions(intSubExprs))(info)
+    Implication(assumptions, obligations)(info)
```
Felalolf, line 156: The assumptions about the values of field reads is only added here, right? But if it is only added as an asumption, then what happens if the assumption comes from a different line? e.g. `x := y.f; return x`

jcp19: This is done not only for field reads but also for variables, indexed expressions and pointer dereferences. As such, the value of a variable will always be assumed to be within bounds (my point here is that when the result of an operation that overflows is stored in a variable, we don't need to check for overflows in posterior uses of the variable because we check for overflows when we do the assignment).

Felalolf (approved):

PR #112, opened 2020-10-29
Fixes IllegalStateException of ViperServer VerificationPromise

ArquintL:
Calling `success` or `failure` on a completed promise results in an IllegalStateException.
Because an exception could occur while reporting any message, the `verificationPromise` might already be completed when the `OverallSuccessMessage` or `OverallFailureMessage` is received.
This PR switches to `trySuccess` and `tryFailure` that result in a no-op if the promise is already completed

### DIFF
--- a/src/main/scala/viper/gobra/backend/ViperServer.scala
+++ b/src/main/scala/viper/gobra/backend/ViperServer.scala
@@ -39,15 +39,15 @@ object ViperServer {
           reporter.report(msg)
 
           msg match {
-            case msg: OverallFailureMessage => verificationPromise success msg.result
-            case _: OverallSuccessMessage   => verificationPromise success Success
+            case msg: OverallFailureMessage => verificationPromise trySuccess msg.result
+            case _: OverallSuccessMessage   => verificationPromise trySuccess Success
             case _ =>
           }
         } catch {
-          case e: Throwable => verificationPromise failure e
+          case e: Throwable => verificationPromise tryFailure e
         }
 
-      case e: Throwable => verificationPromise failure e
+      case e: Throwable => verificationPromise tryFailure e
     }
   }
 }


Felalolf (approved):
In theory, either sucess or failure should be executed once with the old code,
meaning that the code should work as long as ViperServer does not misbehave.
But I do not mind the change.

ArquintL:
I wouldn't call this a misbehavior of ViperServer. Imagine the following case (which actually happened in practice):
1. Verify the example by Wytse that previously failed to back translate
2. At some point, there is a partial verification error message reporting the result for verifying just one method
3. Backtranslation already fails here as the partial verification message basically includes the same error reason
4. Promise is completed with a failure as back translation has failed
5. After potential further partial messages, the overall (failure) message arrives
6. Independently of failure, success or even crash while backtranslating the overall message, any call to complete the promise will fail at this point (unless `try` is used)

I agree that only failing the promise once at the end would be nicer as right now the promise is completed too early. The only alternative that comes right now to my mind would be as follows:
- keep track of whether a message had to be caught so far (maybe including the first exception description)
- expect an overall message and then check whether an exception has been observed so far and fail correspondingly.

Although this seems like a plan for failing back translations, I'm not sure how I would handle the other case, i.e. directly receiving an exception from akka. Should this immediately lead to a promise failure or would be nevertheless expect an overall message?

Felalolf:
I see. Thank you for the clarification.

ArquintL:
> I see. Thank you for the clarification.

So what do you think? Should I leave it as it is right now in the PR or implement the thing I've just sketched?

Felalolf:
I think you should leave it as it is right now in the PR. These cases are caused by Gobra crashing.
Whenever Gobra crashes, I consider this a bug, thus we do not need complicated solutions to adapt to them.

PR #113, opened 2020-10-29
Adapted Gobra to new adressability modifier design

Felalolf:
The modifier for shared variables is now @ instead of !.
Furthermore, parameters cannot be annotated as shared anymore. 
Instead the first line of the code block has a statement `share x,y,z` where `x,y,z` are the parameters that are used as shared variables in the body of the method.


src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -691,6 +691,11 @@ case class PFunctionSpec(
                       isPure: Boolean = false,
                       ) extends PSpecification
 
+case class PBodyParameterInfo(
+                           /** stores indices of parameters that are used as addressable variables in the code body */
```
ArquintL, line 695: `indices` seems a bit inappropriate as you're storing IdnUses

src/main/scala/viper/gobra/ast/frontend/PrettyPrinter.scala
```diff
@@ -113,6 +116,18 @@ class DefaultPrettyPrinter extends PrettyPrinter with kiama.output.PrettyPrinter
...
   }
 
+  def showBodyParameterInfoWithBlock(info: PBodyParameterInfo, block: PBlock): Doc = {
+    this.block(
+      showBodyParameterInfo(info) <> showStmtList(block.stmts)
+    )
+  }
+
+  def showBodyParameterInfo(info: PBodyParameterInfo): Doc = {
+    if (info.addressedParameters.isEmpty) {
```
ArquintL, line 126: should be `nonEmpty`

src/main/scala/viper/gobra/ast/frontend/PrettyPrinter.scala
```diff
@@ -206,7 +221,7 @@ class DefaultPrettyPrinter extends PrettyPrinter with kiama.output.PrettyPrinter
       case PContinue(label) => "continue" <> opt(label)(space <> showLabel(_))
       case PGoto(label) => "goto" <+> showLabel(label)
       case PDeferStmt(exp) => "defer" <+> showExpr(exp)
-      case PBlock(stmts) => block(showStmtList(stmts))
+      case n: PBlock => block(showStmtList(n.stmts))
```
ArquintL, line 224: this change seems unnecessary

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -490,6 +492,16 @@ object Parser {
...
       "{" ~> repsep(statement, eos) <~ eos.? <~ "}" ^^ PBlock
 
+    lazy val blockWithoutBraces: Parser[PBlock] =
+      repsep(statement, eos) <~ eos.? ^^ PBlock
+
+    lazy val blockWithBodyParameterInfo: Parser[(PBodyParameterInfo, PBlock)] =
+      "{" ~> bodyParameterInfo ~ blockWithoutBraces <~ "}"
+
+    lazy val bodyParameterInfo: Parser[PBodyParameterInfo] =
+      Constants.SHARE_PARAMETER_KEYWORD ~> repsep(idnUse, ",") <~ eos.? ^^ PBodyParameterInfo |
```
ArquintL, line 502: Share**d**?

src/main/scala/viper/gobra/frontend/info/implementation/property/Addressability.scala
```diff
@@ -102,4 +102,17 @@ trait Addressability extends BaseProperty { this: TypeInfoImpl =>
       case c: Constant => AddrMod.constant
       case e => Violation.violation(s"Expected variable, but got $e")
     }}
+
+  /** a parameter can be used as shared if it is included in the shared clause of the enclosing function or method */
+  lazy val isParameterUsedAsShared: PParameter => Boolean =
```
ArquintL, line 107: would `isParameterShared` be better?

Felalolf:
Currently, one test is failing. However, the test is not failing only sometimes on my machine. So something weird is going on.

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -691,6 +691,11 @@ case class PFunctionSpec(
...
+                           addressedParameters: Vector[PIdnUse]
```
ArquintL, line 696: Imho `addressableParameters` or `sharedParameters` would be better

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GoifyingPrinter.scala
```diff
@@ -124,6 +140,11 @@ class GoifyingPrinter(info: TypeInfoImpl) extends DefaultPrettyPrinter {
     case m => super.showMember(m)
   }
 
+  override def showBodyParameterInfo(info: PBodyParameterInfo): Doc = {
+    if (info.addressedParameters.isEmpty) {
```
ArquintL, line 144: again, should be `nonEmpty`

src/main/scala/viper/gobra/util/Constants.scala
```diff
@@ -0,0 +1,13 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2011-2020 ETH Zurich.
+
+package viper.gobra.util
+
+object Constants {
+
+  val ADDRESSABILITY_MODIFIER = "@"
+  val SHARE_PARAMETER_KEYWORD = "share"
```
ArquintL, line 12: I'm surprise to see `share` here, I expected `shared`. If it is indeed `share` then ignore my previous comment about `SHARE_PARAMETER_KEYWORD`

ArquintL (approved):

src/main/scala/viper/gobra/ast/frontend/Ast.scala
Felalolf: you are right. Before I used a set on integers.

src/main/scala/viper/gobra/frontend/Parser.scala
Felalolf: What do you mean?

src/main/scala/viper/gobra/frontend/info/implementation/property/Addressability.scala
Felalolf: I was unsure about the name. A parameter cannot be shared, but you can declare a shared copy of it for the loop. That is why I would be against `isParameterShared`, but I am open for other suggestions.

src/main/scala/viper/gobra/frontend/Parser.scala
ArquintL: I thought the keyword is `shared` and hence the constant should be named `SHARED_PARAMETER_KEYWORD`

src/main/scala/viper/gobra/ast/frontend/Ast.scala
Felalolf: For me, this is the same as for `isParaemterShared`. The parameter itself is not shared.

src/main/scala/viper/gobra/util/Constants.scala
Felalolf: I theory we could have general `share` and `unshare` statements later.

src/main/scala/viper/gobra/frontend/info/implementation/property/Addressability.scala
ArquintL: I see. `isParameterUsedAsShared` sounds for me more like that it is indeed used as shared (more than just being declared as such). Hence, my favorite so far would be `canParameterBeUsedAsShared`

Felalolf: I will change it to that.

src/main/scala/viper/gobra/ast/frontend/Ast.scala
Felalolf: I will change it to `addressableParameters`

Felalolf: Or how about `sharableParameters`?

ArquintL: Then we at least stick to the same terminology. But add a comment that this means that these params have been declared as shared in the body (because sharable doesn‘t sound at all like some declaration things)

Felalolf: Is the new name and description fine for you?

ArquintL: Yes

PR #115, opened 2020-10-30
Restrictions for Pure Members

### LINKED ISSUE #114: Ressources in postcondition of pure function is not rejected.
The following code exhibits different behaviour when run in the testsuite or when run as a single verification.
Furthermore, when run as a single verification, the verification goes through, but does not fail even though the verification should fail.

```
package pkg;

type coordinate struct{
    x, y, val int
}

pred cooMem(self *coordinate) {
    acc(self.x) && acc(self.y) && acc(self.val)
}


requires cooMem(self)
ensures  cooMem(self)
pure func (self *coordinate) value() int {
    return unfolding cooMem(self) in self.val
}
```

ArquintL:
Adds a check for pure methods and functions that the postconditions are pure

Fixes #114

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostMemberTyping.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostMemberTyping.scala
@@ -50,14 +50,14 @@ trait GhostMemberTyping extends BaseTyping { this: TypeInfoImpl =>
     member.body match {
       case Some((_, b: PBlock)) => isPureBlock(b)
       case None => noMessages
-      case Some(b) => message(member, s"For now the body of a pure method or pure function is expected to be a single return with a pure expression, got $b instead")
+      case Some(b) => message(member, s"For now, the body of a pure method or pure function is expected to be a single return with a pure expression, got $b instead")
     }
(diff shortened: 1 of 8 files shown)


src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostMemberTyping.scala
```diff
@@ -28,30 +28,38 @@ trait GhostMemberTyping extends BaseTyping { this: TypeInfoImpl =>
...
 
+  private def isSingleResultArg(member: PCodeRootWithResult): Messages = {
+    message(member, "For now, pure methods and pure functions must have exactly one result argument", member.result.outs.size != 1)
+  }
+
+  private def isSinglePureReturnExpr(member: PBodyMember): Messages = {
+    member.body match {
+      case Some((_, b: PBlock)) => isPureBlock(b)
+      case None => noMessages
+      case Some(b) => message(member, s"For now the body of a pure method or pure function is expected to be a single return with a pure expression, got $b instead")
```
Felalolf, line 53: Comma after "for now"

Felalolf (approved):
Nice!

PR #116, opened 2020-10-30
Slices

wytseoortwijn:
Basic support for slices.


Felalolf:
can you please merge option before I look at this PR?

Felalolf:
Could you also merge with master please

Felalolf:
Since the permission and shared annotation syntax had changed, you might have to change your test cases, as well.
I used the following Python code to help me with that:
1) For shared annotation:
 ```
for fn in files:
  for line in fileinput.input(fn, inplace=True):
    line = re.sub(r'(_[a-zA-Z0-9_]+|[a-zA-Z][a-zA-Z0-9_]*)!', r'\1@', line.rstrip())
    print(line)
```
2) For Accessebility Predicate:
```
for fn in files:
  for line in fileinput.input(fn, inplace=True):
    line = re.sub(r'(acc\s*\(\s*)([^\*\s])', r'\1&\2', line.rstrip())
    line = re.sub(r'(acc\s*\(\s*)\*', r'\1', line.rstrip())
    print(line)
```

There is nothing smart there. Just in case this might save you two google searches.

File: src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala [hunk 1]
```diff
@@ -0,0 +1,461 @@
...
+        if (generateFunctionBodies) Some(body) else None
+      )()
+    }
+  }
+
+  /**
+    * Definition of the "ssliceFromArray" Viper function supporting the
+    * Viper domain of Slices. This function is used in the translation
+    * of array slicing expressions "a[i:j]" in Gobra (without a third
+    * slicing index). Its Viper definition is:
```
Felalolf, line 350: "without a third slicing index" -> "without a capacity argument"

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -0,0 +1,461 @@
...
+  private def fullSliceFromArray(typ : vpr.Type, base : vpr.Exp, i : vpr.Exp, j : vpr.Exp, k : vpr.Exp)(ctx : Context)(pos : vpr.Position = vpr.NoPosition, info : vpr.Info = vpr.NoInfo, errT : vpr.ErrorTrafo = vpr.NoTrafos) : vpr.FuncApp =
+    fullSliceFromArrayGenerator(Vector(base, i, j, k), typ)(pos, info, errT)(ctx)
+
+  /** An application of the "sfullSliceFromSlice[`typ`](...)" Viper function. */
+  private def fullSliceFromSlice(typ : vpr.Type, base : vpr.Exp, i : vpr.Exp, j : vpr.Exp, k : vpr.Exp)(ctx : Context)(pos : vpr.Position = vpr.NoPosition, info : vpr.Info = vpr.NoInfo, errT : vpr.ErrorTrafo = vpr.NoTrafos) : vpr.FuncApp =
+    fullSliceFromSliceGenerator(Vector(base, i, j, k), typ)(pos, info, errT)(ctx)
+
+  /** Gives the 'nil' slice of type `typ`. */
+  private def nilSlice(typ : vpr.Type)(ctx : Context)(pos : vpr.Position, info : vpr.Info, errT : vpr.ErrorTrafo) : vpr.Exp =
+    ctx.option.none(ctx.slice.typ(typ))(pos, info, errT)
```
Felalolf, line 157: Can your encoding of the nil slice derive that a slice with permissions is not nil?
E.g. the following code should verify:
```
requies acc(&x[0])
func foo(x []int) {
  assert x != nil
}
```

Felalolf (commented):
I did not look through all the tests. I trust that what you did makes sense there.

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -987,7 +993,8 @@ object Desugar {
             case dop : in.ArrayLit => in.IntLit(dop.length)(src)
             case _ => dop.typ match {
               case _: in.ArrayT => in.Length(dop)(src)
```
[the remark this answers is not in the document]
wytseoortwijn: Yes, `len(e)` and `cap(e)` are both `n` if `e : [n]T`.

(https://golang.org/ref/spec#Length_and_capacity)

src/main/scala/viper/gobra/translator/implementations/components/SlicesImpl.scala
```diff
@@ -0,0 +1,276 @@
...
+    *
+    *   axiom slice_len_leq_cap {
+    *     forall s : Slice[T] :: { slen(s) } { scap(s) } slen(s) <= scap(s)
+    *   }
+    *
+    *   axiom slice_cap_leq_alen {
+    *     forall s : Slice[T] :: { soffset(s), scap(s) } { alen(sarray(s)) }
+    *       soffset(s) + scap(s) <= alen(sarray(s))
+    *   }
+    * }
```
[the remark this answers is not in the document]
wytseoortwijn: Slice index expressions `s[i]` are translated into `ShArrayloc(sarray(s), i')` (with `i'` calculated from `i` and the info stored in `s`), and `ShArrayloc` is injective. My idea is thus to lean on the injectivity on `ShArrayloc`.

So for any two slices `s1` and `s2` that overlap in locations (i.e., `ShArrayloc(sarray(s1), i1) = ShArrayloc(sarray(s2), i2)` for some `i1` and `i2`) we have that `sarray(s1) = sarray(s2)` and `i1 = i2`.

In other words, any two slices either have the exact same underlying array, or two completely disjoint underlying arrays.

(One of my previous encodings used and axiomatised an extra domain function `sslot(s, i)`, in addition to `ShArrayloc`, to translate indexing operations to, and this was problematic due to injectivity.)

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -0,0 +1,461 @@
...
+    *   ensures soffset(result) == i
+    *   ensures slen(result) == j - i
+    *   ensures scap(result) == k - i
+    *   ensures sarray(result) == a
+    * {
+    *   sconstruct(a, i, j - i, k - i)
+    * }
+    * }}}
+    */
+  private val fullSliceFromArrayGenerator : FunctionGenerator[vpr.Type] = new FunctionGenerator[vpr.Type] {
```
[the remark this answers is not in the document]
wytseoortwijn: I could certainly do that.

The Go language spec refers to slicing expressions of the form `s[i:j:k]` as "full slice expressions", and `s[i:j]` as "simple slice expressions" (https://golang.org/ref/spec#Slice_expressions). That is why I prefixed with "full".

src/main/scala/viper/gobra/translator/implementations/components/ArraysImpl.scala
```diff
@@ -6,132 +6,116 @@
...
 import viper.gobra.translator.interfaces.Collector
 import viper.gobra.translator.interfaces.components.Arrays
-import viper.silver.ast.DomainFuncApp
 import viper.silver.{ast => vpr}
 
 /**
-  * Contains functionality for generating and using the "Array" Viper domain.
+  * A specialized version of the array domain, where the element type is a parameter.
   */
 class ArraysImpl extends Arrays {
```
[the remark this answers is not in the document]
wytseoortwijn: Yes, I actually just discarded the old Array component that was still lingering around, and replaced it with the one you wrote (relocated actually).

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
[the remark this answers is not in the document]
Felalolf: I see, then fullSliceFromArrayGenerator is fine for me, as well.

src/main/scala/viper/gobra/translator/implementations/components/SlicesImpl.scala
```diff
@@ -0,0 +1,276 @@
...
+    )(pos, info, errT)
+  }
+
+  /**
+    * A function application of the "sloc" function.
+    */
+  override def loc(base : vpr.Exp, index : vpr.Exp, t : vpr.Type)(pos : vpr.Position, info : vpr.Info, errT : vpr.ErrorTrafo) : vpr.Exp = {
+    arrays.loc(
+      array(base, t)(pos, info, errT),
+      add(offset(base)(pos, info, errT), index)(pos, info, errT)
```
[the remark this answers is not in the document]
wytseoortwijn: This `add` function allows defining a trigger for `forall i int :: 0 <= i && i < cap(s) ==> acc(&s[i])`. The Viper translation would be something like:

```viper
forall i : Int :: { ShArrayloc(sarray(optGet(s)), sadd(soffset(optGet(s)), i)) }
  0 <= i && i < (s == optNone() ? 0 : scap(optGet(s))) ==>
    acc(ShArrayloc(sarray(optGet(s)), sadd(soffset(optGet(s)), i)).val, write)
```

If such a trigger is not specified, then Gobra has problems proving that there is permission for reading `s[0]` in, for example, the following program:

```go
requires forall i int :: 0 <= i && i < cap(s) ==> acc(&s[i])
func foo(s []int) {
  n := s[0]
}
```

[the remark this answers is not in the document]
Felalolf: I see, I searched for add in triggers, but I forgot that obviously add is in a trigger when loc is in a trigger. Makes sense.

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -0,0 +1,461 @@
...
+        lowT <- goE(low)
+        highT <- goE(high)
+        maxOptT <- option(max match {
+          case Some(e) => Some(goE(e))
+          case None => None
+        })
+      } yield {
+        val (pos, info, errT) = exp.vprMeta
+        val typT = ctx.typeEncoding.typ(ctx)(typ)
+        val baseOptT = ctx.option.get(baseT, ctx.slice.typ(typT))(pos, info, errT)
```
Felalolf, line 107: What happens when the slice is nil?

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -0,0 +1,461 @@
...
+  }
+
+  override def reference(ctx : Context) : in.Location ==> CodeWriter[vpr.Exp] = default(super.reference(ctx)) {
+    case (exp @ in.IndexedExp(base :: ctx.Slice(typ), idx)) :: _ / Shared => for {
+      baseT <- ctx.expr.translate(base)(ctx)
+      idxT <- ctx.expr.translate(idx)(ctx)
+      typT = ctx.typeEncoding.typ(ctx)(typ)
+    } yield {
+      val (pos, info, errT) = exp.vprMeta
+      val getT = ctx.option.get(baseT, ctx.slice.typ(typT))(pos, info, errT)
```
Felalolf, line 138: What happens if the slice is nil? Also, what happens with out of bounds access?

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -0,0 +1,461 @@
...
+      }
+
+      case in.Length(exp :: ctx.Slice(typ)) => for {
+        expT <- goE(exp)
+        typT = ctx.typeEncoding.typ(ctx)(typ)
+        (pos, info, errT) = exp.vprMeta
+      } yield vpr.CondExp(
+        vpr.EqCmp(expT, nilSlice(typT)(ctx)(pos, info, errT))(pos, info, errT),
+        vpr.IntLit(0)(pos, info, errT),
+        ctx.slice.len(ctx.option.get(expT, ctx.slice.typ(typT))(pos, info, errT))(pos, info, errT)
```
Felalolf, line 65: What happens when the slice is nil?

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -0,0 +1,461 @@
...
+      )(pos, info, errT)
+
+      case in.Capacity(exp :: ctx.Slice(typ)) => for {
+        expT <- goE(exp)
+        typT = ctx.typeEncoding.typ(ctx)(typ)
+        (pos, info, errT) = exp.vprMeta
+      } yield vpr.CondExp(
+        vpr.EqCmp(expT, nilSlice(typT)(ctx)(pos, info, errT))(pos, info, errT),
+        vpr.IntLit(0)(pos, info, errT),
+        ctx.slice.cap(ctx.option.get(expT, ctx.slice.typ(typT))(pos, info, errT))(pos, info, errT)
```
Felalolf, line 75: What happens when the slice is nil?

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
wytseoortwijn: No. Slices are encoded as option types. And given any expression `e : Option[T]`, assuming something about `optGet(e)` does not imply that `e != optNone()`. For example:

```viper
method m(o : Option[Int])
{
  assert 0 < optGet(o) ==> o != optNone() // does not verify
}
```

That is the reason why your code snippet also will not verify.

(This is exactly what I tried to solve in my encoding of Option types which appeared unsound...)

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
wytseoortwijn: Then the generated Viper expression would be equivalent to `0`.

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
wytseoortwijn: Then the generated Viper expression would be equivalent to `0`.

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
wytseoortwijn: There are some problems here I see...

This for example currently doesn't verify but should:

```go
requires s == nil
func foo(s []int) {
  assert s[:] == nil
}
```
And this would actually crash Gobra:
```go
requires s == nil
func foo(s []int) {
  assert s[1:] == nil
}
```

I'll have a look at that, thanks!

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
wytseoortwijn: Hmm, but I could of course translate `acc(&s[i])` into something like `!isNone(s) && acc(...).val)`. What do you think of that?

Felalolf: The whole nil business is quite complex. That is also why I intentionally did not pick option for shared arrays. 
Maybe you could reuse the nil from array. There I made it such that only a single representation of null exists, which is an array of size 1 with all nil entries. The size embeddings preserve nil, i.e. they do not enforce their respective size constraint as this would violate the uniqueness of nil (which is inconsistent together with injectivity). Furthermore I adapted that loc is only injective wihtin its bound (thus forcing that there is a single nil).

If you want to keep the options, then maybe an axiom along the lines of the following example might help:
```
Slice[T] {
  ....
  axiom {
    forall s: Option[Slice[T]], i: Int ::     loc(sarray(get(s)), i) != (nil(); T) ==> s != none()
  }
}
```
This would require axiomatizing all our nil representations, which is a bit slow, but possible, e.g.
```
domain Nil[T] {
  function nil(): T

  axiom { (nil(): Ref) == null }
  axiom { forall i: Int :: {loc((nil(): Array[T]), i)} loc((nil(): Array[T]), i) == (nil(): T) }
  axiom { (nil(): Emb17[T]) == box17((nil(): Array[T])) && unbox17((nil(): Emb17[T])) == (nil(): Array[T]) }
  axiom { (nil(): Emb42[T]) == box42((nil(): Array[T])) && unbox42((nil(): Emb42[T])) == (nil(): Array[T]) }
  axiom { (nil(): Emb77[T]) == box77((nil(): Array[T])) && unbox77((nil(): Emb77[T])) == (nil(): Array[T]) }
}
```

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
Felalolf: The first case is very interesting. If you are reusing my nil for shared arryas, then we should change that nil has a size of zero (currently it is a size of 1). Indeed, calling len and cap on a nil slice does return 0.

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
Felalolf: I missed the conditional

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
wytseoortwijn: Yes, having such a different representation for `nil` slices could very well work, thanks!

I considered alternatives, but picked option types since they allow straightforward encoding of e.g. `len(s)` and `cap(s)` in case `s = nil`. But perhaps option types aren't the best choice after all...

I think the `nil` slice can alternatively be represented as a slice that has the `nil` array as its underlying array, and appropriate offset/len/cap values. For example `sconstruct(nilArray(), 0, 1, 1)`. Such a slice would be unique.

Felalolf: I think length and cap also have to be 0. I wrote something about this further down.

wytseoortwijn: Yes, I agree

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
wytseoortwijn: Why? One could define the nil slice to be equivalent to `nilArray[0:0:0]`, even when `len(nilArray) = 1`.

wytseoortwijn: Aah wait, I think I see why. Then the nil array could potentially be sliced to have one element.

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
Felalolf: I should also point out the following:
Go distinguishes between a slice of nil and a slice of an array of size 0, i.e. `var x []int = nil; var y [0]int; assert x[:] != y[:]`

wytseoortwijn: Hmm, yes. Go wouldn't accept that, since slices are in principle incomparable. But on the (Viper) encoding level these should indeed be different.

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
Felalolf: You could prevent this with an additional check in the encoding of the slice expression. 
I am not entirely sure myself if nil should have size 1 or 0. 
When you say that nil has size 0, then it hard to distinguish a zero sized shared array and nil.
I think that this is also the reason why I picked size 1 for nil.

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
Felalolf: Could we quickly have a zoom meeting?

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
wytseoortwijn: With the current way default arrays are encoded, I don't think they can be of length 0, can they?

If the encoding would be
```viper
function arrayDefault(): ShArray[Ref]
  ensures (ShArraylen(result): Int) == 0
  ensures (forall i: Int :: { ShArrayloc(result, i) } ShArrayloc(result, i) == null)
```
Then the `nil` slice might no longer be unique.

Felalolf: I do not entirely understand what you mean. Why would nil not be unique anymore. Because of arrays of size 0? I think that this depends on the equality relation. Since we do not have partial functions, two arrays are different even if their size is 0, but their elements are different. So as long as 0-sized arrays do not have array locations that are nil (which every non-nil array must satisfy), we should be fine.

wytseoortwijn: What I meant is that, for any two arrays `a1` and `a2`, both of length 0 and both having `null` at every location, one cannot prove that `a1 == a2`, right? And this can be proven for any length greater than 0.

wytseoortwijn: Perhaps the way the equality relation is defined could resolve that. But I don't yet see how.

Could we not change the injectivity axiom for arrays to be the following (i.e., dropping the bounds)?
```viper
axiom {
  forall a : Array[T], i : Int :: { loc(a, i) } fst(loc(a, i)) == a && snd(loc(a, i)) == i
}
```

Felalolf: The purpose of the bounds is to prevent an unsoundness in combination with nil.
Regarding nil, the property that we want is that if an array location is nil, then the array is nil.
This is equivalent to that 1) the nil array has nil locations for every index and 2) only the nil array has nil entries.
However, this violates injectivity of loc, since a nil location must identify the corresponding array and the corresponding index. 
I saw two solutions to solve this: 1) make injectivity hold for all arrays that are not nil or 2) make injectivity hold only on valid indexes and give nil either size 1 or 0.

src/main/scala/viper/gobra/ast/internal/utility/GobraStrategy.scala
```diff
@@ -105,13 +105,15 @@ object GobraStrategy {
       case (_: OptionNone, Seq(t : Type)) => OptionNone(t)(meta)
       case (_: OptionSome, Seq(op : Expr)) => OptionSome(op)(meta)
       case (_: OptionGet, Seq(op : Expr)) => OptionGet(op)(meta)
+      case (_: Slice, Seq(base : Expr, low : Expr, high : Expr, max : Option[Expr@unchecked])) => Slice(base, low, high, max)(meta)
```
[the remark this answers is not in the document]
wytseoortwijn: Fixed

src/main/scala/viper/gobra/ast/internal/utility/Nodes.scala
```diff
@@ -76,6 +76,7 @@ object Nodes {
         case StructUpdate(base, field, newVal) => Seq(base, field, newVal)
         case IndexedExp(base, idx) => Seq(base, idx)
         case ArrayUpdate(base, left, right) => Seq(base, left, right)
+        case Slice(base, low, high, max) => Seq(base, low, high) ++ max
```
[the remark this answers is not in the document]
wytseoortwijn: Fixed

src/main/scala/viper/gobra/translator/Names.scala
```diff
@@ -32,7 +32,13 @@ object Names {
   def embeddingUnboxFunc: String = "unbox"
 
   // pointer
-  def pointerField(t: vpr.Type) : String = s"val$$_$t"
+  def pointerField(t : vpr.Type) : String = {
+    val ts = t.toString()
+      .replace('[', '_')
+      .replace("]", "")
```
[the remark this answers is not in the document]
wytseoortwijn: Sure, I did that

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -0,0 +1,461 @@
...
+          sliceT <- ctx.expr.translate(in.Slice(tmp, in.IntLit(0)(lit.info), in.IntLit(litA.length)(lit.info), None)(lit.info))(ctx)
+          _ <- local(tmpT)
+          _ <- write(initT)
+          _ <- write(assignT)
+        } yield sliceT
+      }
+    }
+  }
+
+  override def reference(ctx : Context) : in.Location ==> CodeWriter[vpr.Exp] = default(super.reference(ctx)) {
```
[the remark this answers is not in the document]
wytseoortwijn: Fixed

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
wytseoortwijn: Done

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -128,16 +120,47 @@ class SliceEncoding(arrayEmb : SharedArrayEmbedding) extends LeafTypeEncoding {
...
   }
 
+  /**
+    * Encodes the comparison of two expressions.
+    *
+    * [lhs: []T == rhs: []T] ->
+    *   len([lhs]) == len([rhs]) &&
+    *   cap([lhs]) == cap([rhs]) &&
+    *   offset([lhs]) == offset([rhs]) &&
+    *   array([lhs]) == array([rhs])
```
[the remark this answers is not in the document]
wytseoortwijn: Go defines that slices are only comparable to `nil`. I think that already resolves it.

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -29,87 +30,78 @@ class SliceEncoding(arrayEmb : SharedArrayEmbedding) extends LeafTypeEncoding {
...
       }
 
-      case exp @ in.Slice((base : in.Expr) :: ctx.Slice(typ), low, high, max) => for {
+      case exp @ in.Slice((base : in.Expr) :: ctx.Slice(_), low, high, max) => for {
         baseT <- goE(base)
         lowT <- goE(low)
         highT <- goE(high)
         maxOptT <- option(max match {
           case Some(e) => Some(goE(e))
           case None => None
```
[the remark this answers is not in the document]
wytseoortwijn: Done

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -29,87 +30,78 @@ class SliceEncoding(arrayEmb : SharedArrayEmbedding) extends LeafTypeEncoding {
...
+      } yield withSrc(ctx.slice.cap(expT), exp)
 
       case exp @ in.Slice((base : in.Location) :: ctx.Array(_, _) / Shared, low, high, max) => for {
         baseT <- ctx.typeEncoding.reference(ctx)(base)
+        unboxedBaseT = arrayEmb.unbox(baseT, base.typ.asInstanceOf[in.ArrayT])(base)(ctx)
         lowT <- goE(low)
         highT <- goE(high)
         maxOptT <- option(max match {
           case Some(e) => Some(goE(e))
           case None => None
```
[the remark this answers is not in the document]
wytseoortwijn: Done, thanks!

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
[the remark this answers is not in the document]
Felalolf: Ah, so you cannot compare to slices?

Felalolf (approved):

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1141,6 +1151,13 @@ object Desugar {
           for { elemsD <- sequence(elems.map(e => compositeValD(ctx)(e, typ))) }
             yield in.ArrayLit(typ, elemsD)(src)
 
+        case CompositeKind.Slice(in.SliceT(typ, addressability)) =>
+          Violation.violation(addressability == Addressability.literal, "Literals have to be exclusive")
+          val indices = info.keyElementIndices(lit.elems)
+          val elems = lit.elems.zip(indices).map(e => (e._2, e._1.exp)).sortBy(_._1).map(_._2)
+          for { elemsD <- sequence(elems.map(e => compositeValD(ctx)(e, typ))) }
+            yield in.SliceLit(typ, elemsD)(src)
```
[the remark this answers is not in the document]
wytseoortwijn: Sure! I just made one

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
wytseoortwijn: A verification error with a suitable message will now be presented in case of an out-of-bounds access.

Nil slices should be handled correctly, as they are encoded roughly as a default array that is sliced to have capacity `0`.

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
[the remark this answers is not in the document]
wytseoortwijn: No. One could only write `s == nil` or `nil == s` with `s` a slice. Any expression `s1 == s2` would not be "well-formed" Go.

PR #117, opened 2020-10-30
New Access Predicate Design

Felalolf:
Adapts the syntax for access predicats as discussed in the meeting,
Now, access predicates take a predicate or a pointer.


src/main/scala/viper/gobra/frontend/info/implementation/property/Addressability.scala
```diff
@@ -60,7 +61,7 @@ trait Addressability extends BaseProperty { this: TypeInfoImpl =>
           case t => Violation.violation(s"Expected slice, array, map, or sequence, but got $t")
         }
       case n: PDot => resolve(n) match {
-        case Some(s: ap.FieldSelection) => AddrMod.fieldLookup(addressability(s.base))
+        case Some(s: ap.FieldSelection) => AddrMod.fieldLookup(addressabilityMemberPath(addressability(s.base), s.path))
```
ArquintL, line 64: why wasn't it necessary earlier to iterate over the path?

src/test/resources/regressions/examples/infinite_list.gobra
```diff
@@ -9,7 +9,7 @@ type node struct {
 };
 
 pred infList(headPtr *node) {
-  headPtr != nil && acc(headPtr.value) && acc(headPtr.next) && infList(headPtr.next)
+  headPtr != nil && acc(&headPtr.value) && acc(&headPtr.next) && infList(headPtr.next)
```
ArquintL, line 12: After seeing it in action, I agree that it makes more sense to expect an address in `acc(...)`

ArquintL (approved):

ArquintL:
> Now, access predicates take a predicate or a pointer.

This sounded at first like `acc(pred(...))` with `pred(...)` being a predicate call to me. But I guess you meant an access predicate can be a predicate call or `acc(p)` with p being of pointer type. After a quick look into the Viper tutorial, I couldn't find a term for the stuff inside an `acc(...)`.

src/main/scala/viper/gobra/frontend/info/implementation/property/Addressability.scala
Felalolf: That was a bug.

src/test/resources/regressions/examples/infinite_list.gobra
Felalolf: That is reassuring to hear.

PR #125, opened 2020-11-11
Fix issue 124

### LINKED ISSUE #124: Parse error when importing package containing the '.' character
A common pattern in *Go* is to have the package paths mirroring the repository structure of the project, such as
```go
import (
	"os"
	"github.com/scionproto/scion/go/dispatcher/dispatcher"
	"github.com/scionproto/scion/go/lib/common"
	"github.com/scionproto/scion/go/lib/log"
	"github.com/scionproto/scion/go/lib/sock/reliable"
)
```
I found that this leads to a parser error because the parser is not expecting the package description to contain a `.` character. For the minimal example
```go
package main
import "github.com/scionproto/scion/go/lib/addr"

func main() {
    x := addr.F()
}
```
Gobra produces the error message
```
[info] Gobra has found 1 error(s):
[info] 	<2:15> src/test/resources/regressions/issues/test.gobra:2:15: '"' expected but '.' found
[info] import "github.com/scionproto/scion/go/lib/addr";
```

jcp19:
Closes #124 by extending the import path parser to read the `.` character.


src/test/resources/regressions/issues/000124.gobra
```diff
@@ -0,0 +1,11 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package main
+
+// ##(-I src/test/resources/regressions/issues/issue124/)
```
ArquintL, line 6: What do you think? Would `000124` be a better folder name? This way, the folder should appear next to `000124.gobra`

src/test/resources/regressions/issues/000124.gobra
```diff
@@ -0,0 +1,11 @@
...
+import gobra "github.com/gobra"
```
ArquintL, line 7: Could you add a second test which tests whether `gobra` is inferred from the import path (i.e. by using `import "github.com/gobra"`)

ArquintL (approved):

src/test/resources/regressions/issues/000124.gobra
jcp19: Yes, that is a better name indeed!

PR #131, opened 2020-11-24
Fix issue 128

### LINKED ISSUE #128: Problems with conversions between types
The following snippet compiles in *Go*
```go
package main

func main() {
	y := uint16(1)
	x := uint32(y)
	// ...
}
```
but leads to an error in *Gobra*
```
Gobra has found 1 error(s):
<7:7> /Users/joao/Desktop/conversions.gobra:7:7: property error: IntT(IntegerKind(uint16)) is not convertible to IntT(IntegerKind(uint32))
x := uint32(y);
```

jcp19:
Fixes #128. This PR adds a new type of expression (`Conversion(newType, expr)`) to the internal language, representing a conversion ```newType(expr)``` at the surface language level.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/implementation/property/Convertibility.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/property/Convertibility.scala
@@ -6,7 +6,7 @@
 
 package viper.gobra.frontend.info.implementation.property
 
-import viper.gobra.frontend.info.base.Type.{DeclaredT, PointerT, Single, Type}
+import viper.gobra.frontend.info.base.Type.{DeclaredT, IntT, PointerT, Single, Type}
 import viper.gobra.frontend.info.implementation.TypeInfoImpl
 
 trait Convertibility extends BaseProperty { this: TypeInfoImpl =>
@@ -18,6 +18,7 @@ trait Convertibility extends BaseProperty { this: TypeInfoImpl =>
   } {
     case (Single(lst), Single(rst)) => (lst, rst) match {
       case (left, right) if assignableTo(left, right) => true
+      case (IntT(_), IntT(_)) => true
       case (left, right) => (underlyingType(left), underlyingType(right)) match {
         case (l, r) if identicalTypes(l, r) => true
         case (PointerT(l), PointerT(r)) if identicalTypes(underlyingType(l), underlyingType(r)) &&
--- a/src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
@@ -135,7 +135,7 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
       case (Right(_), Some(p: ap.Conversion)) => // requires single argument and the expression has to be convertible to target type
         val msgs = message(n, "expected a single argument", p.arg.size != 1)
         if (msgs.nonEmpty) msgs
-        else convertibleTo.errors(exprType(p.arg.head), typeType(p.typ))(n) ++ isExpr(p.arg.head).out
+        else convertibleTo.errors(typeType(p.typ), exprType(p.arg.head))(n) ++ isExpr(p.arg.head).out
 
       case (Left(callee), Some(p: ap.FunctionCall)) => // arguments have to be assignable to function
         exprType(callee) match {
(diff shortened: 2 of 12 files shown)


src/main/scala/viper/gobra/frontend/info/implementation/property/Convertibility.scala
```diff
@@ -18,6 +18,7 @@ trait Convertibility extends BaseProperty { this: TypeInfoImpl =>
   } {
     case (Single(lst), Single(rst)) => (lst, rst) match {
       case (left, right) if assignableTo(left, right) => true
+      case (IntT(_), IntT(_)) => true
```
Felalolf, line 21: I am sure you checked this, but just to make sure: In Go, any integer kind can be converted to every integer kind?

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -135,7 +135,7 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
       case (Right(_), Some(p: ap.Conversion)) => // requires single argument and the expression has to be convertible to target type
         val msgs = message(n, "expected a single argument", p.arg.size != 1)
         if (msgs.nonEmpty) msgs
-        else convertibleTo.errors(exprType(p.arg.head), typeType(p.typ))(n) ++ isExpr(p.arg.head).out
+        else convertibleTo.errors(typeType(p.typ), exprType(p.arg.head))(n) ++ isExpr(p.arg.head).out
```
Felalolf, line 138: The order seems wrong to me. The type of the expression should be convertable to the argument type.

src/main/scala/viper/gobra/translator/implementations/translator/ExpressionsImpl.scala
```diff
@@ -95,6 +95,8 @@ class ExpressionsImpl extends Expressions {
         case errors => Violation.violation(s"invalid trigger pattern (${errors.head.readableMessage})")
       }
 
+      case in.Conversion(typ, expr) => goE(expr)
```
Felalolf, line 98: For which types does this work? The typechecker should check that only those supported types are used with conversion expressions.

src/main/scala/viper/gobra/frontend/info/implementation/property/Convertibility.scala
jcp19: The go language spec states that
>  A non-constant value x can be converted to type T in any of these cases:
    x is assignable to T.
    ignoring struct tags (see below), x's type and T have identical underlying types.
    ignoring struct tags (see below), x's type and T are pointer types that are not defined types, and their pointer base types have identical underlying types.
    **x's type and T are both integer or floating point types.**
    x's type and T are both complex types.
    x is an integer or a slice of bytes or runes and T is a string type.
    x is a string and T is a slice of bytes or runes.

In particular, there is no constraint on the types of integers allowed for conversion. One thing that I should note however is that this is valid for a non-constant value x. The well-defidedness of a conversion of a constant expression x to a type T relies on a different criterion however
> A constant value x can be converted to type T if x is representable by a value of T.

Besides,
> **A constant x is representable by a value of type T if one of the following conditions applies:
    x is in the set of values determined by T.**
    (other clauses not relevant for the discussion)

This latter case (for constants) is not implemented in Gobra because `Convertibility` is implemented as a binary relation between types instead of being implemented as a relation between an expression and a type to which the expression is converted to. This simplification is consistent with the one made for the `Assignability` property, which is also a relation between a value x and a type according to the Go lang. spec. but is implemented in Gobra as a `Property[(Type, Type)]`

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
jcp19: I believe you are right, I miss-interpreted the property signature

src/main/scala/viper/gobra/translator/implementations/translator/ExpressionsImpl.scala
Felalolf: As stated in the meeting, I am not sure that conversions on structs are properly supported.

jcp19: I think that they are supported. According to the Go spec, if two types have the same underlying type (possibly differing only on struct tags, which are not supported in Gobra at the moment) then they can be converted to each other. This is implemented in the convertibility property as follows
https://github.com/viperproject/gobra/blob/c17026c340cc041e7fd71bb50566b2725234952f/src/main/scala/viper/gobra/frontend/info/implementation/property/Convertibility.scala#L19-L22
Both structures are encoded in the same way at the viper level, as tuples, and field accesses are encoded as tuple accesses. Given that two struct types are only equal if they have the same fields, of the same types, and in the same order, field accesses at the viper level are always well typed.

I added a test for struct types, as suggested by you :)

Felalolf (approved):

PR #132, opened 2020-11-24
Fix for issue 129

wytseoortwijn:


src/test/resources/regressions/features/arrays/array-literal-fail2.gobra
```diff
@@ -1,9 +0,0 @@
-// Any copyright is dedicated to the Public Domain.
-// http://creativecommons.org/publicdomain/zero/1.0/
-
-package pkg
-
-func foo() {
-	//:: ExpectedOutput(type_error)
-	a := [2]int{42}
```
Felalolf, line 8: Is this now  a supported literal?

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1144,25 +1144,24 @@ object Desugar {
...
-          val elems = lit.elems.zip(indices).map(e => (e._2, e._1.exp)).sortBy(_._1).map(_._2)
-          for { elemsD <- sequence(elems.map(e => compositeValD(ctx)(e, typ))) }
-            yield in.ArrayLit(typ, elemsD)(src)
+          for {
+            elemsD <- sequence(lit.elems.map(e => compositeValD(ctx)(e.exp, typ)))
+          } yield {
+            val mappedElemsD = info.keyElementIndices(lit.elems).zip(elemsD).toMap
+            val supplementedElemsD = Range.BigInt(0, len, 1).map(i => mappedElemsD.getOrElse(i, in.DfltVal(typ)(Source.Parser.Internal))).toVector
+            in.ArrayLit(typ, supplementedElemsD)(src)
+          }
```
Felalolf, line 1155: Why are you not using keyedElementsD here?

src/test/resources/regressions/features/arrays/array-literal-fail2.gobra
wytseoortwijn: Yes, `[2]int{42}` would be the same as `[2]int {42, 0}`.

src/main/scala/viper/gobra/frontend/Desugar.scala
wytseoortwijn: Because array literals always have a fixed length, whereas the slice/sequence literals should be derived from the keyed elements.

For example, `[4]int { 2:12 }` is equivalent to `[4]int { 0, 0, 12, 0 }`, whereas `seq[int] { 2:12 }` is equivalent to `seq[int] { 0, 0, 12 }` (and the same for slices).

wytseoortwijn: One downside of this fix, is that for example `seq[int] { 100000:12 }` would generate a huge `Seq(...)` literal on the level of Viper...

Would you say that this is problematic? I don't expect such situations to some up very often.

Felalolf: I do think that this is problematic. It would be nicer to handle this at the level of the encoding.

Felalolf: However, you can also let it be for now. We could also discuss this in the gobra meeting.

Felalolf (approved):

wytseoortwijn:
I tried several things to improve the translation of big sequence/array/slice literals, as discussed earlier.

Generating abstract functions for translating literals would not really work, e.g. since generating an abstract function to represent the following literal would already be tricky (since something needs to be done for `a` and `b`):

```go
func foo (a int, b int) {
  ghost xs := seq[int] { a, b }
}
```

I tried representing sequence literals using a chain of sequence updates, for example by translating `[4]int{ 3:12, 2:8 }` into something like `emptySeqOfSize(4)[3 := 12][2 := 8]`. But that did not really work, since the axiomatisation of sequence (updates) didn't seem strong enough.

My current proposal is to explicitly generate large literals, and abbreviate large gaps between keys. For example the Gobra sequence `seq[int] { 2, 1, 5:3, 124:5, 127:9, 1000:12 }` now translates into `Seq(2, 1, 0, 0, 0, 3) ++ emptySeqOfSize(118) ++ Seq(5, 0, 0, 9) ++ emptySeqOfSize(872) ++ Seq(12)`.

Felalolf:
Another approach would be the approach discussed at the meeting. For these sparse literals, we generate a function whose postcondition defines the values. For the example `seq[int] { 2, 1, 5:3, 124:5, 127:9, 1000:12 }`, this results in something along the lines of:
```
function seqLit$1(): Seq[Int]
  ensures |result| == 1000
  ensures result[0] == 2 && result[1] == 1 && result[5] == 3 && result[124] == 5 && result[127] == 9 && result[1000] == 12
  ensures forall i: Int :: {result[i]} !(i in Set(0, 1, 5, 124, 127, 1000)) ==> result[i] == 0 
```

wytseoortwijn:
Yes, I tried that, but encountered tricky situations, such as for example:

```go
requires xs == seq[int] { 12:a, 1000:b }
func foo (a int, b int, ghost xs seq[int]) {
}
```

How would you generate a `seqLit$1` function for the literal in the precondition?

Felalolf:
I think that your other example is wrong. You meant something like `seq[int] { a: x, b: y, c: z }`. However, as I see it, the problem is to figure out the length of the sequence. But this is a problem regardless of the approach. You could do something as follows:
```
function seqLit$2(a: Int, b: Int, c: Int, x: Int, y: Int, z: Int): Seq[Int]
  requires |Set(a,b,c)| == 3 // to guarantee that the indeces are distinct.
  ensures a <= |result| && b <= |result| && c <= |result|
  ensures result[a] == x && result[b] == y && result[c] == z
  ensures forall i: Int :: {result[i]}  0 <= i && i < |result|  && !(i in Set(a, b, c)) ==> result[i] == 0 
```

wytseoortwijn:
Which example do you mean? I don't yet see how any of them are wrong.

I think that determining the length of a (sequence) literal does not necessarily have to be a problem, since all keys have to be constant and non-negative.

Felalolf:
I meant the example `seq[int] { a, b }`. That said, also for the example below I am not sure what you mean with tricky.
```
requires xs == seq[int] { 12:a, 1000:b }
func foo (a int, b int, ghost xs seq[int]) {
}
```

I would just encode this as something like:
```
function seqLit$3(a: Int, b: Int): Seq[Int]
  ensures 1000 == |result|
  ensures result[12] == a && result[1000] == b
  ensures forall i: Int :: {result[i]}  0 <= i && i < |result|  && !(i in Set(12, 1000)) ==> result[i] == 0 

method foo(a: int, b; Int, xs: Seq[Int])
  requires seqLit$3(a,b) == xs 
{}
```
Do you mean with tricky the free variables? As in the example, free variables have to be collected and used as parameters.

Felalolf:
Also, "wrong" is the inorrect word. I assumed that you wanted to highlight the problem with distinct keys and size, which was not in the example. So I thought that you made a mistake in the example. But now it seems to me that you meant the free variables.

wytseoortwijn:
Yes, that makes sense. I'm not sure why exactly I saw a problem in that... I think I just worried a bit that such an implementation might break certain other Gobra features.

I just implemented it, and it indeed seems to break the sequence multiplicity operator and I implemented a while ago, as well as some sequence-to-(multi)set conversions. For example, the following doesn't verify any longer:

```go
func foo () {
  assert 12 # seq[int] { 12 } == 1
}
```
nor does
```go
func bar () {
  assert mset(seq[int] { 12 }) == mset[int] { 12 }
}
```

The problem is that the axiomatisations of these operations are defined over singleton sequences and the sequence append operator. For example

```viper
axiom seqmultiplicity_singleton {
  forall x: T, y: T :: { seqmultiplicity(x, Seq(y)) }
    seqmultiplicity(x, Seq(y)) == (x == y ? 1 : 0)
}

axiom seqmultiplicity_app {
  forall x: T, xs: Seq[T], ys: Seq[T] :: { seqmultiplicity(x, xs ++ ys) }
    seqmultiplicity(x, xs ++ ys) == seqmultiplicity(x, xs) + seqmultiplicity(x, ys)
}
```

And likewise for the axiomatisation for sequence conversions. Not quite sure yet how to adapt those. (But hopefully I am just overlooking something obvious also here).

wytseoortwijn:
I tried to find an alternative formulation for the sequence multiplicity axiomatisation as discussed in the Gobra meeting, but so far did not succeed.

Just to try a few things I for example considered a sort-of "unfolding-style" version containing something along the lines of

- `!(x in xs) ==> multiplicity(x, xs) == 0`
- `x in xs ==> exists i : Int :: 0 <= i && i < |xs| && xs[i] == x && multiplicity(x, xs) == 1 + multiplicity(x, xs[..i]) + multiplicity(x, xs[i + 1..])`

But of course doing such a thing leads to matching loops.

I also considered defining `multiplicity(x, xs)` in terms of an extra operation that gives the set of all indices in `xs` at which the value `x` is present; something like

```viper
domain SeqMultiplicity[T]
{
  function indices(x : T, xs : Seq[T]) : Set[Int]
  function multiplicity(x : T, xs : Seq[T]) : Int

  axiom {
    forall x : T, xs : Seq[T] :: { indices(x, xs) }
      forall i : Int :: { i in indices(x, xs) } { xs[i] }
        i in indices(x, xs) <==> (0 <= i && i < |xs| && xs[i] == x)
  }

  axiom {
    forall x : T, xs : Seq[T] :: { multiplicity(x, xs) }
      multiplicity(x, xs) == |indices(x, xs)|
  }
}
```

But such a formulation would have the same problem, only now shifted from `multiplicity` to `indices`.

Would you have ideas perhaps of other adaptions that could be tried? I'm looking for something that allows verifying for example

```viper
method foo(xs : Seq[Int])
  requires |xs| == 5
  requires xs[0] == 0
  requires xs[1] == 4
  requires xs[2] == 2
  requires xs[3] == 4
  requires xs[4] == 0
{
  assert multiplicity(0, xs) == 2
}

method bar(xs : Seq[Int])
  requires |xs| == 100000
  requires forall i : Int :: 0 <= i && i < |xs| ==> xs[i] == 12
{
  assert multiplicity(12, xs) == 100000
}
```

Felalolf:
Since we will probably not find a full solution now, you can just pick what you think works best.

wytseoortwijn:
Then I'll go with the approach of abbreviating large gaps as function calls for now.

PR #144, opened 2020-12-14
Add blank identifier

### LINKED ISSUE #135: No support for blank identifier in assignments
It is a common idiom in Go to assign an expression to a blank identifier when its result is ignored:
```go
  x, _ := 1, i
```

However, doing that in Gobra leads to a parser error
```
/Users/joao/Desktop/Example.gobra:74:6: string matching regex '(?:_[a-zA-Z0-9_]+|[a-zA-Z][a-zA-Z0-9_]*)' expected but '_' found
  x, _ := 1, i;
     ^
```

jcp19:
Adds support for blank identifiers and fixes #135. 

All tests seem to be passing except for the two that are cancelled during testing.
```
[info] - regressions/features/interfaces/boxAndUnbox1.gobra [Gobra] !!! CANCELED !!!
[info]   1 ignored errors
[info]   type_assertion_error:failed_type_assertion, issue 134 (AnnotationBasedTestSuite.scala:81)
[info]   + Time required: 1290 ms 
[info] - regressions/features/interfaces/safeTypeAssertion1.gobra [Gobra] !!! CANCELED !!!
[info]   1 ignored errors
[info]   assert_error:assertion_error, issue 134 (AnnotationBasedTestSuite.scala:81)
[info]   + Time required: 1588 ms 
```
This already happened before my PR.


src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -651,33 +664,37 @@ object Desugar {
...
           case PVarDecl(typOpt, right, left, addressable) =>
 
-            if (left.size == right.size) {
-              sequence((left zip right).map{ case (l, r) =>
+            if (left.size == right.size && left.size == addressable.size) {
+              sequence((left zip right zip addressable).map{ case ((l, r), addr) =>
                 for{
-                  le <- unit(in.Assignee.Var(assignableVarD(ctx)(l)))
                   re <- goE(r)
+                  addressability = if(addr) Addressability.Shared else Addressability.Exclusive
```
jcp19, line 685: I can probably reverse these changes and just used Addressability.Exclusive because Addressability does not seem to be used anywhere, it is in fact ignored when getVar creates a new variable

src/main/scala/viper/gobra/frontend/info/implementation/resolution/NameResolution.scala
```diff
@@ -151,8 +154,8 @@ trait NameResolution { this: TypeInfoImpl =>
       case n: PPackage => n.declarations flatMap { m =>
 
         def actualMember(a: PActualMember): Vector[PIdnDef] = a match {
-          case d: PConstDecl => d.left
-          case d: PVarDecl => d.left
+          case d: PConstDecl => d.left.filter(_.isInstanceOf[PIdnDef]).map(_.asInstanceOf[PIdnDef])
```
Felalolf, line 157: You can do this with a collect `d.left.collect{ case x: PIdnDef => x }`

src/main/scala/viper/gobra/frontend/info/implementation/typing/IdTyping.scala
```diff
@@ -174,6 +175,39 @@ trait IdTyping extends BaseTyping { this: TypeInfoImpl =>
...
 
+    case Wildcard(decl, _) => getWildcardType(decl)
+
     case _ => violation("untypable")
   }
+
+  /**
+    * Gets the type of a blank identifier if it occurs in the `left` list
+    */
+  def getBlankAssigneeType(n: PNode, left: Vector[PNode], right: Vector[PExpression]): Type = {
```
Felalolf, line 186: How is this different from `getBlankIdType` from ExprTyping.scala?

src/main/scala/viper/gobra/frontend/info/implementation/typing/IdTyping.scala
```diff
@@ -174,6 +175,39 @@ trait IdTyping extends BaseTyping { this: TypeInfoImpl =>
...
+    require(n.isInstanceOf[PIdnNode] || n.isInstanceOf[PBlankIdentifier])
+    // joao: the use of "eq" here is a hack to find the index of the wildcard in the left list
+    // and then use that index to retrieve the corresponding expression
+    val pos = left indexWhere (n eq _)
```
Felalolf, line 190: We do this quite often.

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
```diff
@@ -55,7 +55,8 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
         assignableTo.errors(exprType(right), exprType(left))(n)
 
     case n@PShortVarDecl(rights, lefts, _) =>
-      if (lefts.forall(pointsToData))
+      // TODO: check that at least one of the variables is new
+      if (lefts.filter(_.isInstanceOf[PIdnUnk]).forall(pointsToData))
```
Felalolf, line 59: Could be done in one forall

Felalolf (commented):
I looked through it, but Linard should also have a look at it. For me it seemed fine, but I went over it quickly.

src/main/scala/viper/gobra/frontend/info/implementation/property/Addressability.scala
```diff
@@ -110,6 +111,7 @@ trait Addressability extends BaseProperty { this: TypeInfoImpl =>
     attr[PIdnNode, AddrMod] { n => regular(n) match {
       case v: Variable => if (v.addressable) AddrMod.sharedVariable else AddrMod.exclusiveVariable
       case _: Constant => AddrMod.constant
+      case _: Wildcard => AddrMod.rValue
```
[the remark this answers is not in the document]
jcp19: yes, I think that `defaultValue` and `exclusiveVariable` are indeed more suitable here

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1065,6 +1083,10 @@ object Desugar {
 
           case g: PGhostExpression => ghostExprD(ctx)(g)
 
+          case b: PBlankIdentifier =>
+            val typ = typeD(info.typ(b), Addressability.exclusiveVariable)(src)
```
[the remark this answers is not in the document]
jcp19: Sorry, I forgot to change that.  In fact, what is stored in an `ap.BlankIdentifier()` is the corresponding PBlankIdentifier in the AST. The type of a blank identifier is found using the `getBlankIdType` method. Essentialy, what it does is to find the corresponding expression of a given `_` that appears on the LHS of an assignment and returns its type

src/main/scala/viper/gobra/frontend/info/implementation/typing/IdTyping.scala
jcp19: `getBlankIdType` accepts a `PBlankIdentifier b` and does a case split on the expressions where `b` could occur. It then delegates all the logic of finding the value of the blank identifier to the `getBlankAssigneeType` method. `getWildcardType` in file `IdTyping.scala` has an analogous behavior for `PWildcard`s

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -611,18 +612,18 @@ case class PLabelDef(name: String) extends PDefLikeLabel
...
-trait PUseLikePkg extends PPackegeNode
+trait PDefLikePkg extends PPackageNode
+trait PUseLikePkg extends PPackageNode
 
 case class PPkgDef(name: PPkg) extends PDefLikePkg
 case class PPkgUse(name: PPkg) extends PUseLikePkg
 
 
-case class PWildcard() extends PDefLikeId with PUseLikeId {
+case class PWildcard() extends PDefLikeId with PUseLikeId with PUnkLikeId {
```
ArquintL, line 626: What is the difference between `PBlankIdentifier` and `PWildcard`?

src/main/scala/viper/gobra/frontend/info/implementation/resolution/AmbiguityResolution.scala
```diff
@@ -98,6 +98,8 @@ trait AmbiguityResolution { this: TypeInfoImpl =>
       case Right(_) => None // unknown pattern
     }
 
+    case b: PBlankIdentifier => Some(ap.BlankIdentifier(b))
```
ArquintL, line 101: why do we need ambiguity resolution? Isn't it straight forward what it is?

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
```diff
@@ -55,7 +55,8 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
         assignableTo.errors(exprType(right), exprType(left))(n)
 
     case n@PShortVarDecl(rights, lefts, _) =>
-      if (lefts.forall(pointsToData))
+      // TODO: check that at least one of the variables is new
+      if (lefts.forall{ x => !x.isInstanceOf[PIdnUnk] || pointsToData(x) })
```
ArquintL, line 59: Why is the inequality to `PIdnUnk` added? And why does at least one variable need to be new?

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GhostAssignability.scala
```diff
@@ -58,6 +58,10 @@ trait GhostAssignability {
 
       case _ => error(left, "ghost error: selections on types are not assignable")
     }
+
+    case PBlankIdentifier() =>
+      // message(left, "ghost error: ghost cannot be assigned to blank identifier", isRightGhost)
```
ArquintL, line 63: remove comment

ArquintL (approved):
Looks good to be. I just haven't understood yet the difference between wildcard and blank identifier yet. Isn't it the same?

src/main/scala/viper/gobra/ast/frontend/Ast.scala
jcp19: `PBlankIdentifier` and `PWildcard` in essence represent the same concept but they fall into different syntactic categories. `PWildcard` is an identifier (i.e. a subtype of `PIdnNode`) whereas `PBlankIdentifier` is a subtype of `PAssignee` and thus, it is also a `PExpression`. The reason for this seemingly artificial divide between the two is that declaration statements (i.e. `PVarDecl`, `PShortVarDecl` and `PConstDecl`) expect the LHS of the operation to be a vector of ids:
```scala
case class PShortVarDecl(right: Vector[PExpression], left: Vector[PUnkLikeId], addressable: Vector[Boolean]) extends PSimpleStmt with PGhostifiableStatement
```
while assignment expressions expect a vector of expressions (more precisely, a vector of `PAssignee`s):
```scala
case class PAssignment(right: Vector[PExpression], left: Vector[PAssignee]) extends PSimpleStmt with PGhostifiableStatement
```
**Obs**: `PWildcard` is a subtype of `PUnkLideId` and the second is a subtype of `PIdnNode`

src/main/scala/viper/gobra/frontend/info/implementation/resolution/AmbiguityResolution.scala
jcp19: I don't think so. This case allows us to retrieve the exact location where a `PBlankIdentifier` occurs. The location is important in order to find the type `T` of the expression assigned to the blank identifier. Furthermore, T must be known during desugaring because blank identifiers are desugard into fresh variables of type `T`

in order to create a fresh variable in the place of the identifier of the same type. In order to compute the

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
jcp19: 1. my goal here was to have `x.isInstanceOf[PIdnUnk] implies pointsToData(x)`, which is equivalent to the expression in parentheses. Nevertheless, now that I think of it, it would make sense to make Wildcard a subtype of `DataEntity` and thus, I will undo the changes in this line
2. That is mentioned in the [specification for short variable declarations](https://golang.org/ref/spec#Short_variable_declarations), specifically 
> a short variable declaration may redeclare variables provided they were originally declared earlier in the same block (or the parameter lists if the block is the function body) with the same type, and at least one of the non-blank variables is new.

src/main/scala/viper/gobra/ast/frontend/Ast.scala
ArquintL: Ah makes sense! Thanks

PR #146, opened 2020-12-14
Removes ViperServer Dependency and adds explicit Executor

ArquintL:
This PR changes two things:
1. ViperServer is removed as a dependency (Gobra clients such as Gobra-IDE will have to add ViperServer as a direct dependency in case they need it)
2. An explicit executor called `GobraExecutionContext` can be passed around that is used for e.g. futures. `DefaultGobraExecutionContext` is an instance of `GobraExecutionContext` that is used for the CLI


src/main/scala/viper/gobra/util/GobraExecutionContext.scala
```diff
@@ -0,0 +1,11 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2011-2020 ETH Zurich.
+
+package viper.gobra.util
+
+import scala.concurrent.ExecutionContext
+
+trait GobraExecutionContext extends ExecutionContext
```
jcp19, line 11: Why not put `GobraExecutionContext` and `DefaultGobraExecutionContext` in the same file?

jcp19 (approved):

src/main/scala/viper/gobra/util/GobraExecutionContext.scala
ArquintL: Great idea! I think I'm still thinking the Java way, i.e. having one file per "thing"

ArquintL:
Tests successfully run locally. I'm merging despite failing CI since SBT update and therefore compatibility with updated silver, silicon, and carbon is coming with PR #147

Issue #140, opened 2020-12-10
Support for Mutex


ArquintL:
- [x] Add stub mechanism to ship stubs for common go libraries with Gobra
- [ ] Create stub for mutex

Depends on #139

PR #148, opened 2020-12-16
Support for Stubs

ArquintL:
This PR adds `src/main/resources/stubs` to store and use stubs. A simple assertion library taken from (an old version of) SCION is used for demonstration purposes and test cases.
`src/main/resources/README.md` provides a short explanation of the feature and how it influences Gobra's operation

First part of #140

### DIFF
--- a/src/main/scala/viper/gobra/frontend/PackageResolver.scala
+++ b/src/main/scala/viper/gobra/frontend/PackageResolver.scala
@@ -77,6 +77,11 @@ object PackageResolver {
       val nullableResourceUri = getClass.getClassLoader.getResource(stubDir).toURI
       for {
         resourceUri <- Option.when(nullableResourceUri != null)(nullableResourceUri)
+        // when executing the tests, the resource URI points to regular files in the filesystem (i.e. the
+        // URI scheme is "file"). However, when directly running Gobra, files in the resource folder are contained in
+        // the jar file. As this is some kind of zip file, accessing the stubs as regular files is not possible.
+        // BaseJarResource provides an abstraction that internally uses an adhoc filesystem to enable file-like access
+        // to these resources.
         resource <- resourceUri.getScheme match {
           case s if s == fileUriScheme => Some(FileResource(Paths.get(resourceUri).toFile))
           case s if s == jarUriScheme => Some(BaseJarResource(resourceUri, stubDir))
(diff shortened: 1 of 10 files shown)


src/main/scala/viper/gobra/frontend/PackageResolver.scala
```diff
@@ -50,68 +58,102 @@ object PackageResolver {
...
+
+  private lazy val getStubResources: Vector[InputResource] = {
+    // get path to stubs
+    stubDirectories.flatMap(stubDir => {
+      val nullableResourceUri = getClass.getClassLoader.getResource(stubDir).toURI
+      for {
+        resourceUri <- Option.when(nullableResourceUri != null)(nullableResourceUri)
+        resource <- resourceUri.getScheme match {
+          case s if s == fileUriScheme => Some(FileResource(Paths.get(resourceUri).toFile))
+          case s if s == jarUriScheme => Some(BaseJarResource(resourceUri, stubDir))
```
jcp19, line 82: I did not get why there are two possible Uri schemes when you get a stub resource

src/main/scala/viper/gobra/frontend/PackageResolver.scala
```diff
@@ -124,13 +166,140 @@ object PackageResolver {
...
+        stream = None
+      }
+      case _ =>
+    }
+  }
+
+  case class FileResource(file: File) extends InputResource {
+    override lazy val path: Path = file.toPath
+
+    def resolve(pathComponent: String): FileResource =
```
jcp19, line 217: isn't the `override` keyword missing for this declaration?

jcp19 (approved):

src/main/scala/viper/gobra/frontend/PackageResolver.scala
ArquintL: When you run Gobra's tests then files in the stub folder are regular files in your filesystem. However, when you invoke Gobra with `sbt run -i ...`, sbt seems to pack everything into a jar first and execute the jar. The stub files also get packed into the jar. As the jar is some kind of zip file, accessing the stubs as regular files (e.g. calling `toFile` on a `Path`) fails. I have found two ways how to access such "packed files" in a jar: (1) you specify the path to such a file and read it using streams or (2) you create a new "filesystem" for the files in the jar. I went with option 2 as this allows scanning the directory structure as I do not make any assumption about which files are present or not.

src/main/scala/viper/gobra/frontend/PackageResolver.scala
ArquintL: It is. IntelliJ does not even issue a warning for this

src/main/scala/viper/gobra/frontend/PackageResolver.scala
jcp19: That makes sense. Thanks!

jcp19 (approved):

PR #153, opened 2020-12-18
Support for Fractional Permissions

### LINKED ISSUE #150: Add support for fractional permissions
Currently, there is no support in Gobra for specifying partial permissions to references. In fact,  the only code I found regarding parsing access assertions in the Parser is the following:
```scala
    lazy val access : Parser[PAccess] =
      "acc" ~> "(" ~> expression <~ ")" ^^ PAccess
```
which indicates that Gobra only supports reasoning about full permissions

ArquintL:
This PR adds support for fractional field and predicate permissions.
`none`, `write`, `_`, and `x/y` where x and y are two expressions of integer type are supported as permission amounts. These permission amounts are directly translated to corresponding silver constructs.
Note that `write` is added as a new keyword (`none` is already a keyword used for the option type).
Although there is now a `PermissionT` type internally, it's currently not exposed to Gobra users (i.e. similar to the `perm` type in Viper)
Closes #150


src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
```diff
@@ -144,6 +144,18 @@ trait GhostExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
       }
     }
+
+    case expr: PPermission => expr match {
+      case PFullPerm() => noMessages
+      case PNoPerm() => noMessages
+      case PFractionalPerm(left, right) => (intConstantEval(left), intConstantEval(right)) match {
+        case (l, r) if l.isEmpty || r.isEmpty => error(left, s"expected a constant, but got $left", l.isEmpty) ++
+          error(right, s"expected a constant, but got $right", r.isEmpty)
+        case (_, Some(dividend)) => error(right, s"expected a non-zero dividend, but got $right", dividend == 0)
```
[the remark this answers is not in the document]
ArquintL: I actually tried that but Scala is for some reason not able to implicitly convert `0` to a `BigInt`. Because `BigInt` also does not seem to have an unapply function, the best imho I can do is this: `case (_, Some(zero)) if zero == 0 => ...`

src/main/scala/viper/gobra/ast/internal/utility/GobraStrategy.scala
```diff
@@ -55,7 +55,7 @@ object GobraStrategy {
       case (_: SepAnd, Seq(l: Assertion, r: Assertion)) => SepAnd(l, r)(meta)
       case (_: ExprAssertion, Seq(exp: Expr)) => ExprAssertion(exp)(meta)
       case (_: Implication, Seq(l: Expr, r: Assertion)) => Implication(l, r)(meta)
-      case (_: Access, Seq(acc: Accessible)) => Access(acc)(meta)
+      case (_: Access, Seq(acc: Accessible, perm: Permission)) => Access(acc, perm)(meta)
```
Felalolf, line 58: The cases for the other new nodes are missing.

src/main/scala/viper/gobra/ast/internal/utility/Nodes.scala
```diff
@@ -57,7 +57,7 @@ object Nodes {
         case SepForall(vars, triggers, body) => vars ++ triggers ++ Seq(body)
         case ExprAssertion(exp) => Seq(exp)
         case Implication(left, right) => Seq(left, right)
```
Felalolf, line 59: the cases for the other new nodes are missing.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
```diff
@@ -335,6 +349,8 @@ trait GhostExprTyping extends BaseTyping { this: TypeInfoImpl =>
 
       // Others
       case PReceive(_) => false
+
+      case _: PPermission => true
```
Felalolf, line 353: The children of a fractional permission could be impure (not now, but maybe in the future).

src/main/scala/viper/gobra/translator/implementations/translator/PermissionsImpl.scala
```diff
@@ -0,0 +1,34 @@
...
+package viper.gobra.translator.implementations.translator
+
+import viper.gobra.ast.internal.{FractionalPerm, FullPerm, NoPerm, WildcardPerm}
+import viper.gobra.ast.{internal => in}
+import viper.gobra.translator.interfaces.{Collector, Context}
+import viper.gobra.translator.interfaces.translator.Permissions
+import viper.gobra.translator.util.ViperWriter.CodeWriter
+import viper.silver.{ast => vpr}
+
+class PermissionsImpl extends Permissions {
```
Felalolf, line 16: Is there a reason why you added this as an translator? I would have just add this a typeEncoding because later we will have to change it to that either way (when we want to support perm as a regular type). I nothing is speaking against this, I would ask you to add the permission enocoding as an type encoding, e.g. look at bool encoding.

Felalolf (changes requested):
Just send me a whatsapp message when I should look at it again.

src/main/scala/viper/gobra/translator/implementations/translator/PermissionsImpl.scala
ArquintL: It seemed more natural to me and I think I got inspired by expressions / assertions. But I don't mind changing it to a type encoding

Felalolf (approved):
Seems good. With this, we do not yet have full support for perm in the frontend, e.g. perm parameter and variables are missing. This should be added at some point.

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -1125,7 +1125,20 @@ object Parser {
...
     lazy val access : Parser[PAccess] =
-      "acc" ~> "(" ~> expression <~ ")" ^^ PAccess
+      "acc" ~> "(" ~> expression <~ ")" ^^ { exp => PAccess(exp, PFullPerm().at(exp)) } |
+      "acc" ~> "(" ~> expression ~ ("," ~> permission <~ ")") ^^ PAccess
+
+    lazy val permission: Parser[PPermission] =
+      fractionalPermission |
+      "write" ^^^ PFullPerm() |
+      "none" ^^^ PNoPerm() |
+      "_" ^^^ PWildcardPerm()
```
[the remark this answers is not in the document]
ArquintL: Isn't it always clear from the context that the permission wildcard is meant? And should I bring this topic up in one of our future meeting and/or should I change the syntax in this PR?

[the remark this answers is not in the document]
Felalolf: Maybe just wait until todays meeting if that is fine. Right now, it would still be easy to change.

PR #154, opened 2020-12-22
Add go routines

### LINKED ISSUE #138: Spawning Go Routines
- [x] Add support for "go" keyword
- [ ] Restrict postconditions of called functions to `true`

jcp19:
Fixes #138: Spawning go routines


src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -728,6 +728,38 @@ object Desugar {
               }
             } yield in.Seqn(Vector(dPre, exprAss, clauseBody))(src)
 
+          case PGoStmt(exp) =>
+            def unexpectedExprError(exp: PExpression) = violation(s"unexpected expression $exp in go statement")
+
+            def getArgs(args: Vector[PExpression]): Writer[Vector[in.Expr]] = {
+              sequence(args map exprD(ctx)).map {
+                // go function chaining feature
+                case Vector(in.Tuple(targs)) if args.size > 1 => targs
```
jcp19, line 737: This line has a small error, there is no point in comparing `args.size` to 1. Instead, I wanted to compare the number of parameters in the function definition to 1

src/main/scala/viper/gobra/ast/internal/PrettyPrinter.scala
```diff
@@ -201,6 +201,11 @@ class DefaultPrettyPrinter extends PrettyPrinter with kiama.output.PrettyPrinter
       (if (targets.nonEmpty) showVarList(targets) <+> "=" <> space else emptyDoc) <>
         showExpr(recv) <> meth.name <> parens(showExprList(args))
 
+    case GoFunctionCall(func, args) => "go" <> space <> func.name.name <> parens(showExprList(args))
```
ArquintL, line 204: `<> space <>` can be simplified to `<+>`

src/main/scala/viper/gobra/reporting/VerifierError.scala
```diff
@@ -231,6 +231,11 @@ case class SynthesizedAssertionFalseError(info: Source.Verifier.Info) extends Ve
   override def message: String = info.comment.reduce[String] { case (l, r) => s"$l; $r" }
 }
 
+case class GoCallPreconditionError(node: Source.Verifier.Info) extends VerificationErrorReason {
+  override def id: String = "go_call_precondition_error"
+  override def message: String = s"${node.origin.tag.trim} might not satisfy method precondition."
```
ArquintL, line 236: I would change the message to something like `... might not satisfy precondition of called function or method` since this error (in my understanding) can occur for both go method and go function calls, can't it?

src/main/scala/viper/gobra/translator/implementations/translator/StatementsImpl.scala
```diff
@@ -37,6 +39,23 @@ class StatementsImpl extends Statements {
...
+    def genExhale(pre: Vector[in.Assertion], formalParams: Vector[in.Parameter.In], args: Vector[in.Expr]): Writer[vpr.Stmt] = {
+      Violation.violation(args.length == formalParams.length, "number of passed arguments must match number of expected arguments")
+      val errorT = (x: Source.Verifier.Info, _: ErrorReason) => PreconditionError(x) dueTo GoCallPreconditionError(x)
+      for {
+        vArgss <- sequence(args map goE)
+        funcArgs <- sequence(formalParams map goE)
+        substitutions = (funcArgs zip vArgss).toMap
+        preCond <- sequence(pre map goA)
+        preCondInstance = preCond.map{ _.replace(substitutions) }
+        and = preCondInstance.foldRight[vpr.Exp](vpr.TrueLit()(pos, info, errT))((x,y) => vpr.And(x,y)(pos, info, errT))
```
ArquintL, line 54: Please use `ViperUtil.bigAnd` since foldRight / foldLeft has significant performance impact on Silicon. Having a central place to change that (if ever necessary) seems advantageous to me

src/main/scala/viper/gobra/translator/implementations/translator/StatementsImpl.scala
```diff
@@ -37,6 +39,23 @@ class StatementsImpl extends Statements {
     def goA(a: in.Assertion): CodeWriter[vpr.Exp] = ctx.ass.translate(a)(ctx)
     def goE(e: in.Expr): CodeWriter[vpr.Exp] = ctx.expr.translate(e)(ctx)
 
+    /**
+      * Generates an Exhale statement for the assertions `pre` parametrized by `formalParams` and instantiated with `args`
+      */
+    def genExhale(pre: Vector[in.Assertion], formalParams: Vector[in.Parameter.In], args: Vector[in.Expr]): Writer[vpr.Stmt] = {
```
ArquintL, line 45: I would change the function name and documentation to something like `translateGoCall` because:
- the error transformer is specific to go calls
- this function is only used to handle go calls
- I would put emphasis on handling go calls and consider the generation of exhales an implementation detail

ArquintL (approved):
Looks great! I only have a few minor remarks

src/main/scala/viper/gobra/reporting/VerifierError.scala
jcp19: yes, you are right!

src/main/scala/viper/gobra/translator/implementations/translator/StatementsImpl.scala
jcp19: I did not know about this function 😅 Thanks @ArquintL!

src/main/scala/viper/gobra/translator/implementations/translator/StatementsImpl.scala
jcp19: Yes, looking back, I totally agree with your comment

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -172,6 +187,10 @@ case class SafeTypeAssertion(resTarget: LocalVar, successTarget: LocalVar, expr:
 
 case class FunctionCall(targets: Vector[LocalVar], func: FunctionProxy, args: Vector[Expr])(val info: Source.Parser.Info) extends Stmt
 case class MethodCall(targets: Vector[LocalVar], recv: Expr, meth: MethodProxy, args: Vector[Expr])(val info: Source.Parser.Info) extends Stmt
+// Go function and method calls must receive a Member (more precisely, a fully desugared function or method)
+// in order to retrieve its pre-condition
+case class GoFunctionCall(func: FunctionMember, args: Vector[Expr])(val info: Source.Parser.Info) extends Stmt
+case class GoMethodCall(recv: Expr, meth: MethodMember, args: Vector[Expr])(val info: Source.Parser.Info) extends Stmt
```
Felalolf, line 193: This might be bad, but I am not entirely certain how case classes behave. 
I would be worried about the case where a function calls itself with a go statement. 
Do when then have an infinite structure? At least, we would not have a tree anymore.

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -728,6 +728,40 @@ object Desugar {
...
+            }
+
+            exp match {
+              case inv: PInvoke => info.resolve(inv) match {
+                case Some(p: ap.FunctionCall) => p.callee match {
+                  case ap.Function(_, st.Function(decl, _, _)) =>
+                    val func = functionD(decl)
+                    for { args <- getArgs(decl.args.length, p.args) } yield in.GoFunctionCall(func, args)(src)
+
+                  case ap.ReceivedMethod(recv, _, _, st.MethodImpl(decl, _, _)) =>
```
Felalolf, line 751: it seems to me that the case for a method expression is missing.

src/main/scala/viper/gobra/reporting/VerifierError.scala
```diff
@@ -231,6 +231,11 @@ case class SynthesizedAssertionFalseError(info: Source.Verifier.Info) extends Ve
   override def message: String = info.comment.reduce[String] { case (l, r) => s"$l; $r" }
 }
 
+case class GoCallPreconditionError(node: Source.Verifier.Info) extends VerificationErrorReason {
+  override def id: String = "go_call_precondition_error"
+  override def message: String = s"${node.origin.tag.trim} might not satisfy precondition of called function or method."
```
Felalolf, line 236: "precondition of callee"

src/main/scala/viper/gobra/translator/implementations/translator/StatementsImpl.scala
```diff
@@ -37,6 +39,26 @@ class StatementsImpl extends Statements {
...
+      val errorT = (x: Source.Verifier.Info, _: ErrorReason) => PreconditionError(x) dueTo GoCallPreconditionError(x)
+      for {
+        vArgss <- sequence(args map goE)
+        funcArgs <- sequence(formalParams map goE)
+        substitutions = (funcArgs zip vArgss).toMap
+        preCond <- sequence(pre map goA)
+        preCondInstance = preCond.map{ _.replace(substitutions) }
+        and = vu.bigAnd(preCondInstance)(pos, info, errT)
+        _ <- assert(and, errorT)
+      } yield vpr.Exhale(and)(pos, info, errT)
```
Felalolf, line 59: I think this needs to be surrounded by seqn( ... )

Felalolf, line 59: Here, you are first asserting and then exhaling the conjunct. The assertion is superfluous. You should add the error trasnformer directly to the exhale statement.

Felalolf (changes requested):

Felalolf:
Please, send me a message on whatsapp when the things are changed, so that I do not miss it.

src/main/scala/viper/gobra/ast/internal/Program.scala
jcp19: yes, this causes an infinite loop indeed! I added a test for this case and I solved the issue by adding a flag to the function and method desugarer to ignore the body when desugaring.

src/main/scala/viper/gobra/translator/implementations/translator/StatementsImpl.scala
jcp19: Most likely I am missing something but why should I use seqn here if I only want to write a single Exhale statement?

src/main/scala/viper/gobra/frontend/Desugar.scala
jcp19: You are right

jcp19: (fixed in commit 89fcad2993e1125c984c9638da397a30afea850e)

src/main/scala/viper/gobra/translator/implementations/translator/StatementsImpl.scala
Felalolf: Without seqn you can get some reordering issues. If you have 
```
for {
  w <- for { _ <- write(B) } yield C
} yield vpr.Seqn(Vector(A, w))
```
Then you would generate B;A;C instead of A;B;C. If A is influenced by state that is changed by B, then this is wrong.
Essentially, 
```
seqn(
  for {
   _ <- write(S_1)
   _ <- write(S_2)
   ...
   - <- write(S_N)
  } yield Z
)
```
is the same as ``unit(vpr.Seqn(S_1, S_2, ... , S_N, Z))`. In an earlier version, I had more complicated writer that enforced this. But dealing with them was a pain, so I reduced it to the two writer that we have now. However, as a rule of thump, for the encoding of statements, if you have a 'write' or 'assert' (anything that generates statements), then you should put a seqn arround.

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -30,13 +30,38 @@ case class Program(
...
+                 definedMethods: Map[MethodProxy, MethodMember],
+                 definedFunctions: Map[FunctionProxy, FunctionMember],
+                 definedMPredicates: Map[MPredicateProxy, MPredicate],
+                 definedFPredicates: Map[FPredicateProxy, FPredicate]
+                 ) {
   def lookup(t: DefinedT): Type = definedTypes(t.name, t.addressability)
+  def lookup(m: MethodProxy): MethodMember = definedMethods(m)
+  def lookup(f: FunctionProxy): FunctionMember = definedFunctions(f)
+  def lookup(m: MPredicateProxy): MPredicate = definedMPredicates(m)
+  def lookup(f: FPredicateProxy): FPredicate = definedFPredicates(f)
```
Felalolf, line 43: great

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -41,7 +41,13 @@ object Desugar {
     val importedPrograms = imported.map(_._2)
     val types = mainProgram.types ++ importedPrograms.flatMap(_.types)
     val members = mainProgram.members ++ importedPrograms.flatMap(_.members)
-    val table = new in.LookupTable(mainDesugarer.definedTypes ++ importedDesugarers.flatMap(_.definedTypes))
+    def combineTableField[X,Y](f: Desugarer => Map[X,Y]): Map[X,Y] =
+      f(mainDesugarer) ++ importedDesugarers.flatMap(f)
+    val table = new in.LookupTable(combineTableField(_.definedTypes),
```
Felalolf, line 46: `combineTableField(_.definedTypes),` should be on the next line (code style).

Felalolf (approved):
Seems good. I only have one tiny style remark.

src/main/scala/viper/gobra/translator/implementations/translator/StatementsImpl.scala
jcp19: Thank you for a thorough explanation!

PR #160, opened 2021-01-08
Add support for `make` and `new` functions

jcp19:
This PR adds support for the [`make`](https://golang.org/pkg/builtin/#make) and [`new`](https://golang.org/pkg/builtin/#make) functions available in *Go*. It still needs debugging because *Gobra* crashes when inhaling the footprint of each cell of the slice (see https://github.com/viperproject/gobra/pull/160#discussion_r553672974).

@ArquintL once this PR is reviewed and I incorporate the feedback, you can build upon it to encode `make` for channels


src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -144,6 +145,85 @@ class SliceEncoding(arrayEmb : SharedArrayEmbedding) extends LeafTypeEncoding {
...
+            eqCap <- ctx.typeEncoding.equal(ctx)(capExpr, capArg, makeStmt)
+            _ <- write(vpr.Inhale(eqCap)(pos, info, errT))
+
+            // inhales len(a) == [len]
+            eqLen <- ctx.typeEncoding.equal(ctx)(lenExpr, lenArg, makeStmt)
+            _ <- write(vpr.Inhale(eqLen)(pos, info, errT))
+
+            // inhales forall i: int :: {loc(a, i)} 0 <= i && i < [cap] ==> Footprint[ a[i] ]
+            vprIt = ctx.typeEncoding.variable(ctx)(it)
+            footprint <- addressFootprint(ctx)(in.IndexedExp(slice, it)(makeStmt.info), in.FullPerm(slice.info))
```
jcp19, line 194: Gobra runs into an Exception here. For some reason, the condition `typ(ctx).isDefinedAt(t)` in the body of the `addressFootprint` method is not true, which causes Gobra to crash. @Felalolf, do you have any idea of why this happens?

jcp19:
The PR should now be ready for review. The most interesting part is the encoding of slices, specially the methods `statement` and `addressFootprint`. The later inhales permissions to the fields of a slice and to all of its cells.

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -144,6 +146,134 @@ class SliceEncoding(arrayEmb : SharedArrayEmbedding) extends LeafTypeEncoding {
...
+    *    inhales cap(a) == [cap]
+    *    inhales len(a) == [len]
+    *    inhales forall i: int :: {loc(a, i)} 0 <= i && i < [cap] ==> Footprint[ a[i] ]
+    *    inhales forall i: int :: {loc(a, i)} 0 <= i && i < [len] ==> [ a[i] == dfltVal(T) ]
+    */
+  override def statement(ctx: Context): in.Stmt ==> CodeWriter[vpr.Stmt] = {
+    default(super.statement(ctx)) {
+      case makeStmt@in.MakeSlice(target, typeParam, lenArg, optCapArg) =>
+        val (pos, info, errT) = makeStmt.vprMeta
+        val slice = in.LocalVar(Names.freshName, in.SliceT(typeParam.withAddressability(Shared), Addressability.Shared))(makeStmt.info)
```
Felalolf, line 163: Should exclusive.

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -144,6 +146,134 @@ class SliceEncoding(arrayEmb : SharedArrayEmbedding) extends LeafTypeEncoding {
...
+    * We do not use let because (at the moment) Viper does not accept quantified permissions with let expressions.
+    */
+  override def addressFootprint(ctx: Context): (in.Location, in.Permission) ==> CodeWriter[vpr.Exp] = {
+    case (loc :: ctx.Slice(t) / Shared, perm) =>
+      val (pos, info, errT) = loc.vprMeta
+
+      val fieldAccWriter = for {
+        vprPerm <- ctx.typeEncoding.expr(ctx)(perm)
+        l <- ctx.expr.translate(loc)(ctx).map(_.asInstanceOf[vpr.FieldAccess])
+      } yield vpr.FieldAccessPredicate(l, vprPerm)(pos, info, errT)
```
Felalolf, line 228: This part should not be required for make. If you make `a` exclusive, then these permissions are not required.

Felalolf (changes requested):

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -167,7 +167,15 @@ object Assignee {
...
 }
 
-case class Make(target: LocalVar, typ: CompositeObject)(val info: Source.Parser.Info) extends Stmt
+sealed trait MakeStmt extends Stmt {
+  val target: LocalVar
+}
+
+case class MakeSlice(override val target: LocalVar, typeParam: Type, lenArg: Expr, capArg: Option[Expr])(val info: Source.Parser.Info) extends MakeStmt
+case class MakeChannel(override val target: LocalVar, typeParam: Type, bufferSizeArg: Option[Expr])(val info: Source.Parser.Info) extends MakeStmt
+case class MakeMap(override val target: LocalVar, keyTypeParam: Type, valueTypeParam: Type, initialSpaceArg: Option[Expr])(val info: Source.Parser.Info) extends MakeStmt
```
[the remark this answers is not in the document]
jcp19: I see your point and I can definetly change that. Regarding the specific case of Channels, that was by design given that you can only make bidirectional channels

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -144,6 +146,134 @@ class SliceEncoding(arrayEmb : SharedArrayEmbedding) extends LeafTypeEncoding {
...
+      )
+
+      for {
+        fieldAcc <- fieldAccWriter
+        cellPerms <- cellPermsWriter
+      } yield vpr.And(fieldAcc, cellPerms)(pos, info, errT)
+  }
+
+  /** Returns: Forall idx :: {'trigger'(idx)} 0 <= idx && idx < 'length' => ['body'(idx)] */
+  private def boundedQuant(length: vpr.Exp, trigger: vpr.LocalVar => Seq[vpr.Trigger], body: in.BoundVar => CodeWriter[vpr.Exp])
```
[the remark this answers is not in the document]
jcp19: Yes, I only performed minor changes such as making the `length` param of type `vpr.Expr` instead of `BigInt`

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -277,6 +277,30 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
     }
 
+    case _: PNew => noMessages
+
+    case m@PMake(typ, args) =>
+      args.flatMap { arg =>
+        val constEval = intConstantEval(arg)
+        val isConstExpr = constEval.isDefined
+        error(m, s"arguments to make must be non-negative", isConstExpr && constEval.get < 0)
+      } ++ (typ match {
```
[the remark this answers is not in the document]
jcp19: Good catch!

ArquintL:
@jcp19 let me know when you've incorporated Felix' suggestions and I will then have a look

src/main/scala/viper/gobra/ast/internal/PrettyPrinter.scala
```diff
@@ -188,7 +188,17 @@ class DefaultPrettyPrinter extends PrettyPrinter with kiama.output.PrettyPrinter
     case While(cond, invs, body) => "while" <> parens(showExpr(cond)) <> line <>
       hcat(invs  map ("invariant " <> showAss(_) <> line)) <> block(showStmt(body))
 
-    case Make(target, typ) => showVar(target) <+> "=" <+> "new" <> brackets(showCompositeObject(typ))
+    case New(target, expr) => showVar(target) <+> "=" <+> "new" <> parens(showExpr(expr))
+
+    case MakeSlice(target, typeParam, lenArg, capArg) => showVar(target) <+> "=" <+> "make" <>
+      parens("[]" <> showType(typeParam) <> comma <+> showExprList(lenArg +: capArg.toVector))
```
Felalolf, line 194: The `[]` before the type is now superfluous.

src/main/scala/viper/gobra/reporting/VerifierError.scala
```diff
@@ -179,6 +179,10 @@ case class OverflowError(info: Source.Verifier.Info) extends VerificationError {
   override def localMessage: String = "Expression may cause integer overflow"
 }
 
+case class MakePreconditionError(info: Source.Verifier.Info) extends VerificationError {
+  override def localId: String = "make_precondition_error"
+  override def localMessage: String = s"The provided length might not be smaller or equal to the provided capacity, or either length or capacity might be negative"
```
Felalolf, line 184: It is not either length or capacity, might could be negative as well.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -277,6 +277,34 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
       }
     }
 
+    case _: PNew => noMessages
+
+    case m@PMake(typ, args) =>
+      args.flatMap { arg =>
+        exprType(arg) match {
+          case IntT(config.typeBounds.Int) | IntT(config.typeBounds.UntypedConst) =>
```
Felalolf, line 285: I would hav expected a check with `assignableTo.errors(exprType(e), ...)(n)`. Also, maybe a stupid question: does the length and capacity argument have to be an int or untyped constant? What about long? or unsigned int, etc?

Felalolf (approved):
Seems good, I only have minor comments.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
jcp19: Yes, it has to be an `int` or, more precisely, according to the go spec
>  Each of the size arguments n and m must be of integer type or an untyped constant. A constant size argument must be non-negative and representable by a value of type int; if it is an untyped constant it is given type int.

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -711,11 +711,22 @@ object Parser {
...
         len |
         cap |
         ghostUnaryExp |
         primaryExp
 
+    lazy val make : Parser[PMake] =
+      "make" ~> ("(" ~> typ ~ expSequence <~ ")") ^^ PMake
+
+    lazy val expSequence: Parser[Vector[PExpression]] =
+      ("," ~> rep1sep(expression, ",") <~ ",".?).? ^^ (opt => opt.getOrElse(Vector.empty))
```
ArquintL, line 725: Is a comma at the end (e.g. `make([]int, 5,)`) just accepted by the Go compiler or is this part of the spec? I couldn't find anything about this in the section on `make` but the compiler indeed seems to be happy

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -275,6 +275,29 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
+          error(m, s"too many arguments to make($typ)", args.length > 2) ++
+            error(m, s"missing len argument to make($typ)", args.isEmpty) ++
+            check(args){
+              case args if args.length == 2 =>
+                val maybeLen = intConstantEval(args(0))
+                val maybeCap = intConstantEval(args(1))
+                error(m, s"len larger than cap in make($typ)", maybeLen.isDefined && maybeCap.isDefined && maybeLen.get > maybeCap.get)
+            }
+
+        case _: PBiChannelType | _: PMapType =>
```
ArquintL, line 295: Isn't it fine to use `PSendChannelType` as well as `PRecvChannelType`?

src/main/scala/viper/gobra/translator/encodings/TypeEncoding.scala
```diff
@@ -192,23 +192,23 @@ trait TypeEncoding extends Generator {
 
   /**
     * Encodes statements.
-    * This includes make-statements.
+    * This includes new-statements and make-statements.
```
ArquintL, line 195: I only see an implementation for new-stmts

ArquintL (approved):
Nice work! I only have a few very minor remarks and/or questions

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1135,6 +1135,37 @@ object Desugar {
             val typ = typeD(info.typ(b), Addressability.exclusiveVariable)(src)
             unit(freshExclusiveVar(typ)(src))
 
+          case PMake(t, args) =>
+            def elemD(t: Type): in.Type = typeD(t, Addressability.defaultValue)(src)
```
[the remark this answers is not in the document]
Felalolf: The idea is that the Addressability unifies the logic regarding when something is addressable or non-addressable. The proper solution would be to add a `make` field to Addressability that has the value Exclusive. I assumed that João just has some interpretation under which `make` produces default values.

src/main/scala/viper/gobra/ast/internal/transform/OverflowChecksTransform.scala
```diff
@@ -97,30 +97,41 @@ object OverflowChecksTransform extends InternalTransform {
...
-
-    case m@Make(_, typ) =>
-      val info = createAnnotatedInfo(typ.info)
-      val assertBounds = Assert(assertionExprInBounds(typ.op, typ.op.typ)(info))(info)
+      Seqn(genOverflowChecksExprs(recv +: args) :+ m)(m.info)
+
+    case m@New(_, expr) =>
+      val info = createAnnotatedInfo(expr.info)
+      val assertBounds = Assert(assertionExprInBounds(expr, expr.typ)(info))(info)
       Seqn(Vector(assertBounds, m))(m.info)
```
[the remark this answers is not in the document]
jcp19: Yes, no doubt! this was an old node called Make that I renamed to New and later forgot to update the code for generating the overlfows

src/main/scala/viper/gobra/frontend/Desugar.scala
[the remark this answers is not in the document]
jcp19: Yes, at first that was my interpretation. I later realized that it was not correct but forgot to update this

src/main/scala/viper/gobra/frontend/Parser.scala
jcp19: I did not find anything about this in the spec but, empirically, I have seen trailing commas working on function/method calls and some other places in *Go*. I found however [this stack overflow entry](https://stackoverflow.com/questions/29353429/golang-commas-with-methods-functions) that poses a question related to yours. Given that the compiler accepts calls like `make([]int, 5,)`, I did not bother to change the parser.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
jcp19: I was under the impression that you could not make send and receive channels for some reason. I still think that it does not make sense to make a channel of these types but I tried`make` with both types and it worked to my surprise. @Felalolf I am sorry if I induced you in error about this.

PS: I found a Github discussion about this same issue(https://github.com/golang/go/issues/2431)

jcp19: I will change this in my next commit

src/main/scala/viper/gobra/translator/encodings/TypeEncoding.scala
jcp19: This is a vestige of a previous implementation, I forgot to change that

src/main/scala/viper/gobra/frontend/Desugar.scala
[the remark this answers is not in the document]
ArquintL: Ah I thought `defaultValue` represents the default addressability (as opposed to represent the addressability of default values) ^^ my bad

src/main/scala/viper/gobra/frontend/Parser.scala
ArquintL: Haha okay somehow reminds me of C++ compilers and undefined behavior ^^

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
ArquintL: I would go with a better error message in case someone tries to create a send or receive only channel but disallowing them seems fine to me

src/main/scala/viper/gobra/frontend/Desugar.scala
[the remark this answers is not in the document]
jcp19: I also thought that was the case before being aware of the existence of a node called `DfltVal` :D

PR #161, opened 2021-01-08
Defunc

Felalolf:
Here the pull request that I was asked for


src/main/scala/viper/gobra/translator/encodings/preds/PredEncoding.scala
```diff
@@ -0,0 +1,175 @@
...
+  /**
+    * Encodes statements.
+    * This includes make-statements.
+    *
+    * The default implements:
+    * [v: *T = make(lit)] -> var z (*T)°; inhale Footprint[*z] && [*z == lit]; [v = z]
+    *
+    * [unfold acc(Q{d1, ..., dk}(e1, ..., en), p)] ->
+    *   exhale acc(eval_S([Q{d1, ..., dk}], [e1], ..., [en]), [p])
+    *   inhale acc(Q(a1, ..., ak), [p])
```
ArquintL, line 94: Is this exhale/inhale sequence the reason why the `eval_S` predicate can be abstract? Also, is this a consequence of the restriction of knowing the base when folding/unfolding?

ArquintL (approved):
Looks good to me. I only have a question to the encoding.
Would be nice to have a mechanism to write encoding test cases but we'll soon find out whether it works :P

src/main/scala/viper/gobra/translator/encodings/preds/PredEncoding.scala
Felalolf: I would state this a bit differently. There is a soundness issue when you do not know the base. But we cannot write a `eval_S` that way in Viper, so we make it abstract. Because it is abstract, we cannot fold and unfold, but have to exhale and inhale instead.

ArquintL: Does this mean that the soundness issue prevents writing eval_S‘s body as a case split over the IDs? I.e. having an unknown case is not enough

jcp19 (approved):
Everything looks fine. I would argue in favor of deleting all dead code before merging with master but that is just a minor observation

PR #162, opened 2021-01-08
Defunc Fixes

ArquintL:
- PredEncoding is added as a typeEncoding
- `instance` only used the arguments but not predicate expression `base`


Felalolf (approved):
No comment (it is hard to get something with zero bugs when you cannot run it).

ArquintL:
It‘s impressive that I only found two ;)

PR #163, opened 2021-01-12
Add defunc frontend

jcp19:
Added front-end support for first class predicates. Currently, there are two quirks that I need to address:
1. predicate constructors are written with the rather ugly and asymmetric delimiters `!<` and `!>`, to avoid any ambiguity in parsing them
2. There is no support for `MPredicate`s in predicate constructors

I do not intend to postpone the implementation of these features. In fact, I plan to implement them next. Nonetheless, I believe that it is better to implement such features in a different PR to allow @ArquintL to make progress while I work on them.

### DIFF
--- a/src/main/scala/viper/gobra/ast/frontend/Ast.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/Ast.scala
@@ -594,7 +594,7 @@ case class PMethodReceivePointer(typ: PNamedOperand) extends PMethodRecvType
 
 case class PFunctionType(args: Vector[PParameter], result: PResult) extends PTypeLit with PScope
 
-case class PPredType(args: Vector[PType]) extends PTypeLit with PScope
+case class PPredType(args: Vector[PType]) extends PTypeLit
 
 case class PInterfaceType(
                            embedded: Vector[PInterfaceName],
--- a/src/main/scala/viper/gobra/ast/frontend/Ast.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/Ast.scala
@@ -488,8 +488,6 @@ case class PMake(typ: PType, args: Vector[PExpression]) extends PActualExpressio
   */
 case class PNew(typ: PType) extends PActualExpression
 
-// TODO: doc
-// TODO: check if types of the params are the correct ones in the type system
 sealed trait PPredCtrBase extends PNode {
   val id: PIdnUse
 }
@@ -596,7 +594,6 @@ case class PMethodReceivePointer(typ: PNamedOperand) extends PMethodRecvType
 
 case class PFunctionType(args: Vector[PParameter], result: PResult) extends PTypeLit with PScope
 
-// TODO: what is PScope? should this be an instance of that
 case class PPredType(args: Vector[PType]) extends PTypeLit with PScope
 
 case class PInterfaceType(
--- a/src/main/scala/viper/gobra/ast/frontend/Ast.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/Ast.scala
@@ -488,15 +488,15 @@ case class PMake(typ: PType, args: Vector[PExpression]) extends PActualExpressio
   */
 case class PNew(typ: PType) extends PActualExpression
 
-sealed trait PPredCtrBase extends PNode {
+sealed trait PPredConstructorBase extends PNode {
   val id: PIdnUse
 }
-case class PFPredBase(override val id: PIdnUse) extends PPredCtrBase
-case class PMPredBase(recvWithId: PDot) extends PPredCtrBase {
+case class PFPredBase(override val id: PIdnUse) extends PPredConstructorBase
+case class PMPredBase(recvWithId: PDot) extends PPredConstructorBase {
   override val id: PIdnUse = recvWithId.id
   val recv: PExpressionOrType = recvWithId.base
 }
-case class PPredConstructor(id: PPredCtrBase, args: Vector[Option[PExpression]]) extends PActualExpression
+case class PPredConstructor(id: PPredConstructorBase, args: Vector[Option[PExpression]]) extends PActualExpression
 
 /**
   * Types
--- a/src/main/scala/viper/gobra/ast/frontend/Ast.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/Ast.scala
@@ -488,18 +488,14 @@ case class PMake(typ: PType, args: Vector[PExpression]) extends PActualExpressio
   */
 case class PNew(typ: PType) extends PActualExpression
 
-/* TODO: implement this (in Program)
-sealed trait PPredCtrBase extends Node
-case class FPredBase(id: IdUse) extends PredCtrBase
-case class PPredicateConstrucotr(id: PPredicateCtrBase, ...)
- */
-
 // TODO: doc
-// TODO: check if types of the params are the correct ones
-case class PFPredConstructor(predId: PIdnUse, args: Vector[Option[PExpression]]) extends PActualExpression
-// TODO: does it also make sense to leave the receiver unnaplied?
-// TODO: implement everything related to this node after having everything done for PFPredConstructor, implement shared trait PPredConstructor
-case class PMPredConstructor(predId: PIdnUse, receiver: Option[PReceiver], args: Vector[Option[PExpression]]) extends PActualExpression
+// TODO: check if types of the params are the correct ones in the type system
+sealed trait PPredCtrBase extends PNode {
+  val id: PIdnUse
+}
+case class PFPredBase(override val id: PIdnUse) extends PPredCtrBase
+case class PMPredBase(override val id: PIdnUse, recv: PExpressionOrType) extends PPredCtrBase
+case class PPredConstructor(id: PPredCtrBase, args: Vector[Option[PExpression]]) extends PActualExpression
 
 /**
   * Types
--- a/src/main/scala/viper/gobra/ast/frontend/Ast.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/Ast.scala
@@ -488,6 +488,19 @@ case class PMake(typ: PType, args: Vector[PExpression]) extends PActualExpressio
   */
 case class PNew(typ: PType) extends PActualExpression
 
+/* TODO: implement this (in Program)
+sealed trait PPredCtrBase extends Node
+case class FPredBase(id: IdUse) extends PredCtrBase
+case class PPredicateConstrucotr(id: PPredicateCtrBase, ...)
+ */
+
+// TODO: doc
+// TODO: check if types of the params are the correct ones
+case class PFPredConstructor(predId: PIdnUse, args: Vector[Option[PExpression]]) extends PActualExpression
+// TODO: does it also make sense to leave the receiver unnaplied?
+// TODO: implement everything related to this node after having everything done for PFPredConstructor, implement shared trait PPredConstructor
+case class PMPredConstructor(predId: PIdnUse, receiver: Option[PReceiver], args: Vector[Option[PExpression]]) extends PActualExpression
+
 /**
   * Types
   */
@@ -586,6 +599,9 @@ case class PMethodReceivePointer(typ: PNamedOperand) extends PMethodRecvType
 
 
 case class PFunctionType(args: Vector[PParameter], result: PResult) extends PTypeLit with PScope
+
+// TODO: what is PScope? should this be an instance of that
+case class PPredType(args: Vector[PType]) extends PTypeLit with PScope
 
 case class PInterfaceType(
                            embedded: Vector[PInterfaceName],
--- a/src/main/scala/viper/gobra/ast/frontend/Ast.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/Ast.scala
@@ -492,7 +492,10 @@ sealed trait PPredCtrBase extends PNode {
   val id: PIdnUse
 }
 case class PFPredBase(override val id: PIdnUse) extends PPredCtrBase
-case class PMPredBase(override val id: PIdnUse, recv: PExpressionOrType) extends PPredCtrBase
+case class PMPredBase(recvWithId: PDot) extends PPredCtrBase {
+  override val id: PIdnUse = recvWithId.id
+  val recv: PExpressionOrType = recvWithId.base
+}
 case class PPredConstructor(id: PPredCtrBase, args: Vector[Option[PExpression]]) extends PActualExpression
 
 /**
--- a/src/main/scala/viper/gobra/ast/frontend/AstPattern.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/AstPattern.scala
@@ -48,5 +48,5 @@ object AstPattern {
   case class Predicate(id: PIdnUse, symb: st.FPredicate) extends PredicateKind with Symbolic
   case class ReceivedPredicate(recv: PExpression, id: PIdnUse, path: Vector[MemberPath], symb: st.MPredicate) extends PredicateKind with Symbolic
   case class PredicateExpr(typ: PType, id: PIdnUse, path: Vector[MemberPath], symb: st.MPredicate) extends PredicateKind with Symbolic
-  case class PredExprInstance() extends PredicateKind
+  case class PredExprInstance(/*base: PExpression, args: Vector[PExpression]*/) extends PredicateKind
 }
--- a/src/main/scala/viper/gobra/ast/frontend/AstPattern.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/AstPattern.scala
@@ -48,5 +48,5 @@ object AstPattern {
   case class Predicate(id: PIdnUse, symb: st.FPredicate) extends PredicateKind with Symbolic
   case class ReceivedPredicate(recv: PExpression, id: PIdnUse, path: Vector[MemberPath], symb: st.MPredicate) extends PredicateKind with Symbolic
   case class PredicateExpr(typ: PType, id: PIdnUse, path: Vector[MemberPath], symb: st.MPredicate) extends PredicateKind with Symbolic
-  case class PredExprInstance(/*symb: st.Predicate*/) extends PredicateKind //with Symbolic what is path? // TODO: adapt for MPred
+  case class PredExprInstance() extends PredicateKind
 }
--- a/src/main/scala/viper/gobra/ast/frontend/AstPattern.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/AstPattern.scala
@@ -48,5 +48,5 @@ object AstPattern {
   case class Predicate(id: PIdnUse, symb: st.FPredicate) extends PredicateKind with Symbolic
   case class ReceivedPredicate(recv: PExpression, id: PIdnUse, path: Vector[MemberPath], symb: st.MPredicate) extends PredicateKind with Symbolic
   case class PredicateExpr(typ: PType, id: PIdnUse, path: Vector[MemberPath], symb: st.MPredicate) extends PredicateKind with Symbolic
-  case class PredExprInstance(/*base: PExpression, args: Vector[PExpression]*/) extends PredicateKind
+  case class PredExprInstance(base: PExpression, args: Vector[PExpression]) extends PredicateKind
 }
--- a/src/main/scala/viper/gobra/ast/frontend/AstPattern.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/AstPattern.scala
@@ -48,5 +48,5 @@ object AstPattern {
   case class Predicate(id: PIdnUse, symb: st.FPredicate) extends PredicateKind with Symbolic
   case class ReceivedPredicate(recv: PExpression, id: PIdnUse, path: Vector[MemberPath], symb: st.MPredicate) extends PredicateKind with Symbolic
   case class PredicateExpr(typ: PType, id: PIdnUse, path: Vector[MemberPath], symb: st.MPredicate) extends PredicateKind with Symbolic
-
+  case class PredExprInstance(/*symb: st.Predicate*/) extends PredicateKind //with Symbolic what is path? // TODO: adapt for MPred
 }
--- a/src/main/scala/viper/gobra/frontend/Desugar.scala
+++ b/src/main/scala/viper/gobra/frontend/Desugar.scala
@@ -1692,7 +1692,7 @@ object Desugar {
         case PUnfold(exp) =>
           info.resolve(exp.pred) match {
             case Some(_: ap.PredExprInstance) => for {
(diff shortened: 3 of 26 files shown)


src/main/scala/viper/gobra/ast/frontend/AstPattern.scala
```diff
@@ -48,5 +48,5 @@ object AstPattern {
   case class Predicate(id: PIdnUse, symb: st.FPredicate) extends PredicateKind with Symbolic
   case class ReceivedPredicate(recv: PExpression, id: PIdnUse, path: Vector[MemberPath], symb: st.MPredicate) extends PredicateKind with Symbolic
   case class PredicateExpr(typ: PType, id: PIdnUse, path: Vector[MemberPath], symb: st.MPredicate) extends PredicateKind with Symbolic
-
+  case class PredExprInstance() extends PredicateKind
```
Felalolf, line 51: I think that this should take the base expression of the predicate instance (predicate call) as a parameter.

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
```diff
@@ -64,6 +64,7 @@ trait Assignability extends BaseProperty { this: TypeInfoImpl =>
 
         // for ghost types
       case (BooleanT, AssertionT) => true
+      case (PredT(Vector()), AssertionT) => true
```
Felalolf, line 67: That case is wrong. Only a predicate instance is an assertion.

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -768,6 +771,27 @@ object Parser {
         ghostPrimaryExp |
         operand
 
+    // TODO: change delimiter to { and implement required ambiguity resolution, current: !< X, ..., Z !>
+    // declaredPred!<d1, ..., dn!>
+    lazy val fpredConstruct: Parser[PPredConstructor] =
+      (idnUse ~ predConstructArgs) ^^ { p => PPredConstructor(PFPredBase(p._1), p._2) }
+
+    lazy val mpredConstruct: Parser[PPredConstructor] =
+      ???
```
Felalolf, line 780: I am not sure that you can use this as an implementation. Wouldn't the parser throw an exception if it is evaluated? You could use the failing parser.

Felalolf: I see now,  currently, you use `fpredConsturct` instead of `predConstruct`. That is fine.

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -768,6 +771,27 @@ object Parser {
...
+
+    lazy val predConstructArgs: Parser[Vector[Option[PExpression]]] =
+      ("!<" ~> (rep1sep(predConstructArg, ",") <~ ",".?).? <~ "!>") ^^ (opt => opt.getOrElse(Vector.empty))
+
+    lazy val predConstructArg: Parser[Option[PExpression]] =
+      (expression ^^ Some[PExpression]) | ("_" ^^^ None)
+
+    // TODO: add support for mpred, change fpredConstruct to predConstruct
+    lazy val predExprInstance: Parser[PExpression] =
+      fpredConstruct ~ callArguments ^^ PInvoke
```
Felalolf, line 793: Can you not just delete this case? This should already be covered by the invoke parser.

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -587,6 +594,8 @@ case class PMethodReceivePointer(typ: PNamedOperand) extends PMethodRecvType
 
 case class PFunctionType(args: Vector[PParameter], result: PResult) extends PTypeLit with PScope
 
+case class PPredType(args: Vector[PType]) extends PTypeLit with PScope
```
Felalolf, line 597: Because the predicate type now only takes types and no named parameters, you can remove the PScope.

src/main/scala/viper/gobra/frontend/info/implementation/resolution/AmbiguityResolution.scala
```diff
@@ -89,6 +89,7 @@ trait AmbiguityResolution { this: TypeInfoImpl =>
           resolve(e) match {
             case Some(p: ap.FunctionKind) => Some(ap.FunctionCall(p, n.args))
             case Some(p: ap.PredicateKind) => Some(ap.PredicateCall(p, n.args))
+            case _ if exprType(e).isInstanceOf[PredT] => Some(ap.PredExprInstance())
```
Felalolf, line 92: Here, I would put the expression e into the `ap.PredExprInstance()` (as mentioned in an earlier comment).

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostStmtTyping.scala
```diff
@@ -20,7 +21,17 @@ trait GhostStmtTyping extends BaseTyping { this: TypeInfoImpl =>
...
+    case PUnfold(acc) => wellDefFoldable(acc)
   }
+
+  private def wellDefFoldable(acc: PPredicateAccess): Messages =
+    resolve(acc.pred) match {
+      case Some(_: ap.PredExprInstance) =>
+        error(
+          acc,
+          s"expected a predicate constructor, but got ${acc.pred.base}",
+          !acc.pred.base.isInstanceOf[PPredConstructor])
```
Felalolf, line 34: great

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
Felalolf: I know understand why you did that. You changed that predicates have a predicate type, instead of a function type that returns an assertion, but this is wrong. It should stay the old way.

src/main/scala/viper/gobra/frontend/info/implementation/resolution/Enclosing.scala
```diff
@@ -82,7 +82,7 @@ trait Enclosing { this: TypeInfoImpl =>
           case i: PInvoke => (exprOrType(i.base), resolve(i)) match {
             case (Right(target), Some(_: ap.Conversion)) => Some(symbType(target))
             case (Left(callee), Some(p: ap.FunctionCall)) => Some(typ(callee).asInstanceOf[Type.FunctionT].args(p.args.indexOf(n)))
-            case (Left(callee), Some(p: ap.PredicateCall)) => Some(typ(callee).asInstanceOf[Type.FunctionT].args(p.args.indexOf(n)))
+            case (Left(callee), Some(p: ap.PredicateCall)) => Some(typ(callee).asInstanceOf[Type.PredT].args(p.args.indexOf(n)))
```
Felalolf, line 85: reverse (see comment in Assignability)

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -78,13 +78,13 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
-          case f: FunctionT => extentFunctionType(f, typeSymbType(p.typ))
+          case f: FunctionT => extendFunctionType(f, typeSymbType(p.typ))
           case t => violation(s"a method should be typed to a function type, but got $t")
         }
 
         case Some(p: ap.PredicateExpr) => memberType(p.symb) match {
-          case f: FunctionT => extentFunctionType(f, typeSymbType(p.typ))
-          case t => violation(s"a predicate should be typed to a function type, but got $t")
+          case e: PredT => extendPredicateType(e, typeSymbType(p.typ))
+          case t => violation(s"a predicate should be typed to a predicate type, but got $t")
```
Felalolf, line 87: reverse (see comment in Assignability)

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -94,7 +94,7 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
         }
         case Some(p: ap.Function) => FunctionT(p.symb.args map p.symb.context.typ, p.symb.context.typ(p.symb.result))
         case Some(_: ap.NamedType) => SortT
-        case Some(p: ap.Predicate) => FunctionT(p.symb.args map p.symb.context.typ, AssertionT)
+        case Some(p: ap.Predicate) => PredT(p.symb.args map p.symb.context.typ)
```
Felalolf, line 97: reverse (see comment in Assignability)

Felalolf: I will stop writing that, you get the gist of it.

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
Felalolf: To give a simple example, the following is not ok, but might work currently:

```
requires acc(x) && p
func foo(x *int, p pred())
```

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostMiscTyping.scala
```diff
@@ -29,8 +29,8 @@ trait GhostMiscTyping extends BaseTyping { this: TypeInfoImpl =>
   }
 
   private[typing] def ghostMemberType(typeMember: GhostTypeMember): Type = typeMember match {
-    case MPredicateImpl(decl, _) => FunctionT(decl.args map miscType, AssertionT)
-    case MPredicateSpec(decl, _) => FunctionT(decl.args map miscType, AssertionT)
+    case MPredicateImpl(decl, _) => PredT(decl.args map miscType)
+    case MPredicateSpec(decl, _) => PredT(decl.args map miscType)
```
Felalolf, line 33: Should be reversed.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostIdTyping.scala
```diff
@@ -26,7 +26,7 @@ trait GhostIdTyping { this: TypeInfoImpl =>
   private[typing] def ghostEntityType(entity: GhostRegular, id: PIdnNode): Type = entity match {
 
     case x: BoundVariable => typeSymbType(x.decl.typ)
-    case predicate: Predicate => FunctionT(predicate.args map predicate.context.typ, AssertionT)
+    case predicate: Predicate => PredT(predicate.args map predicate.context.typ)
```
Felalolf, line 29: reverse

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1165,6 +1165,16 @@ object Desugar {
               _ <- write(in.New(target, zero)(src))
             } yield target
 
+          case PPredConstructor(base, args) => base match {
+            case PFPredBase(id) =>
+              val proxy = fpredicateProxyD(id)
+              val idT = typeD(info.typ(id), Addressability.rValue)(src).asInstanceOf[in.PredT]
```
Felalolf, line 1171: When you reverse the typing of predicaten, then here you will have to transforma a Functiontype to a predicate type.

src/main/scala/viper/gobra/ast/frontend/AstPattern.scala
Felalolf: and the arguments of the instance are also missing.

Felalolf: e.g. `case class PredExprInstance(base: PExpression, args: Vector[PExpression]) extends PredicateKind`

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1649,8 +1661,27 @@ object Desugar {
         case PAssume(exp) => for {e <- goA(exp)} yield in.Assume(e)(src)
         case PInhale(exp) => for {e <- goA(exp)} yield in.Inhale(e)(src)
         case PExhale(exp) => for {e <- goA(exp)} yield in.Exhale(e)(src)
-        case PFold(exp)   => for {e <- goA(exp)} yield in.Fold(e.asInstanceOf[in.Access])(src)
-        case PUnfold(exp) => for {e <- goA(exp)} yield in.Unfold(e.asInstanceOf[in.Access])(src)
+        case PFold(exp)   =>
+          info.resolve(exp.pred) match {
+            case Some(_: ap.PredExprInstance) => for {
+              // the type system guarantees that a pred expr instance in a Fold is in format predName{p1,...,pn}(a1, ...., am)
```
Felalolf, line 1667: not the type system, but the welldefined checks guarantee that ...

src/main/scala/viper/gobra/ast/frontend/AstPattern.scala
jcp19: I was a bit hesitant to add these two fields because I was always able to retrieve them in other ways but I agree that it is useful to have this information here

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
jcp19: Yes, you are right! I will revert all related changes

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -166,8 +166,8 @@ object Desugar {
         case NoGhost(x: PConstDecl) => constDeclD(x)
         case NoGhost(x: PMethodDecl) => Vector(registerMethod(x))
         case NoGhost(x: PFunctionDecl) => Vector(registerFunction(x))
-        case x: PMPredicateDecl => Vector(mpredicateD(x))
-        case x: PFPredicateDecl => Vector(fpredicateD(x))
+        case x: PMPredicateDecl => Vector(registerMPredicate(x))
```
ArquintL, line 169: Why wasn't it necessary to register predicates before but now it is?

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -768,6 +769,25 @@ object Parser {
         ghostPrimaryExp |
         operand
 
+    // TODO: change delimiter to { and implement required ambiguity resolution, current: !< X, ..., Z !>
+    // declaredPred!<d1, ..., dn!>
+    lazy val fpredConstruct: Parser[PPredConstructor] =
+      (idnUse ~ predConstructArgs) ^^ { p => PPredConstructor(PFPredBase(p._1), p._2) }
```
ArquintL, line 775:
```suggestion
      (idnUse ~ predConstructArgs) ^^ { 
        case identifier ~ args => PPredConstructor(PFPredBase(identifier).at(identifier), args) 
      }
```
I hope this compiles ^^
There are two improvements: (1) getting rid of the tuple and (2) add position information to the `PFPredBase` which is important if you create a nested node yourself (check out other occurrences of `.at(` in the parser)

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -768,6 +769,25 @@ object Parser {
...
+
+    lazy val mpredConstruct: Parser[PPredConstructor] =
+      (expression <~ ".") ~ idnUse ~ predConstructArgs ^^ {
```
ArquintL, line 778: Does `(expression <~ ".") ~ idnUse` work? I have the impression that `expression` might already parse the ".". You could try to replace this first part by `selection` which parses a `PDot`. Note that PDot allows a type in front of "." but it would be fine for me that the type checker currently does not support mpredicate constructors that take the receiver expression as an additional argument

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -768,6 +769,25 @@ object Parser {
...
+        case recv ~ id ~ args => PPredConstructor(PMPredBase(id, recv), args)
```
ArquintL, line 779:
```suggestion
        case recv ~ id ~ args => PPredConstructor(PMPredBase(id, recv).range(id, recv), args)
```
Same issue with position information as above

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -396,6 +427,14 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
 
     case PMake(typ, _) => typeSymbType(typ)
 
+    case PPredConstructor(base, args) =>
+      base match {
+        case PFPredBase(id) =>
+          val idT = idType(id)
+          PredT(idT.asInstanceOf[FunctionT].args diff args.filter(_.isDefined).map(x => exprType(x.get)))
```
ArquintL, line 434: Is `diff` the right thing to do here? As diff takes the multiset difference between the function arguments and the applied arguments I think there might be problems if the same type occurs multiple times on the left or right

src/main/scala/viper/gobra/frontend/info/implementation/typing/MiscTyping.scala
```diff
@@ -115,7 +115,10 @@ trait MiscTyping extends BaseTyping { this: TypeInfoImpl =>
     }
 
   /** extends a function type by adding a type as the first argument type **/
-  def extentFunctionType(functionT: FunctionT, base: Type): Type = FunctionT(base +: functionT.args, functionT.result)
+  def extendFunctionType(functionT: FunctionT, base: Type): Type = FunctionT(base +: functionT.args, functionT.result)
+
+  /** extends a predicate type by adding a type as the first argument type **/
+  def extendPredicateType(predT: PredT, base: Type): Type = PredT(base +: predT.args)
```
ArquintL, line 121: is this function still needed after Felix' suggestions?

src/main/scala/viper/gobra/frontend/info/implementation/typing/TypeTyping.scala
```diff
@@ -52,6 +52,7 @@ trait TypeTyping extends BaseTyping { this: TypeInfoImpl =>
     case n: PMethodReceiveName => isType(n.typ).out
     case n: PMethodReceivePointer => isType(n.typ).out
     case _: PFunctionType => noMessages // parameters and result is implied by well definedness of children
+    case _: PPredType => noMessages // well defidedness implied by well defidedness of children
```
ArquintL, line 55: It's called "definedness"

src/test/resources/regressions/features/defunc/mutex1.gobra
```diff
@@ -0,0 +1,67 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+// These examples were designed and proposed by Linard Arquint
+package pkg
+
+// Mutex definition
+type Mutex struct {
```
ArquintL, line 8: we'll need to rewrite this one as soon as we have the mutex stub but leave it for now

ArquintL (approved):

src/test/resources/regressions/features/defunc/mutex1.gobra
jcp19: Won't that be true only if we import the `sync` package which is not done in this test?

ArquintL: True but that would make imho sense

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -488,6 +488,13 @@ case class PMake(typ: PType, args: Vector[PExpression]) extends PActualExpressio
   */
 case class PNew(typ: PType) extends PActualExpression
 
+sealed trait PPredCtrBase extends PNode {
```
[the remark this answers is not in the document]
Felalolf: "constructor", That is my amazing abreviation.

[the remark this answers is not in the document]
ArquintL: Hahaha I was thinking of "Control" or something like that. I would prefer `PPredConstructorBase`

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1649,8 +1665,27 @@ object Desugar {
         case PAssume(exp) => for {e <- goA(exp)} yield in.Assume(e)(src)
         case PInhale(exp) => for {e <- goA(exp)} yield in.Inhale(e)(src)
         case PExhale(exp) => for {e <- goA(exp)} yield in.Exhale(e)(src)
-        case PFold(exp)   => for {e <- goA(exp)} yield in.Fold(e.asInstanceOf[in.Access])(src)
-        case PUnfold(exp) => for {e <- goA(exp)} yield in.Unfold(e.asInstanceOf[in.Access])(src)
+        case PFold(exp)   =>
+          info.resolve(exp.pred) match {
+            case Some(_: ap.PredExprInstance) => for {
+              // the well-defidedness checks guarantees that a pred expr instance in a Fold is in format predName{p1,...,pn}(a1, ...., am)
```
Felalolf, line 1684: well-definedness

src/test/resources/regressions/features/defunc/mutex1.gobra
jcp19: sure, let's change this then once we add the mutex stubs

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
Felalolf: Yeah, you should probably zip and then collect.

src/main/scala/viper/gobra/frontend/Desugar.scala
jcp19: If I am not mistaken, Felix's encoding performs lookups for predicate proxys in the LookupTable available in the `Context` trait. That table is filled here. Before we had defunctionalization, we didn't perform such lookups.

src/main/scala/viper/gobra/frontend/Parser.scala
jcp19: This was a local change that I did not intend to commit. Anyhow, thanks for the `selection` suggestion, that was helpful!

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -472,6 +512,8 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
               case _ => None
             }
 
+          case Some(_: ap.PredExprInstance) => ??? // TODO
```
ArquintL, line 515: This would be another task for the second PR in which you're adapting the constructor's syntax

ArquintL (approved):
Looks good to me. I've left one comment for a "TODO" that is left in the code but you do not need to address it as part of this PR

Felalolf (approved):

PR #167, opened 2021-01-14
Fix issue 166

### LINKED ISSUE #166: Problems with selective importing
The content of imported packages is loaded lazily. While working on the stubs, I came across this bug. I have two source files `A` and `B` from different packages and `A` contains a call to a function defined in `B`.

- File `A`
```go
package pkg

import "bug"

func main() {
    bug.F1()
}
```

- File `B`
```go
package bug

pred Test() { true }

pure func f() bool {
    return true
}

requires Test()
func F1() {}

requires f()
func F2() {}
```

If I verify file `A` in Gobra, I get the following error
```
[error] Exception in thread "main" java.util.concurrent.ExecutionException: java.lang.RuntimeException: Predicate name Test_bug_F not found in program.
[error] 	at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122)
[error] 	at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:191)
[error] 	at viper.silicon.Silicon.verify(Silicon.scala:199)
[error] 	at viper.silicon.Silicon.verify(Silicon.scala:157)
[error] 	at viper.gobra.backend.Silicon.$anonfun$verify$1(Silicon.scala:27)
[error] 	at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:672)
[error] 	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:431)
[error] 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
[error] 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
[error] 	at java.base/java.lang.Thread.run(Thread.java:834)
[error] Caused by: java.lang.RuntimeException: Predicate name Test_bug_F not found in program.
[error] 	at scala.sys.package$.error(package.scala:27)
[error] 	at viper.silver.ast.Program.findPredicate(Program.scala:270)
[error] 	at viper.silicon.supporters.qps.PredicateSnapGenerator$$anonfun$setup$1.applyOrElse(PredicateSnapGenerator.scala:26)
[error] 	at viper.silicon.supporters.qps.PredicateSnapGenerator$$anonfun$setup$1.applyOrElse(PredicateSnapGenerator.scala:24)
```
This occurs because the `Test` predicate is not included in the final viper file because it is not explicitly accessed. If the `main` function of `A` is changed to
```go
func main() {
    fold bug.Test() // added explicit use of the Test predicate
    bug.F1()
}
```
then everything works again.

An analogous case occurs if I replace `F1` with `F2` in the `main` function of file `A` because the function `f()` accessed in the contract of `F2` is also not explicitly mentioned in `A`, and thus, it is not added to the final code.

This has subtle consequences in case we ever opt to follow Go's practice of exporting only the upper case names, because we would need to **guarantee that all entities accessed in the contract of an exported function/method are also exportable** (i.e. have an upper case name).

A **potential fix** for this would be to perform a transitive closure to detect all dependencies of a function contract or else, to import everything in the final viper file.

jcp19:
Temporary workaround for issue #166 by disabling selective importing, all contents of an imported package appear in the final viper output

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Desugar.scala
+++ b/src/main/scala/viper/gobra/frontend/Desugar.scala
@@ -26,7 +26,7 @@ object Desugar {
       val typeInfo: TypeInfo = tI.getTypeInfo
       val importedPackage = typeInfo.tree.originalRoot
       val d = new Desugarer(importedPackage.positions, typeInfo)
-      (d, d.packageD(importedPackage, tI.isUsed))
+      (d, d.packageD(importedPackage))
     }}
     // desugar the main package, i.e. the package on which verification is performed:
     val mainDesugarer = new Desugarer(pkg.positions, info)
--- a/src/test/resources/regressions/issues/000166.gobra
+++ b/src/test/resources/regressions/issues/000166.gobra
@@ -1,3 +1,6 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
 package pkg
 
 // ##(-I src/test/resources/regressions/issues/000166/)
--- a/src/test/resources/regressions/issues/000166.gobra
+++ b/src/test/resources/regressions/issues/000166.gobra
@@ -1,0 +1,11 @@
+package pkg
+
+// ##(-I src/test/resources/regressions/issues/000166/)
+
+import "bug"
+
+func main() {
+	//:: ExpectedOutput(precondition_error)
+	bug.F1()
+	bug.F2()
+}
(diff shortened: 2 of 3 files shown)


ArquintL (approved):
Do you want to close the issue? For me it would make more sense to just reference the issue and say that this is a quick workaround such that we do not forget about it later

jcp19:
> Do you want to close the issue? For me it would make more sense to just reference the issue and say that this is a quick workaround such that we do not forget about it later

Yes, that is indeed better. The issue will remain open

PR #168, opened 2021-01-15
Channels

### LINKED ISSUE #142: Support for Channels
TODO

ArquintL:
Sorry for the large diff!
This PR contributes in two ways: (1) A mechanism to easily add built-in functions, methods, fpredicates, and mpredicates and (2) the channel encoding based on Martin's concurrency outline.

## Built-in Members
The core is `viper.gobra.frontend.info.base.BuiltInMemberTag`. This file declares several tags that represent built-in members as well as their associated types.
There is a new type `AbstractType` ~~are two new types, `AuxType` and `SingleAuxType`~~ that represent types that depend on more resp. one other type(s). Their particular use-case right now are the built-in members and allow in particular to make a built-in member's type depending on the arguments resp. receiver with which it is called.
Note that `AbstractType` depends on the argument types for functions and fpredicates but only on the singleton vector containing the receiver's type for methods and mpredicates.
~~Note that `AuxType` is used for functions and fpredicates and allow for a dependency on the types of all arguments. However, methods and mpredicates are restricted to have a dependency only on their receiver's type as this heavily simplifies the member resolution. `SingleAuxType` is used for them.~~
I've added a short description to the top of `BuiltInMemberTag` that lists the steps needed to add a built-in member.

### Drawbacks
Imho there are certain aspects that maybe could solved more nicely:
- ~~Split between BuiltInMemberTag and Desugarer: When adding a new built-in member changes have to be performed in the Desugarer as well. It would be nicer that the generation of the member would happen in the BuiltInMemberTag file as well.~~
- There is still some coupling between `BuiltInMemberTag` and `BuiltInMembersImpl` since e.g. input parameters have to be created for the parameter types. However, I don't think it's possible to completely get rid of this coupling since the encoding depends on the parameters (and their semantic). ~~Loose coupling of information associated with built-in members: The tag's getters as well as `BuiltInMemberTag.types(...)` provide ghostness, pureness, arity, and typing information. However, this information is not reused in the desugarer when e.g. creating the particular parameters for a member. It would be nice that one would need to specify this information only once and it would also solve the problem that the information gets out-of-sync meaning that the desugarer generates a member that does not fit the types used during typing.~~

## Channels
Channels are encoded using the following functions, methods, fpredicates, and mpredicates:
```
// fpredicates
pred PredTrue([args]) {
    true
}


// mpredicates
pred (c chan T) IsChannel(k Int)
pred (c chan T) SendChannel()
pred (c chan T) RecvChannel()
pred (c chan T) Closed()
pred (c chan T) ClosureDebt(P pred(), dividend int, divisor int)
pred (c chan T) Token(P pred())


// functions
requires divisor > 0
requires acc(c.SendChannel(), dividene/divisor) && c.ClosureDebt(P, divisor - dividend, divisor) && P()
ensures c.Closed()
func close(c chan T, ghost dividend int, divisor int, P pred())


// methods
requires acc(c.SendChannel(), _)
pure func (c chan T).SendGivenPerm(res pred(T))

requires acc(c.SendChannel(), _)
pure func (c chan T).SendGotPerm(res pred())

requires acc(c.RecvChannel(), _)
pure func (c chan T).RecvGivenPerm(res pred())

requires acc(c.RecvChannel(), _)
pure func (c chan T).RecvGotPerm(res pred(T))

requires c.IsChannel(k)
requires k > 0 ==> (B == pred_true{} && C == pred_true{})
requires A == D && B == C
ensures c.SendChannel() && c.RecvChannel()
ensures c.SendGivenPerm() == A && c.SendGotPerm() == B
ensures c.RecvGivenPerm() == C && c.RecvGotPerm() == D
ghost func (c chan T).Init(k int, A pred(T), B pred(), C pred(), D pred(T))

requires divisor > 0
requires acc(c.SendChannel(), dividend/divisor)
ensures c.ClosureDebt(P, dividend, divisor) && c.Token(P)
ghost func (c chan T) CreateDebt(dividend int, divisor int, P pred())

requires c.Token(P) && acc(c.Closed(), _)
ensures c.Closed() && P()
ghost func (c chan T) Redeem(P pred())


// make stmt
[r := make(chan T, bufferSize)] ->
    assert 0 <= [bufferSize]
    var a [ chan T ]
    inhale [a].isChannel([bufferSize])
    r := a

// send stmt
[(c : channel[T] <- (m : T)] ->
    assert acc(SendChannel([c], wildcard)
    exhale [c].SendGivenPerm()([m])
    inhale [c].SendGotPerm()()

// receive operation (expression)
R[ <- (c: channel[T] ] ->
    exhale acc([c].RecvChannel(), wildcard)
    exhale [c].RecvGivenPerm()()
    var res [ T ]
    inhale [c].RecvChannel()
    // if a non-zero value is received, this implies that the channel is not closed and thus the invariant holds:
    // if a zero value is received, no knowledge about the channel's close state is gained
    inhale res != Dflt[T] ==> [c].RecvGotPerm()(res)
    res

// safe receive operation (stmt)
[resTarget, successTarget := <- (c : channel[T])] ->
    exhale acc([c].RecvChannel(), wildcard)
    exhale [c].RecvGivenPerm()()
    var res [ T ]
    var ok Bool
    inhale [c].RecvChannel()
    inhale ok ==> [c].RecvGotPerm()(res)
    inhale not(ok) ==> [c].Closed() && res == Dflt[T]
    resTarget := res
    successTarget := ok
```
Note that the last 4 operations (make, send, recv, and safe recv are present in the internal representation and will be encoded by `ChannelEncoding`. All others are handled by `BuiltInMembersImpl` and get translated to 'normal' program members which are then encoded.

### Drawbacks
- Encoding is now fully located in the encoding part of Gobra but just split among two files. ~~Encoding is split across `BuiltInMemberTag`, the desugarer, and `ChannelEncoding`. Would be nice to have them at the same place~~
- No meaningful error messages yet for failing preconditions. It would be of great help to know which part did fail since looking up the signature of the built-in members is cumbersome.
- Debt redistribution is heavily simplified and the viewshift in make's precondition is replaced by the condition `A == D && B == C`

### Interesting Test Cases
- `channel-simple6.gobra`: Example transferring a permission with the close operation from the go-routine closing the channel back to `main`. The assertion stmts illustrate pre- and postconditions of the invoked functions/methods.
- `channel-simple7.gobra`: Example sending an int pointer together with its permissions to a different go-routine and back

Closes #142


src/main/scala/viper/gobra/frontend/info/base/BuiltInMemberTag.scala
```diff
@@ -0,0 +1,339 @@
...
+      {
+        case (_, _) => noMessages // it is well-defined for arbitrary arguments
+      },
+      {
+        case args => FunctionT(args, AssertionT)
+      })
+    case _ => unknownTagAuxType(tag)
+  }
+
+  def types(tag: BuiltInSingleAuxTypeTag)(config: Config): SingleAuxType = tag match {
```
jcp19, line 204: Having to extend monolithic methods (`types(...)`, `argGhostTyping(...)`, and `returnGhostTyping(...)`) when we add new tags seems a bit error-prone to me, despite the fact that this is well-documented in the beginning of the file. Is there any reason for these methods to not be defined in the traits `BuiltInAuxTypeTag` and `BuiltInSingleAuxTypeTag`? In such a scenario, we could modify those traits to look like
```scala
sealed trait BuiltInAuxTypeTag extends BuiltInMemberTag {
    def typ(config: Config): AuxType
    def argGhostTyping(args: Vector[Type])(config: Config): GhostType
    def returnGhostTyping(args: Vector[Type])(config: Config): GhostType
}
```
and similarly for `BuiltInAuxTypeTag`.

In that case, we would override the methods every time we define a new Tag and the compiler would enforce that we implement those methods.

src/main/scala/viper/gobra/translator/encodings/channels/ChannelEncoding.scala
```diff
@@ -0,0 +1,248 @@
...
+  override def statement(ctx: Context): in.Stmt ==> CodeWriter[vpr.Stmt] = {
+    default(super.statement(ctx)){
+      case makeStmt@in.MakeChannel(target, in.ChannelT(typeParam, _), optBufferSizeArg, isChannelPred) =>
+        val (pos, info, errT) = makeStmt.vprMeta
+        val a = in.LocalVar(Names.freshName, in.ChannelT(typeParam.withAddressability(Addressability.channelElement), Addressability.Exclusive))(makeStmt.info)
+        val vprA = ctx.typeEncoding.variable(ctx)(a)
+        val bufferSizeArg = optBufferSizeArg.getOrElse(in.IntLit(0)(makeStmt.info)) // create an unbuffered channel by default
+        seqn(
+          for {
+            // var a [ []T ]
```
jcp19, line 127: Don't you mean `chan T` instead of `[]T`?

jcp19 (commented):
Awesome work! I left some comments about some things that might be improved. Also, @Felalolf should probably take a look at this once we gets better and before we merge this massive update :)

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -762,6 +773,15 @@ case class Conversion(newType: Type, expr: Expr)(val info: Source.Parser.Info) e
   override def typ: Type = newType
 }
 
+/**
+  * Represents a channel receive operation. Currently does not appear in the internal presentation as it will
```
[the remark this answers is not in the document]
ArquintL: Actually this is no longer true

src/main/scala/viper/gobra/theory/Addressability.scala
```diff
@@ -64,6 +65,7 @@ object Addressability {
   val sliceElement: Addressability = arrayElement(pointerBase)
   val mapValue: Addressability = Exclusive
   val mathDataStructureElement: Addressability = Exclusive
+  val channelElement: Addressability = Exclusive
```
Felalolf, line 68: Yes!

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -2112,6 +2603,35 @@ object Desugar {
...
+      case in.SortT => "Sort"
+      case in.ArrayT(len, elemT, _) => s"Array$len${stringifyType(elemT)}"
+      case in.SliceT(elemT, _) => s"Slice${stringifyType(elemT)}"
+      case in.SequenceT(elemT, _) => s"Sequence${stringifyType(elemT)}"
+      case in.SetT(elemT, _) => s"Set${stringifyType(elemT)}"
+      case in.MultisetT(elemT, _) => s"Multiset${stringifyType(elemT)}"
+      case in.OptionT(elemT, _) => s"Option${stringifyType(elemT)}"
+      case in.DefinedT(name, _) => s"Defined$name"
+      case in.PointerT(t, _) => s"Pointer${stringifyType(t)}"
+      case in.TupleT(ts, _) => s"Tuple${ts.map(stringifyType).mkString("")}"
```
Felalolf, line 2620: I do not know whether that this is particularly important, but this transformation is not injective.
`Tuple(Tuple(X), Y)` and `Tuple(Tuple(X, Y))` maps to the same.

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -2031,6 +2268,259 @@ object Desugar {
       for { exprs <- sequence(trigger.exps map exprD(ctx)) } yield in.Trigger(exprs)(src)
     }
 
+    def generateBuiltInFunction(tag: BuiltInFunctionTag, args: Vector[in.Type])(src: Meta): in.FunctionMember = {
```
Felalolf, line 2271: I did not check any of the generateBuiltIng methods.

src/main/scala/viper/gobra/frontend/info/base/BuiltInMemberTag.scala
```diff
@@ -0,0 +1,339 @@
...
+  case object TokenMPredTag extends BuiltInMPredicateTag {
+    override def identifier: String = "Token"
+    override def name: String = "TokenMPredTag"
+  }
+
+
+  /**
+    * Returns a vector of tags belonging to built-in members that should be considered during name resolution
+    */
+  def builtInMembers(): Vector[BuiltInMemberTag] = Vector(
```
Felalolf, line 161: That are a lot of function that are necessary for channels ... wow.

src/main/scala/viper/gobra/translator/encodings/channels/ChannelEncoding.scala
```diff
@@ -0,0 +1,248 @@
...
+  /**
+    * Encodes expressions as values that do not occupy some identifiable location in memory.
+    *
+    * R[ dflt(channel[T]) ] -> 0
+    * R[ nil : channel[T] ] -> 0
+    * R[ <- (c: channel[T] ] ->
+    *   exhale acc([c].RecvChannel(), wildcard)
+    *   exhale [c].RecvGivenPerm()()
+    *   var res [ T ]
+    *   inhale [c].RecvChannel()
```
Felalolf, line 44: If I see this right, then the permission amount increases everytime something is received.
How about `exhale acc([c].RecvChannel(), perm([c].RecvChannel()))` instead?

Felalolf, line 44: I am surprised that `res` does not occur anywhere else.

src/main/scala/viper/gobra/translator/encodings/channels/ChannelEncoding.scala
```diff
@@ -0,0 +1,248 @@
...
+  /**
+    * Encodes
+    *  - allocation of a new channel
+    *  - channel receive operation that returns received message as well as a success flag
+    *
+    * [r := make(chan T, bufferSize)] ->
+    *   assert 0 <= [bufferSize]
+    *   var a [ chan T ]
+    *   inhale [a].isChannel([bufferSize])
+    *   r := a
```
Felalolf, line 97: In the current encoding, we cannot derive that a channel is not nil if we have some permissions associated to it. Shortterm this might not matter, but longterm we should figure this out. Maybe, if you are not fixing this now, could you please create an issue.

Felalolf (approved):
I did not look at everything in great detail, but it seems fine to me.

src/main/scala/viper/gobra/frontend/Desugar.scala
Felalolf: Maybe one comment. If I understood this correctly, then you are essentially adding the encoding of the buitlin operations at the desugarer. I am not entirely sure whether this is a good idea. For this PR it is fine, but in generel the idea is that the desugarer does not perform the encoding. Instead, I would just generate the methods, predicates, and functions in the typeencoding of channels.

ArquintL: Does this imply that the internal representation needs a node that stores the tag and the specific types with which it is instantiated such that the encoding can handle specific tags?

src/main/scala/viper/gobra/translator/encodings/channels/ChannelEncoding.scala
ArquintL: According to Martin, this permission is duplicable and this treatment should simplify the encoding (I don‘t know which version would lead to more performance).

ArquintL: This read operation is indeed pretty useless but I support it for completeness reasons. 
I basically use the same contract as for the safe receive operation but since the success flag is missing all implications in the postcondition can not be expressed. Hence, an arbitrary value (incl. the zero value) is received and one cannot distinguish receiving a ‚valid‘ value and receiving from a closed channel

src/main/scala/viper/gobra/translator/encodings/channels/ChannelEncoding.scala
ArquintL: As we encoding the channel type as an int would this mean that we ‚reserve‘ 0 as the channel‘s nil value and would add contracts that ensure that the channel is != 0?

src/main/scala/viper/gobra/frontend/Desugar.scala
Felalolf: All function and method calls and predicate instances are translated to the usual internal AST node. Then you just have to make sure that the corresponding methods, functions, and predicates are generates. So, from my point of view, no additional nods is needed.

src/main/scala/viper/gobra/translator/encodings/channels/ChannelEncoding.scala
Felalolf: Yeah, yeah, we can think about that at a later point in time. I also do not know what is better for performance. We should evaluate this at some point and then make changes accordingly.

Felalolf: But then there are restrictions on res, it is a zero value or a value that satisfies the channel invariant. Therefore, there should be an assumption on res.

src/main/scala/viper/gobra/translator/encodings/channels/ChannelEncoding.scala
Felalolf: I am not sure if I would call it "adding contracts". For Ref, we know that `acc(x.f)` implies that `x` is not null. This could also be achieved by replacing `acc(x.f)` with `acc(x.f) && x != null`. In theory, we could try the same here. But there are different approaches as well.

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -112,16 +161,40 @@ object Desugar {
...
+    def methodProxy(tag: BuiltInMethodTag, recv: in.Type)(src: Meta): in.MethodProxy = {
+      def genAndStore(tag: BuiltInMethodTag, recv: in.Type): in.MethodMember = {
+        val meth = generateBuiltInMethod(tag, recv)(src)
+        builtInMethods += (tag, recv) -> meth
+        meth
+      }
+
+      val meth = builtInMethods.getOrElse((tag, recv), genAndStore(tag, recv))
+      meth.name
+    }
```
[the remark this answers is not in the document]
ArquintL: Thanks for the inspiration. I was even able to simplify it further by using the same map for all members

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -2031,6 +2268,259 @@ object Desugar {
       for { exprs <- sequence(trigger.exps map exprD(ctx)) } yield in.Trigger(exprs)(src)
     }
 
+    def generateBuiltInFunction(tag: BuiltInFunctionTag, args: Vector[in.Type])(src: Meta): in.FunctionMember = {
+      val proxy = in.FunctionProxy(nm.builtInAuxType(tag, args))(src)
+
+      (tag, args) match {
+        case (CloseFunctionTag, Vector(channelT, dividendT, divisorT /* permissionAmountT */, predicateT)) =>
+          /**
+            * requires divisor > 0
```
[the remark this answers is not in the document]
ArquintL: I just tried it out. I had to add a new error reason but now Gobra no longer crashes and Viper correctly detects that the expression might be negative (in fact `dividend >= 0` is a bit strong as dividend and divisor both being negative would be fine)

src/main/scala/viper/gobra/frontend/info/base/Type.scala
```diff
@@ -138,4 +140,18 @@ object Type {
...
+    * The partial function `typing` maps from types of arguments to the return type.
+    * It should be defined for all argument types for which no error message was returned by `messages`
+    */
+  case class AuxType(messages: (PExpression, Vector[Type]) => Messages, typing: Vector[Type] ==> FunctionT) extends Type
+
+  /**
+    * Parameteric type for built-in mpredicates and methods which are only parametric in their receiver type
+    * Applying `typing` to the receiver type results in a typle of types for the remaining arguments and the return type
+    */
+  case class SingleAuxType(messages: (PExpression, Type) => Messages, typing: Type ==> FunctionT) extends Type
```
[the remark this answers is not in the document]
ArquintL: This is true. My idea was that `SingleAuxType` makes it explicit that there is only a dependency on the receiver type. If you use `AuxType` (or whatever other name we choose for it) you have to be careful to use the singleton vector containing the receiver type instead of the vector of argument types for methods and receivers.
However, I'm happy to unify this and just write this note as a comment. What do you prefer?

src/main/scala/viper/gobra/frontend/Desugar.scala
ArquintL: I moved these `generate...` functions from the Desugarer to `BuiltInMembersImpl` located in the encoding part

src/main/scala/viper/gobra/translator/encodings/channels/ChannelEncoding.scala
ArquintL: Issue #174 created

src/main/scala/viper/gobra/frontend/Desugar.scala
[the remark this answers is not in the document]
jcp19: Awesome! Thanks for checking :)

src/main/scala/viper/gobra/frontend/info/base/Type.scala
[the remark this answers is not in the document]
Felalolf: I would prefer to unify it, but you can add a constructor for a single argument.

src/main/scala/viper/gobra/translator/encodings/channels/ChannelEncoding.scala
ArquintL: I've now added the following inhale making the encoding slightly stronger:
`inhale res != Dflt[T] ==> [c].RecvGotPerm()(res)`
However as Martin pointed out, he would maybe even go as far as simply disallowing this form of receive

src/main/scala/viper/gobra/frontend/info/base/BuiltInMemberTag.scala
ArquintL: Great idea! If you go with `GhostBuiltInMemberTag` and it saves you from implementing `argGhostTyping` and `returnGhostTyping` since it can be derived (this corresponds to the previous "catch-all" case).

ArquintL:
@jcp19 @Felalolf I've implemented your suggestions or linked a corresponding issue for addressing it later. Thanks to @jcp19, a `BuiltInMemberTag` now directly contains its `typ`, `argGhostTyping`, and `returnGhostTyping` (instead of having separate getter functions in the object). Thanks to @Felalolf, there is a `BuiltInMembersImpl` in the encoding part that translates a built-in member to a 'normal' member which is then encoded using the encoding for 'normal' members. In addition, the postcondition for the single-value receive operation is slightly stronger now i.e. receiving a non-zero value implies that the channel invariant holds.
Sorry for resolving your comments but I used it to keep track which comments I've already processed and which not

ArquintL:
@MartinClochard The relevant files containing the encoding are as follows:
- `ChannelEncoding.scala`: encodes make, send, receive, and safe receive
- `BuiltInMembersImpl.scala`: encodes all remaining functions, methods, fpredicates, and mpredicates
In addition, I've summarized the encoding in the initial comment in this PR

src/main/scala/viper/gobra/translator/implementations/translator/BuiltInMembersImpl.scala
```diff
@@ -0,0 +1,454 @@
...
+import viper.gobra.util.Computation
+import viper.gobra.util.Violation.violation
+
+import scala.language.postfixOps
+
+
+/**
+  * Encodes built-in members by translating them to 'regular' members and calling the corresponding encoding
+  */
+class BuiltInMembersImpl extends BuiltInMembers {
```
Felalolf, line 27: Seeing everything that is in here makes me doubt that this right now is the right approach. That said, we should definitely merge it for now and then after the deadline has settled down and we have time to redesign something, then we can look at it again.

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -1070,8 +1147,15 @@ case class InterfaceT(name: String, addressability: Addressability) extends Type
     InterfaceT(name, newAddressability)
 }
 
+case class ChannelT(elem: Type, addressability: Addressability) extends Type {
```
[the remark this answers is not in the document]
ArquintL: Correct, the type checker ensures that you only access the channel in permitted ways but for the encoding we can forget about it. It even simplifies the encoding because there should be an equality between the `Closed` mpredicate for all directions of a particular channel (put differently, the `Closed` mpredicate should take the supertype of all channel directions as receiver) (this is what @MartinClochard calls "Only if c bidirectional , otherwise type error" in the concurrency outline). Martin then recommended me to unify all channel directions instead of adding a top type.

src/main/scala/viper/gobra/frontend/info/base/BuiltInMemberTag.scala
```diff
@@ -0,0 +1,305 @@
...
+        ghostArgs(typ(config).typing(args).args.length)
+      } else {
+        Violation.violation(s"argGhostTyping not defined for $name")
+      }
+
+    def returnGhostTyping(args: Vector[Type])(config: Config): GhostType =
+      if (typ(config).typing.isDefinedAt(args)) {
+        typ(config).typing(args).result match {
+          case VoidType => ghostArgs(0)
+          case _ => ghostArgs(1) // multi return parameters are not used yet by any built-in member
```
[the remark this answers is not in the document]
ArquintL: I was hoping for the same as well. However this does not seem to work for the following reason:
Assume you want to type a function/method call with 5 arguments and arg typing returns `GhostType.isGhost` i.e. a singleton vector containing `true`. `generalGhostAssignableTo` checks whether this assignment is possible but trying to assign 5 arguments (right) to a single assignee (left) results in an `AssignMode.Error`.
This issue does not seem to occur for user-defined functions and methods because `calleeArgGhostTyping` iterates over all parameters, calls `isParamGhost` and inserts a corresponding flag into GhostType for each parameter.

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1161,35 +1379,61 @@ object Desugar {
...
 
-          case PPredConstructor(base, args) => base match {
-            case PFPredBase(id) =>
-              val proxy = fpredicateProxyD(id)
-              val idT = info.typ(id) match {
-                case FunctionT(args, AssertionT) =>
-                  in.PredT(args.map(typeD(_, Addressability.rValue)(src)), Addressability.rValue)
+          case PPredConstructor(base, args) => (base, info.regular(base.id)) match {
+            case (PFPredBase(id), _: st.FPredicate) =>
+              val proxy = fpredicateProxy(id)
```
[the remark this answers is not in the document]
ArquintL: `info.typ(id)` returns different types as well but I tried to unify the code as good as possible

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -831,80 +967,145 @@ object Desugar {
...
+          // go function chaining feature
+          case Vector(in.Tuple(targs)) if parameterCount > 1 => targs
+          case dargs => dargs
+        }
+      }
 
-                  val dRecv = for {
-                    r <- exprD(ctx)(base.recv)
-                  } yield applyMemberPathD(r, base.path)(src)
+      def encodeResults(): (in.Type, Vector[in.LocalVar]) = p.callee match {
```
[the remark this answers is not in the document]
ArquintL: I've inlined the two functions

jcp19 (approved):
I did not catch any obvious mistake. At this point, I think we should merge it

src/main/scala/viper/gobra/frontend/info/base/BuiltInMemberTag.scala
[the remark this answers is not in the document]
Felalolf: Could you then change the comment to something like "As a default, we pick a single ghost result. Other cases have to overwrite this  function."

src/main/scala/viper/gobra/frontend/Desugar.scala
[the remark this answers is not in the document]
Felalolf: you could have just replaced def with a val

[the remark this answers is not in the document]
ArquintL: I know but inlining seemed nicer to me

Issue #169, opened 2021-01-18
NodeNotInTreeException in imported n-ary mpredicates (n > 0)


jcp19:
*Gobra* crashes with the following files
- `bla/imported.gobra`
```go
package bla

type T struct {}

pred (t *T)mPred(x int)
```
- `import_pred.gobra`
```go
package pkg

import "bla"

requires t.mPred(y)
pure func f(t* bla.T, y int) int
```
with the following stack trace
```
[error] Exception in thread "main" org.bitbucket.inkytonik.kiama.relation.NodeNotInTreeException: node not in tree: x int
[error] 	at org.bitbucket.inkytonik.kiama.relation.TreeRelation.apply(TreeRelation.scala:53)
[error] 	at viper.gobra.frontend.info.implementation.typing.BaseTyping.children(BaseTyping.scala:50)
[error] 	at viper.gobra.frontend.info.implementation.typing.BaseTyping.children$(BaseTyping.scala:49)
[error] 	at viper.gobra.frontend.info.implementation.TypeInfoImpl.children(TypeInfoImpl.scala:22)
[error] 	at viper.gobra.frontend.info.implementation.typing.BaseTyping.childrenWellDefined(BaseTyping.scala:52)
[error] 	at viper.gobra.frontend.info.implementation.typing.BaseTyping.childrenWellDefined$(BaseTyping.scala:52)
[error] 	at viper.gobra.frontend.info.implementation.TypeInfoImpl.childrenWellDefined(TypeInfoImpl.scala:22)
[error] 	at viper.gobra.frontend.info.implementation.typing.BaseTyping$$anon$1.safe(BaseTyping.scala:68)
[error] 	at viper.gobra.frontend.info.implementation.typing.BaseTyping$$anon$1.safe(BaseTyping.scala:66)
```
If `mPred` is declared with no parameters (besides the receiver), then Gobra runs as expected

jcp19:
Besides, no exception is thrown if mpred is changed to
```go 
pred mPred(t *T, x int)
```
(with the necessary changes in `import_pred.gobra`

ArquintL:
My guess: The type checker tries to check whether the type of `y` in `t.mPred(y)` conforms to the signature and to do so gets the parameter `x int`. Apparently, it uses the wrong context, i.e. the context of package `pkg` instead of `bla`, to perform some operation on the parameter (such as getting the type).

Reacting to your second finding: The correct context seems to be used for fpredicates but not for mpredicates

[...]

PR #170, opened 2021-01-18
Fix issue 169

jcp19:
Fixes issue #169.

Regarding the case for `MPredicateSpec` in the `GhostMiscTyping.scala`, I did not find an example where Gobra crashes without my change. Nevertheless, I decided to add it for consistency and to avoid any unforeseen bugs due to the use of a wrong context


ArquintL (approved):
Looks good. I wouldn‘t know either how to trigger the other case since `PMPredicateSig` seems to be on a dead path in the parser

Issue #177, opened 2021-01-19
Extremely restrictive triggers


MartinClochard:
Currently in gobra it is impossible to write a trigger that contains an arithmetic operation, a slice access, etc. In a lot of situations, the only meaningful trigger must use those. The complaint come from Gobra itself, not from the Viper backend.

Furthermore, when attempting to do so gobra also complains about the quantified variable not being present in the trigger, while it is (in the rejected trigger).

Example:
```
package pkg
pure func f(p int,q int) bool { return true }
requires forall i int :: 0 <= i && i < len(s) ==> acc(&s[i])
func test(p int,s[] int) {
  assert forall i int :: {f(p+1,s[i])} 0 <= i && i < len(s) ==> f(p+1,s[i])
}
```

PR #178, opened 2021-01-19
Workaround restricted triggers

jcp19:
Temporary workaround to issue #177, as discussed in today's Gobra meeting. The test introduced in this PR is the exact same one that Martin provided in the issue description.


Felalolf (approved):

Felalolf:
Add `//:: IgnoreFile(/gobra/issue/177/)` into the first(!) line of the 8 files that fail the test, currently.

ArquintL:
I was already concerned that the license header check will fail with that line but I'm glad they don't conflict with each other

PR #179, opened 2021-01-19
Add sync stubs

jcp19:
Adds stubs for `Mutex` and `WaitGroup` in the `sync` package and fix bugs that I found while implementing them.

PS: this is a draft PR. Before merging, I plan to add more simple tests for Mutexes and WaitGroups. Everything else will probably remain the same, modulo feedback that I get.


src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -492,7 +492,7 @@ sealed trait PPredConstructorBase extends PNode {
   val id: PIdnUse
 }
 case class PFPredBase(override val id: PIdnUse) extends PPredConstructorBase
-case class PMPredBase(recvWithId: PDot) extends PPredConstructorBase {
+case class PDottedBase(recvWithId: PDot) extends PPredConstructorBase {
```
Felalolf, line 495: I do not understand why this name has to change

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -431,9 +431,15 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
           val idT = idType(id)
           PredT(idT.asInstanceOf[FunctionT].args.zip(args).collect{ case (typ, None) => typ })
-        case p: PMPredBase =>
-          val idT = idType(p.id)
-          PredT(idT.asInstanceOf[FunctionT].args.tail.zip(args).collect{ case (typ, None) => typ })
+        case p: PDottedBase => resolve(p.recvWithId) match {
+          case Some(_: ap.Predicate) =>
+            val recvWithIdT = exprOrTypeType(p.recvWithId)
+            PredT(recvWithIdT.asInstanceOf[FunctionT].args.zip(args).collect{ case (typ, None) => typ })
+          case Some(_: ap.PredicateExpr) => ??? // Predicate expressions are not supported at the moment
```
Felalolf, line 438: Currently, we also do not plan to support them in the future. With my latest proposal, they do not make much sense.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -431,9 +431,15 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
+          case _ =>
+            val idT = idType(p.id)
+            PredT(idT.asInstanceOf[FunctionT].args.tail.zip(args).collect{ case (typ, None) => typ })
```
Felalolf, line 441: What case is this supposed to be? That just looks like the case where the base is a predicate. In my opinion, there should be a violation here stating that an unexpected case was encountered.

src/main/scala/viper/gobra/translator/encodings/preds/PredEncoding.scala
```diff
@@ -46,6 +46,7 @@ class PredEncoding extends LeafTypeEncoding {
     * (1) exclusive operations on T, which includes literals and default values
     *
     * [Q{d1, ..., dk}: pred(S)] -> make_S_ID([d1], ..., [dk]) where ID is the ID for the pattern used in Q{d1, ..., dk}
+    * [dflt(Pred(ts)°)] -> default() where default is a 0-ary function in the domain with name domainName(embedPredType(ts))
```
Felalolf, line 49: `domainName(embedPredType(ts))` is not a name. But I also do not think that you have to provide the name here.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
```diff
@@ -280,6 +281,7 @@ trait GhostExprTyping extends BaseTyping { this: TypeInfoImpl =>
       case n: PInvoke => (exprOrType(n.base), resolve(n)) match {
         case (Right(_), Some(_: ap.Conversion)) => false // Might change at some point
         case (Left(callee), Some(p: ap.FunctionCall)) => go(callee) && p.args.forall(go)
+        case (Left(_), Some(p: ap.PredExprInstance)) => !strong
         case _ => false
```
Felalolf, line 285: It seems to me that the case for predicate instances is also missing:
`case (Left(_), Some(p: ap.PredicateCall)) => !strong`

src/main/scala/viper/gobra/ast/frontend/Ast.scala
Felalolf: I understand now why the name has to change.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
Felalolf: I think the case that is missing is the received predicate, but this should also be added as a pattern match with an `AstPattern`. For this also the implementation is wrong, I think. You should not take the type of the id, but the type of the entire base `p.recvWithId`. Otherwise, you do not account for the receiver of the mpredicate.

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1165,32 +1165,38 @@ object Desugar {
...
+                    dRecv <- exprAndTypeAsExpr(ctx)(p.recv)
+                    dArgs <- sequence(args.map { x => option(x.map(exprD(ctx)(_))) })
+                  } yield in.PredicateConstructor(proxy, idT, Some(dRecv) +: dArgs)(src)
 
-              for {
-                dRecv <- exprAndTypeAsExpr(ctx)(p.recv)
-                dArgs <- sequence(args.map { x => option(x.map(exprD(ctx)(_))) })
-              } yield in.PredicateConstructor(proxy, idT, Some(dRecv) +: dArgs)(src)
-          }
+                case _ => handleFPredBase(p)
```
Felalolf, line 1197: This should be an explicit match on `ap.Predicate`

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1165,32 +1165,38 @@ object Desugar {
               _ <- write(in.New(target, zero)(src))
             } yield target
 
-          case PPredConstructor(base, args) => base match {
-            case PFPredBase(id) =>
-              val proxy = fpredicateProxyD(id)
-              val idT = info.typ(id) match {
+          case PPredConstructor(base, args) =>
+            def handleFPredBase(p: PPredConstructorBase): Writer[in.Expr] = {
+              val proxy = fpredicateProxyD(p.id)
```
Felalolf, line 1170: I am surprised that this works if the base is a predicate from another package.

Felalolf (commented):
I did not look at the actual lock and waitgroup stub.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
jcp19: I will change the type system to complain when a Predicate Expression is used in a predicate constructor

src/main/scala/viper/gobra/frontend/Desugar.scala
jcp19: Do you think I should change it to something else?

jcp19: One thing I also tried was to implement `handleFPredBase` like the following but that triggers a `NodeNotInTreeException` for predicates from other packages

```scala
            def handleFPredBase(p: PPredConstructorBase): Writer[in.Expr] = p match {
              case PFPredBase(id) =>
                val proxy = fpredicateProxyD(id)
                val predConstructType = info.typ(id)
                val idT = predConstructType match {
                  case FunctionT(args, AssertionT) =>
                    in.PredT(args.map(typeD(_, Addressability.rValue)(src)), Addressability.rValue)
                  case t => violation(s"desugarer: $t cannot be converted to a predicate type")
                }
                for {
                  dArgs <- sequence(args.map { x => option(x.map(exprD(ctx)(_))) })
                } yield in.PredicateConstructor(proxy, idT, dArgs)(src)

              case PDottedBase(recvWithId) =>
                info.resolve(recvWithId) match {
                  case Some(ap.Predicate(id, symb)) =>
                    val proxy = fpredicateProxyD(symb.decl)
                    val predConstructType = info.typ(recvWithId)
                    // TODO: abstract this part
                    val idT = predConstructType match {
                      case FunctionT(args, AssertionT) =>
                        in.PredT(args.map(typeD(_, Addressability.rValue)(src)), Addressability.rValue)
                      case t => violation(s"desugarer: $t cannot be converted to a predicate type")
                    }
                    for {
                      dArgs <- sequence(args.map { x => option(x.map(exprD(ctx)(_))) })
                    } yield in.PredicateConstructor(proxy, idT, dArgs)(src)
                }
            }
```

jcp19:
The example `mutex-and-waitgroup1.gobra` and other smaller tests are still a work in progress but I figured that the rest can already be reviewed, specially the stubs for the mutexes and waitgroups.

@ArquintL and @MartinClochard, the files `src/main/resources/stubs/sync/mutex.gobra` and `src/main/resources/stubs/sync/waitgroup.gobra` are the most intersting parts of this PR. They closely follow Martin's proposal, with the modifications 
that we discussed in our various meetings.

@ArquintL some things changed in this PR regarding the treatment of MPredicates when used as bases of Predicate Constructors. In particular, I renamed the class `PMPredBase` to `PDottedBase` because it can not only represent an mpredicate but also to a regular predicate qualified by its package. As such, I had to extend the type checker and desugarer with case splits for these cases. This has led to some conflicts with the code you introduced in #168 to deal with `AbstractType` in `Desugarer.scala` and `ExprTyping.scala`. I believe I correctly fixed the conflicts but it would be great if you could a second look at that.

src/main/resources/stubs/sync/waitgroup.gobra
```diff
@@ -0,0 +1,123 @@
...
+// This is the only rule that generates a Token.
+ghost
+requires R() && g.UnitDebt(pred_true!<!>)
+ensures g.UnitDebt(R) && g.Token(R)
+func (g *WaitGroup) GenerateToken(ghost R pred())
+
+requires dividend >= 0 && divisor > 0 // p >= 0
+requires n > 0 && dividend > 0 ==> acc(g.WaitGroupP(), dividend/divisor) && !g.WaitMode()
+requires (n > 0 && dividend == 0) ==> g.UnitDebt(P)
+requires n < 0 ==> acc(g.UnitDebt(pred_true!<!>), -n/1)
```
ArquintL, line 74: Why is `pred_true` used here instead of `P` like in the document? Wouldn't this require some precondition that `P == PredTrue` or doesn't this matter since parameter P is never used if n is negative?

src/main/resources/stubs/sync/waitgroup.gobra
```diff
@@ -0,0 +1,123 @@
...
+func (g *WaitGroup) Add(n int, ghost dividend int, ghost divisor int, ghost P pred())
+
+requires g.UnitDebt(pred_true!<!>)
+func (g *WaitGroup) Done()
+
+// `dividend > 0 && divisor > 0` implies that the permission p in Martin's proposal is greater than 0
+requires dividend > 0 && divisor > 0 // p > 0
+requires acc(g.WaitGroupP(), dividend/divisor)
+requires g.WaitMode()
+requires acc(g.Token(P))
```
ArquintL, line 89: please drop the `acc` to have a consistent use of predicate with full permission

src/main/resources/stubs/sync/waitgroup.gobra
ArquintL: Nevermind, I've just seen that this contract corresponds to the contract of `Add` in Remark 8.  It might help to add a smaller comment to this contract as well as the contract for `PayDebt` that they follow Remark 8

src/main/resources/stubs/sync/waitgroup.gobra
```diff
@@ -0,0 +1,123 @@
...
+ensures acc(g.WaitGroupP(), dividend/divisor)
+func (g *WaitGroup) Wait(ghost dividend int, ghost divisor int, ghost P pred())
+
+pred CollectFractions(ghost P seq[pred()], ghost dividends seq[int], ghost divisors seq[int]) {
+	len(P) == len(dividends) && len(dividends) == len(divisors) &&
+	forall i int :: 0 <= i && i < len(P) ==> dividends[i] >= 0 && divisors[i] > 0 && acc(P[i](), dividends[i]/divisors[i])
+}
+
+// Problem: CollectFractions(P, [x1, ..., xn], [y1, ..., yn]) does not imply CollectFractions(P, [x1', ..., xn'], [y1', ..., yn']) even if for all i, xi / yi == xi' / yi'. This will not be a problem once we add
+// the Perm type to Gobra. Meanwhile, a new viewshift is required for this case.
```
ArquintL, line 100: What is the status of this viewshift? Does it already exist?

src/main/resources/stubs/sync/waitgroup.gobra
```diff
@@ -0,0 +1,123 @@
...
+requires g.UnitDebt(CollectFractions!<P ++ Q, dividendsP ++ dividendsQ, divisorsP ++ divisorsQ!>)
+requires g.UnitDebt(pred_true!<!>)
+ensures g.UnitDebt(CollectFractions!<P, dividendsP, divisorsP!>) && g.UnitDebt(CollectFractions!<Q, dividendsQ, divisorsQ!>)
+func (g *WaitGroup) SplitSequence(ghost P seq[pred()], ghost Q seq[pred()], ghost dividendsP seq[int], ghost divisorsP seq[int], ghost dividendsQ seq[int], ghost divisorsQ seq[int])
+
+// Special case of the debt redistribution rule
+ghost
+requires len(P) == len(dividendsP) && len(P) == len(divisorsP) // |P| == |p|
+requires len(P) == len(dividendsQ) && len(P) == len(divisorsQ) // |Q| == |q|
+requires len(P) == len(dividendsR) && len(P) == len(divisorsR) // |R| == |r|
```
ArquintL, line 115:
```suggestion
requires len(P) == len(dividendsQ) && len(P) == len(divisorsQ) // |p| == |q|
requires len(P) == len(dividendsR) && len(P) == len(divisorsR) // |q| == |r|
```

src/main/scala/viper/gobra/translator/encodings/preds/DefuncComponentImpl.scala
```diff
@@ -136,6 +136,28 @@ class DefuncComponentImpl extends DefuncComponent {
     encounteredTokens foreach (S => genEval(S) foreach col.addMember)
   }
 
+  /** Returns the default value of the predicate type with arguments 'ts' */
+  override def default(ts: Vector[in.Type])(pos: vpr.Position = vpr.NoPosition, info: vpr.Info = vpr.NoInfo, errT: vpr.ErrorTrafo = vpr.NoTrafos)(ctx: Context): vpr.Exp = {
+    vpr.DomainFuncApp(func = genDefaultFunc(ts)(ctx), Seq.empty, Map.empty)(pos, info, errT)
+  }
+
+  /** Creates a domain function that returns a value of the predicate type with the 'ts' parameter list */
```
ArquintL, line 144:
```suggestion
  /** Creates a domain function that returns the default value of the predicate type with the 'ts' parameter list */
```

src/main/scala/viper/gobra/translator/encodings/preds/PredEncoding.scala
```diff
@@ -46,6 +46,7 @@ class PredEncoding extends LeafTypeEncoding {
     * (1) exclusive operations on T, which includes literals and default values
     *
     * [Q{d1, ..., dk}: pred(S)] -> make_S_ID([d1], ..., [dk]) where ID is the ID for the pattern used in Q{d1, ..., dk}
+    * [dflt(Pred(ts)°)] -> default() where default is a 0-ary function in the domain
```
ArquintL, line 49: I have some troubles imagining where a default value for a `Pred` type shows up. Can you point me at an instance in one of the test cases?
Is it correct that you cannot do much with the default value since e.g. fold/unfold does not work thanks to our restriction that you have to write a pred constructor (instead of just a predicate expression)?

src/test/resources/regressions/features/defunc/mutex-and-waitgroup1.gobra
```diff
@@ -0,0 +1,93 @@
...
+	//    w.WaitGroupP && !w.WaitMode()
+	// }
+
+	n := len(s)
+
+	invariant i >= 0
+	invariant i == 0 ==> w.WaitGroupP()
+	invariant i > 0 ==> acc(w.WaitGroupP(), 1/2) && acc(w.WaitGroupStarted(), 1/2)
+	// Martin suggested the following line instead of the previous one, with the goal of
+	// deleting the invariant `i >= 0`. However, that is insufficient to prove the loop invariant
```
ArquintL, line 52: I came across the "stupid" invariants when working on examples as well. In my opinion the encoding of `for ... { ... }` should be made stronger such that invariants like `i >= 0` and `i <= n` are automatically inserted into the encoding

src/test/resources/regressions/features/defunc/pred-construct-fail1.gobra
```diff
@@ -0,0 +1,31 @@
...
+	acc(rec)
+}
+
+func f(p pred())
+
+func error1() {
+	v := &T{}
+	fold v.AccInv()
+	//:: ExpectedOutput(type_error)
+	f((*T).AccInv!<v!>) // Predicate expressions are not allowed in predicate constructors
```
ArquintL, line 20: Isn't this the same as `f(v.AccInv!<!>)`? And isn't this allowed?

src/test/resources/regressions/features/defunc/waitgroup-simple1.gobra
```diff
@@ -0,0 +1,104 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package pkg
+
+import "sync"
+
+// TODO: remove all occurrences of sync.pred_true when PredTrue is added to the language
```
ArquintL, line 8: fix this TODO as well

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1407,26 +1407,41 @@ object Desugar {
...
+                  val argsT = argsTyp.map(typeD(_, Addressability.rValue)(src))
+                  val proxy = getMPredProxy(recvT, argsT)
+                  val idT = in.PredT(recvT +: argsT, Addressability.rValue)
+                  for {
+                    dRecv <- exprAndTypeAsExpr(ctx)(p.recv)
+                    dArgs <- sequence(args.map { x => option(x.map(exprD(ctx)(_))) })
+                  } yield in.PredicateConstructor(proxy, idT, Some(dRecv) +: dArgs)(src)
+
+                case Some(_: ap.Predicate) =>
+                  val idT = info.typ(p.recvWithId) match {
```
ArquintL, line 1433: looks good. But maybe you could get rid of the code duplication by having a function `constructFPredConstructor` that takes `id` resp. `p.recvWithId` as argument

ArquintL (approved):
I have left some comments but in general it looks good to me

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
ArquintL: @Felalolf Why do `PredicateExpr` not make much sense with your latest proposal?
For me, `PredicateExpr` and `ReceivedPredicate` are two different forms for the equivalent thing (like Go does it for methods)

src/test/resources/regressions/features/defunc/pred-construct-fail1.gobra
jcp19: @Felalolf told me that Predicate Expressions in the base of predicate constructors will not make sense after he implements his latest proposal for interfaces.I don't recall the exact reason for this but maybe Felix can shed a light on this.

src/main/resources/stubs/sync/waitgroup.gobra
jcp19: I did not implement it as it will become completely unnecessary after we introduce the Perm Type and no example that I came accross thus far needed this. However, if you find it useful, I will write it :) stay tuned for the next commits

ArquintL: Makes sense. I was just not sure what ‚a new viewshift is required‘ exactly means

src/main/resources/stubs/sync/waitgroup.gobra
jcp19: I opted not to make references to particular points of Martin's draft because it is not publicly available and, in the future, the structure of the draft may change, invalidating our comments

src/test/resources/regressions/features/defunc/mutex-and-waitgroup1.gobra
jcp19: Yes, that is definitely something that would be helpful!

src/main/scala/viper/gobra/translator/encodings/preds/PredEncoding.scala
jcp19: I believe the default value is pretty much useless per se but it is required for cases where you create a collection of predicates. For example, `x := [5]pred(){}`. Without this, it would crash, as the array encoding first initializes the array with the default value

src/test/resources/regressions/features/defunc/mutex-and-waitgroup1.gobra
ArquintL: Felix has convinced me today that it's not as easy as I thought and would probably be quite heuristic based. E.g. inserting the mentioned invariants would only be correct if `i` is not changed in the loop body and/or captured in e.g. a closure.

src/main/resources/stubs/sync/waitgroup.gobra
jcp19: I added the view shift in the last commit. It is totally a bit of an ugly thing. I think that, if time allows, we should focus on adding the Perm type to the language

jcp19:
This is shaping up to be a big PR. Given that the rest of the team may need the changes from this PR, I would suggest merging this after some potential new feedback. The example `mutex-and-waitgroup1.gobra` is not finished but I would argue that it is better to continue this on a separate branch.

I made some changes again to the type checker to fix some bugs, it would be great if you could look into this again @ArquintL

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1381,6 +1381,19 @@ object Desugar {
             } yield target
 
           case PPredConstructor(base, args) =>
+            def handleFPredicate(base: PPredConstructorBase) = {
```
ArquintL, line 1384: nice!

ArquintL (approved):
I didn't look at the examples/test cases in detail but all other changes look good

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
Felalolf: I mixed up predicate method expressions and values with a predicate type. Former is fine, later is not.

ArquintL: Makes sense

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -363,48 +363,50 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
-                }
+              // the typing function should be defined for these arguments as `msgs` is empty
+              abstractT.typing(givenArgTypes) match {
+                case FunctionT(args, AssertionT) =>
+                  if (givenArgs.isEmpty && args.isEmpty) {
+                    noMessages
+                  } else {
+                    multiAssignableTo.errors(givenArgs map exprType, args)(p) ++
+                      p.args.flatMap(x => x.map(isExpr(_).out).getOrElse(noMessages))
+                  }
```
Felalolf, line 399: I do not understand this part. AbstractT should be the type of a builtin predicate. As such, I would expect that if it is fully applied that then the final type is AssertionT. Long story short, I expected `case AssertionT => noMessages`

Felalolf (commented):
I have some more comments.

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1394,39 +1407,32 @@ object Desugar {
...
+                      }
+                    case t => violation(s"desugarer: $t cannot be converted to a predicate type")
+                  }
+                  val argsT = argsTyp.map(typeD(_, Addressability.rValue)(src))
+                  val proxy = getMPredProxy(recvT, argsT)
+                  val idT = in.PredT(recvT +: argsT, Addressability.rValue)
+                  for {
+                    dRecv <- exprAndTypeAsExpr(ctx)(p.recv)
+                    dArgs <- sequence(args.map { x => option(x.map(exprD(ctx)(_))) })
+                  } yield in.PredicateConstructor(proxy, idT, Some(dRecv) +: dArgs)(src)
```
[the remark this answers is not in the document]
Felalolf: My check whether an abstractT is fully applied is wrong.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
Felalolf: I discussed it with linard. As it is written, this should be correct.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostMiscTyping.scala
```diff
@@ -6,26 +6,40 @@
...
 
 trait GhostMiscTyping extends BaseTyping { this: TypeInfoImpl =>
 
   private[typing] def wellDefGhostMisc(misc: PGhostMisc) = misc match {
     case PBoundVariable(_, _) => noMessages
     case PTrigger(exprs) => exprs.flatMap(isPureExpr)
     case PExplicitGhostParameter(_) => noMessages
+    case p: PPredConstructorBase => p match {
+      case PFPredBase(_) => noMessages
+      case base: PDottedBase => wellDefExpr(base.recvWithId).out ++ (resolve(base.recvWithId) match {
```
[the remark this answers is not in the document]
jcp19: I believe you are correct

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -363,48 +363,50 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
-            val givenArgs = p.args.flatten
-            if (givenArgs.length != p.args.length) {
-              error(p, s"partial application is not supported for built-in predicates")
+    case p@PPredConstructor(base, _) => {
+      def wellTypedApp(base: PPredConstructorBase): Messages = miscType(base) match {
+        case FunctionT(args, AssertionT) =>
+          val unappliedPositions = p.args.zipWithIndex.filter(_._1.isEmpty).map(_._2)
+          val givenArgs = p.args.zipWithIndex.filterNot(x => unappliedPositions.contains(x._2)).map(_._1.get)
+          val expectedArgs = args.zipWithIndex.filterNot(x => unappliedPositions.contains(x._2)).map(_._1)
+          if (givenArgs.isEmpty && expectedArgs.isEmpty) {
```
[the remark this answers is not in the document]
jcp19: `multiAssignableTo` checks, using the following method, that the sizes of the vectors are not 0 in order to detect the AssignMode
```scala
object StrictAssignModi {
  def apply(left: Int, right: Int): AssignMode =
    if (left > 0 && left == right) AssignMode.Single
    else if (left > right && right == 1) AssignMode.Multi
    else AssignMode.Error
}
```
As such, the `if` is required unless I change the code for `StrictAssignModi`. (I think that the change would actually be welcome and easy to do: just replace `left > 0` with `left >= 0` in the condition of the first if)

src/main/scala/viper/gobra/frontend/Desugar.scala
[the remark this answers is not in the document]
ArquintL: I like the idea of calling `info.typ(base)` which should always return a FunctionT for mpredicate and methods (as resolving the AbstractType should happen while typing the base.
For checking that the arguments are fully applied (for functions and fpredicates, i.e. when getting back an AbstractType) I would check whether `dArgs` and `dArgs.flatten` are of equal length

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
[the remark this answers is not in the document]
Felalolf: No, then it is fine. Keep it as it is.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostMiscTyping.scala
```diff
@@ -6,26 +6,40 @@
...
+        case Some(_: ap.Predicate | _: ap.ReceivedPredicate) => noMessages
+        case Some(_: ap.PredicateExpr) =>
+          error(base, s"predicate expressions are not valid bases of predicate constructors")
```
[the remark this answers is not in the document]
jcp19: support for PredicateMethodExpressions added in the last commit

Felalolf (approved):

PR #181, opened 2021-01-19
Fixes encoding of result variable in postconditions of pure methods

### LINKED ISSUE #180: Wrong Postcondition Encoding for Pure Method
The following program results in a Viper file in which the postcondition of `test1` and `test3` do not refer to `result` but an undeclared variable:
```
package pkg

type Tree struct {}

ghost
ensures forall i int :: (0 <= i && i + 1 < len(res) ==> res[i] < res[i + 1])
pure func (t *Tree) test1() (res seq[int]) {
    return seq[int] { }
}

ghost
ensures forall i int :: (0 <= i && i + 1 < len(res) ==> res[i] < res[i + 1])
pure func test2() (res seq[int]) {
    return seq[int] { }
}

ensures forall i int :: (0 <= i && i + 1 < len(res) ==> res[i] < res[i + 1])
pure func (t *Tree) test3() (ghost res seq[int]) {
    return seq[int] { }
}

ensures forall i int :: (0 <= i && i + 1 < len(res) ==> res[i] < res[i + 1])
pure func test4() (ghost res seq[int]) {
    return seq[int] { }
}
```

ArquintL:
Fixes #180

### DIFF
--- a/src/main/scala/viper/gobra/translator/implementations/translator/PureMethodsImpl.scala
+++ b/src/main/scala/viper/gobra/translator/implementations/translator/PureMethodsImpl.scala
@@ -38,7 +38,7 @@ class PureMethodsImpl extends PureMethods {
     val resultType = if (vResults.size == 1) vResults.head.typ else ctx.tuple.typ(vResults map (_.typ))
 
     val fixResultvar = (x: vpr.Exp) => {
-      x.transform { case v: vpr.LocalVar if v.name == meth.results.head => vpr.Result(resultType)() }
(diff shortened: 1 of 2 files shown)


Felalolf (approved):
I will deduce the student who implemented this a point for that (this is a joke).
EDIT: I might have been the criminal. I saw a different commit where I also made the mistake.

ArquintL:
I was surprised that the compiler did not point out this fruitless comparison. Interestingly, IntelliJ directly hinted at this when manually evaluating the expression

PR #186, opened 2021-01-21
Correct type error reporting for short variable declarations

### LINKED ISSUE #185: Type Error Not Detected
The variable declaration in the following program is invalid but Gobra reports no type errors (but crashes in the desugarer):
```
package pkg

func test() (res1 bool, res2 int)

func client() {
    b := test()
}
```

ArquintL:
Fixes type checker to not silently skip well-definedness check for declarations if identifier definition is invalid

Fixes #185


Felalolf (approved):
brilliant solution

PR #188, opened 2021-01-22
Fix to issue 184

jcp19:
This PR was proposed by @ArquintL. He identified the root cause of the problem and showed me the solution. 

The bug was triggered because every call to `PackageResolver.resolve` would reuse the same stub resources. In the case that the resource was of type `BaseJarResource`, in the second call to `resolve`, the underlying file system would already be closed (due to the reference counting mechanism of `BaseJarResource`) and Gobra would crash in the first call to `retain`, with the message `managed file system can no longer be retained when it is already closed`.

I did not write any test cases for this bug because it was not caught in our testing framework in the first place.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/PackageResolver.scala
+++ b/src/main/scala/viper/gobra/frontend/PackageResolver.scala
@@ -71,7 +71,11 @@ object PackageResolver {
     includeDirs.map(FileResource)
   }
 
-  private lazy val getStubResources: Vector[InputResource] = {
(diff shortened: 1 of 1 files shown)


ArquintL:
Adding to @jcp19's description: It's actually not easily possible to write a test case as tests executed with `sbt test` access resources as regular files and thus do not trigger this part of Gobra that deals with resources located in the JAR file

ArquintL (approved):
Love it! 2 words removed; 1 word added

PR #193, opened 2021-01-26
ScalaTest to gather statistical data

ArquintL:
Adds `BenchmarkTests` to run Gobra on a warmup and target folder with test programs for which statistical data should be gathered.
As this builds upon `StatisticalTestSuite` a `Frontend` has to be provided. A frontend wrapper for Gobra is currently provided as part of `BenchmarkTests` but it could be moved to `Gobra.scala` for potential reusability. In addition, the frontend wrapper does not use `Gobra.verify` but directly calls parser, type-checker, etc. with the benefit that statistical data for each step is gathered.

Note the following design decisions and their potential effect on the gathered data:
- `gobraFrontend` is only created once and then reused. Thus, the execution context is reused as well.
- No additional parameters are passed to the backends. In particular, Silicon's default values are used for seeding Z3 (by default Z3 is not seeded randomly), catching exceptions (by default catching exceptions is turned on), and the number of threads used by Silicon (by default `availableProcessors() + 1` is used as the number of provers).


Felalolf (approved):

ArquintL:
I've now split it into two different Tests: `OverallBenchmarkTests` simply executes `Gobra.verify(...)` and `DetailedBenchmarkTests` individually executes Gobra's steps to report performance metrics per step.
A quick run has shown that `OverallBenchmarkTests` suffers from high standard deviations (in the order of 100%), maybe because of the additional use of `Future` and potential scheduling issues

src/test/scala/viper/gobra/OverallBenchmarkTests.scala
```diff
@@ -0,0 +1,102 @@
...
+  *    that can be used to run tests with SBT without the need to distribute/ recompile
+  *    the Gobra sources. To run the test without recompiling the sources, these
+  *    JAR files should be put into a directory test-location/lib/
+  *    where test-location/ is the directory where you invoke SBT via:
+  *    ```
+  *    sbt "set trapExit := false" \
+  *    "test:runMain org.scalatest.tools.Runner -o -s viper.gobra.OverallBenchmarkTests"
+  *    ```
+  *    Note that this command takes the same JVM property arguments as used above.
+  *    Linard (26.1.2020): I was not able to execute the tests using the generated JAR. This might have changed since Scala 2.13.
```
jcp19, line 56: Does this have any practical consequences? Btw, shouldn't the year be 2021?

jcp19 (approved):
I did not look at it in detail but it seems good to me

src/test/scala/viper/gobra/OverallBenchmarkTests.scala
ArquintL: Yes it should be 2021. I don't think this has any practical implications as you can simply check out Gobra and compile it from sources. What might be interesting though is to execute the benchmarks outside of sbt but this is not mentioned in the description nor do I know whether it is possible at all.

ArquintL: I just wanted to add this comment to the description before someone else desperately tries to follow the instructions thinking that it worked once...

ArquintL:
> A quick run has shown that OverallBenchmarkTests suffers from high standard deviations (in the order of 100%), maybe because of the additional use of Future and potential scheduling issues

I have rerun both test suites on a freshly booted mac and also tried to avoid any interactions with the running computer. Both test suites produce comparable results - in terms of mean and relative standard deviation

PR #194, opened 2021-01-26
Mutex and waitgroup example

jcp19:
@MartinClochard and I finished the proof of functional correctness of `sum`. This required some changes in the WaitGroup stub to avoid injectivity-related issues in Silver (by introducing `InjEval` and `TokenById` predicates). Besides, I fixed one of the view-shifts (previously called `GenerateToken`, now called `GenerateTokenAndDebt`) to remove one unnecessary pre-condition and introduced a new viewshift (`GenerateToken`) which allows the user to generate `Tokens` without generating the corresponding Debt (this is also a particular case of the debt redistribution rule).

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
@@ -94,7 +94,7 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
       multiAssignableTo.errors(Vector(miscType(exp)), lefts map exprType)(n) ++
         lefts.flatMap(t => addressable.errors(t)(t))
 
-    case n@PGoStmt(exp) => wellDefAndExpr(exp).out ++ isExecutable.errors(exp)(n)
+    case n@PGoStmt(exp) => isExpr(exp).out ++ isExecutable.errors(exp)(n)
 
     case n: PSelectStmt =>
       n.aRec.flatMap(rec =>
--- a/src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
@@ -94,7 +94,7 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
       multiAssignableTo.errors(Vector(miscType(exp)), lefts map exprType)(n) ++
         lefts.flatMap(t => addressable.errors(t)(t))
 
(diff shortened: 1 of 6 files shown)


jcp19:
PS: In my machine, the example takes almost 1 minute to verify. I am not sure if running it as part of a test suit is a good idea

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
```diff
@@ -94,7 +94,7 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
       multiAssignableTo.errors(Vector(miscType(exp)), lefts map exprType)(n) ++
         lefts.flatMap(t => addressable.errors(t)(t))
 
-    case n@PGoStmt(exp) => isExpr(exp).out ++ isExecutable.errors(exp)(n)
+    case n@PGoStmt(exp) => wellDefAndExpr(exp).out ++ isExecutable.errors(exp)(n)
```
Felalolf, line 97: this change should not be necessary.

Felalolf (approved):

src/test/resources/regressions/features/defunc/mutex-and-waitgroup1.gobra
```diff
@@ -10,84 +10,142 @@ type Accum struct {
...
-requires c.sum != nil
-ensures acc(c) && c.m.LockP() && c.m.LockInv() == (AccInv!<c.sum!>)
-func (c *Accum) Init() {
-	fold AccInv!<c.sum!>()
-	c.m.SetInv(AccInv!<c.sum!>)
+requires *c.sum == 0
+requires forall i int :: 0 <= i && i < len(seenSlice) ==> acc(&seenSlice[i], 1/2)
+requires forall i int :: 0 <= i && i < len(seenSlice) ==> seenSlice[i] == 0
+ensures acc(c) && c.m.LockP() && c.m.LockInv() == (AccInv!<c.sum, seenSlice!>)
+func (c *Accum) Init(ghost seenSlice []int) {
```
ArquintL, line 24: Would it be possible to make this function ghost? As it only performs ghost operations, ghost erasure would produce more meaningful LOCs if this is not erased to an empty function

ArquintL (approved):
Looks good to me. I only have a very minor comment and I'm wondering about the change that Felix has already pointed out

src/test/resources/regressions/features/defunc/mutex-and-waitgroup1.gobra
jcp19: Yes, you are right! There's no actual code here, only ghost code

PR #198, opened 2021-01-28
Channel Initialization Improvement

ArquintL:
@MartinClochard proposed to remove redundant parameters from the `Init` method for channels.
In particular, the new signature is:
```
requires c.IsChannel()
requires c.BufferSize() > 0 ==> (B == pred_true{})
ensures c.SendChannel() && c.RecvChannel()
ensures c.SendGivenPerm() == A && c.SendGotPerm() == B
ensures c.RecvGivenPerm() == B && c.RecvGotPerm() == A
ghost func (c chan T).Init(A pred(T), B pred())
```

As a consequence, `IsChannel` no longer takes the buffer size as a parameter and `BufferSize` is a newly added ghost pure method:
```
requires acc(c.IsChannel(), _)
pure func (c chan T).BufferSize() (k Int)
```

@jcp19 After adapting the type-checker, I've discovered the following bug that is related to type inference for int literals:
If you have `c.Init(0, ...)` in your code and the type-checker expects the new signature, i.e. the first argument should be of type `Pred(T)`, then type inference would return `Pred(T)` as `typCtx` but this then causes a crash in `assignableWithinBounds` as that one expects a type with underlying type int.
I attempted to fix it but please double check whether you are happy with my changes (they are all in `ExprTyping.scala`)

### DIFF
--- a/src/main/scala/viper/gobra/ast/internal/Program.scala
+++ b/src/main/scala/viper/gobra/ast/internal/Program.scala
@@ -234,7 +234,7 @@ sealed trait MakeStmt extends Stmt {
 }
 
 case class MakeSlice(override val target: LocalVar, override val typeParam: SliceT, lenArg: Expr, capArg: Option[Expr])(val info: Source.Parser.Info) extends MakeStmt
-case class MakeChannel(override val target: LocalVar, override val typeParam: ChannelT, bufferSizeArg: Option[Expr], isChannel: MPredicateProxy)(val info: Source.Parser.Info) extends MakeStmt
+case class MakeChannel(override val target: LocalVar, override val typeParam: ChannelT, bufferSizeArg: Option[Expr], isChannel: MPredicateProxy, bufferSize: MethodProxy)(val info: Source.Parser.Info) extends MakeStmt
 // TODO: change type of typeParam to MapT when MapT is implemented
 case class MakeMap(override val target: LocalVar, override val typeParam: Type, initialSpaceArg: Option[Expr])(val info: Source.Parser.Info) extends MakeStmt
 
--- a/src/main/scala/viper/gobra/frontend/info/base/BuiltInMemberTag.scala
+++ b/src/main/scala/viper/gobra/frontend/info/base/BuiltInMemberTag.scala
@@ -123,6 +123,15 @@ object BuiltInMemberTag {
 
 
   /** Built-in Method Tags */
+  case object BufferSizeMethodTag extends BuiltInMethodTag with GhostBuiltInMember {
+    override def identifier: String = "BufferSize"
+    override def name: String = "BufferSizeMethodTag"
+    override def isPure: Boolean = true
+
+    override def typ(config: Config): AbstractType =
+      channelReceiverType(allDirections, _ => FunctionT(Vector(), IntT(config.typeBounds.Int)))
+  }
+
   sealed trait ChannelInvariantMethodTag extends BuiltInMethodTag with GhostBuiltInMember
   sealed trait SendPermMethodTag extends ChannelInvariantMethodTag
   case object SendGivenPermMethodTag extends SendPermMethodTag {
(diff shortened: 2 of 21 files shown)


src/test/resources/regressions/features/channels/channel-fail6.gobra
```diff
@@ -8,11 +8,24 @@ pred sendInvariant(v int) {
 }
 
 // contracts for built-in members might use other built-in members which also need to be generated (even when not mentioned in the program text)
-requires c.IsChannel(1)
+requires c.IsChannel() && c.BufferSize() == 1
```
jcp19, line 11: I do not see a problem in using `BufferSize` for now but in the long term, we should probably use `cap` instead(https://golang.org/pkg/builtin/#cap)

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -234,7 +234,7 @@ sealed trait MakeStmt extends Stmt {
 }
 
 case class MakeSlice(override val target: LocalVar, override val typeParam: SliceT, lenArg: Expr, capArg: Option[Expr])(val info: Source.Parser.Info) extends MakeStmt
-case class MakeChannel(override val target: LocalVar, override val typeParam: ChannelT, bufferSizeArg: Option[Expr], isChannel: MPredicateProxy)(val info: Source.Parser.Info) extends MakeStmt
+case class MakeChannel(override val target: LocalVar, override val typeParam: ChannelT, bufferSizeArg: Option[Expr], isChannel: MPredicateProxy, bufferSize: MethodProxy)(val info: Source.Parser.Info) extends MakeStmt
```
jcp19, line 237: Why do we need to pass the proxy for `BufferSize` here? Won't this field be the same to every instance of `MakeChannel`?

src/main/scala/viper/gobra/frontend/info/base/BuiltInMemberTag.scala
```diff
@@ -168,12 +177,15 @@ object BuiltInMemberTag {
...
     override def typ(config: Config): AbstractType = channelReceiverType(allDirections, c => {
-      val bufferSizeArgType = IntT(config.typeBounds.Int)
-      val sendGivenPermArgType = PredT(Vector(c.elem))
-      val sendGotPermArgType = PredT(Vector()) // pred() because we enforce that sendGotPermArgType == recvGivenPermArgType
-      val recvGivenPermArgType = PredT(Vector())
-      val recvGotPermArgType = PredT(Vector(c.elem))
-      FunctionT(Vector(bufferSizeArgType, sendGivenPermArgType, sendGotPermArgType, recvGivenPermArgType, recvGotPermArgType), VoidType)
+      // init's signature is adapted to the heavy simplifications that are in place for the initial support for channels.
+      // in particular, the permission for SendGivenPerm and RecvGotPerm as well as SendGotPerm and RecvGivenPerm have
+      // to be equal. Thus, they get merge two parameters: The former pair is called `proPerm` as they describe the
```
jcp19, line 182: minor thing: "get merge" in line 182 sounds odd. Besides, I would rename `proPerm` to `sendPerm` as the permissions flow in the direction of the send operation and I would rename "contraPerm" to "recvPerm"

jcp19 (approved):

jcp19:
Regarding the type system bug, indeed, there was an implicit invariant that `typeCtx` must contain an IntT or else, be None. Thank you for making that explicit!

src/test/resources/regressions/features/channels/channel-fail6.gobra
ArquintL: I've created #202 to not forget about this

src/main/scala/viper/gobra/ast/internal/Program.scala
ArquintL: It is the same for all `MakeChannel` having the same channel element type. The reason to attach the proxy to the internal node comes from our current architecture for built-in members: the desugarer notices that the buffer size method is used, creates a unique name for it, attaches the corresponding proxy to the `MakeChannel` node, and keeps track of all built-in members used so far. This will then trigger the encoding to create a corresponding encoding for the buffer size method. The advantage of this architecture is that the built-in member's call site (here the `make` call) does not need to differentiate between calling a built-in and a non-built-in member

src/main/scala/viper/gobra/frontend/info/base/BuiltInMemberTag.scala
ArquintL: I was thinking of calling it `sendPerm` but then opted for a completely different name as one might think that `sendPerm` are `SendGivenPerm` and `SendGotPerm` which is not the case

src/test/resources/regressions/features/channels/channel-fail6.gobra
jcp19: Thanks :)

PR #204, opened 2021-02-02
Fix issue 164

### LINKED ISSUE #164: Different results for seemingly equivalent code
In the following code, the fold error is not reported in the function `foldSingletonLessThanError`
```go
type Singleton struct {
    x int
}

pred (s *Singleton) lessThan(bound int) {
	acc(s) && s.x <= bound
}

requires x >= 0
func foldSingletonLessThan(x int) {
	s@ := Singleton{}
	fold (&s).lessThan!<x!>()
}

// error: x may be less than the value of the singleton
func foldSingletonLessThanError(x int) {
	s@ := Singleton{}
	//:: ExpectedOutput(fold_error)
	fold (&s).lessThan!<x!>()
}
```
If we change `foldSingletonLessThanError` to
```go
// error: x may be less than the value of the singleton
func foldSingletonLessThanError(x int) {
	s@ := Singleton{0} // added a 0 here
	//:: ExpectedOutput(fold_error)
	fold (&s).lessThan!<x!>()
}
```
then the error is reported
> [info] 	<62:2> Fold might fail. 
[info] Assertion s.x <= bound might not hold

Shouldn't both examples produce the same result as it is my understanding that `s@ := Singleton{0}` is equivalent to `s@ := Singleton{}`

jcp19:
Fixes #164. The bug was triggered when local variables with the same name were declared in different functions. In that case, `variables` would return a list with repeated variables. In turn, this would lead to redeclarations of a variable in the same block, in the method `blockD` of the file `Desugar.scala`.

For reference, this is how the current implementation of `blockD` looks like
```scala
    def blockD(ctx: FunctionContext)(block: PBlock): in.Stmt = {
      val vars = info.variables(block) map localVarD(ctx)
      val ssW = sequence(block.nonEmptyStmts map (s => seqn(stmtD(ctx)(s))))
      in.Block(vars ++ ssW.decls, ssW.stmts ++ ssW.res)(meta(block))
    }
```

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/implementation/TypeInfoImpl.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/TypeInfoImpl.scala
@@ -8,6 +8,7 @@ package viper.gobra.frontend.info.implementation
 
 import com.typesafe.scalalogging.StrictLogging
 import org.bitbucket.inkytonik.kiama.attribution.Attribution
+import org.bitbucket.inkytonik.kiama.util.Position
 import viper.gobra.ast.frontend._
 import viper.gobra.frontend.Config
 import viper.gobra.frontend.info.base.SymbolTable.{Regular, TypeMember, UnknownEntity, lookup}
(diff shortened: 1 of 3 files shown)


ArquintL:
What is the reason that `info.variables` returns multiple copies of a local variable in the first place? Do the scopes get mixed up?

jcp19:
> What is the reason that `info.variables` returns multiple copies of a local variable in the first place? Do the scopes get mixed up?

I believe that this is caused by the global search through all the nodes in the tree in the computation of `variablesMap`. because this is done for all ids in the tree, the collected Vector can contain repeated ids when local variables with the same name in different functions are declared.
```scala
  private lazy val variablesMap: Map[PScope, Vector[PIdnNode]] = {
    val ids: Vector[PIdnNode] = tree.nodes collect {
      case id: PIdnDef              => id
      case id: PIdnUnk if isDef(id) => id
    }

    ids.groupBy(enclosingIdScope)
  }

  override def variables(s: PScope): Vector[PIdnNode] =
    variablesMap.getOrElse(s, Vector.empty).distinct.sortWith(_.name < _.name)
```

PS: Your question made me realize that something may be wrong with `enclosingIdScope` instead, because the repeated occurrences of ids occur in different scopes

ArquintL:
Exactly, I'm just suprised that this hasn't occurred so far as using the same variable name in different scopes seems common.
Note however that G allows to redeclare variables with the same name in nested scopes

Felalolf:
Since only a single variable with the same name should be defined in the same scope, it indeed seems like enclosingIDScope might not do entirely what we want it to do.

jcp19:
The problem is caused by the fact that `ids.groupBy(enclosingIdScope)` groups the ids by their corresponding scope. **Two scopes which are structurally the same correspond to the same key, independently of their position.** If two id nodes with the same name occur in different functions with **exactly the same body**, then they will have the same key: a node corresponding to the body of the function. Thus, both id nodes are added to the corresponding value in the map, which looks like a variable is defined twice in that scope.

My last commit adds location information to the keys of `variablesMap`. @Felalolf I opted to change the type of `variablesMap` instead of doing a `map(((scope, pos), vars) => (scope, vars))` at the end of its definition. This way, the variable list returned by `variables` returns a list of the exact id nodes that occur in the scope passed as an argument. Otherwise, this might not be the case (it could return the nodes used in another scope which is structurally equal to the one passed as an argument). I believe it is better to maintain precise position information whenever possible.

ArquintL (approved):
Thanks for looking into this again. I'm convinced by the explanation and like the solution

jcp19:
While testing the solution, I also discovered that the generated viper file would often contain multiple declarations of the default value of a Pred type in the corresponding domain, e.g.
```dafny
domain Pred_S0 {
  function Pred_S0_make_1(x0: Ref, x1: Int): Pred_S0
  function Pred_S0_make_0(x0: Ref, x1: Slice[Ref]): Pred_S0
  
  function Pred_S0_default(): Pred_S0
  function Pred_S0_default(): Pred_S0
}

```
where S is the serialization of a Pred type. This is due to a mistake by me: the line
```scala
    funcs ++= genDefaultFuncMap.get(S)
```
(file `DefuncComponentImpl`) was executed multiple times inside a loop when it should be executed only once.

Felalolf (approved):

PR #206, opened 2021-02-03
More Examples and Examples with Seeded Bugs

ArquintL:


src/test/resources/regressions/examples/all_nearest_smaller_values.gobra
```diff
@@ -0,0 +1,218 @@
...
+    left[i] = 0
+  } else {
+    left[i] = h
+  }
+
+  stack.push(i)
+
+  i = i + 1
+}
+*/
```
jcp19, line 218: is this comment taken into account when measuring the LoC?

jcp19 (approved):
The examples seem fine to me. I did not look too much at the examples with seeded bugs. There is one failing test that needs to be fixed before merging

src/test/resources/regressions/examples/all_nearest_smaller_values.gobra
ArquintL: I'll double check this

PR #208, opened 2021-02-04
added visitor pattern example for paper

Felalolf:
title. Also fixed some bugs.


ArquintL (approved):
Looks good, I really like the example

PR #213, opened 2021-02-10
Implicit Unqualified Imports

### LINKED ISSUE #158: Support for the error type
The type `error` is ubiquitous in *Go* code and, in particular, it is used extensively throughout the *SCION* codebase. As such, it would be very useful to implement it in *Gobra*. 

It is defined as
```go
type error interface {
    Error() string
}
```

Gobra does not support non-empty interfaces and the `string` type at the moment. Nevertheless, it may still be worth to implement a pared down version of `error` (e.g.  as an empty interface) so that it can parse (some) code that relies on it (e.g. code that does not call `Error()` on a receiver with static type `error`). SCION is full of usages of `error` that satisfy this criterion - commonly, `error` is just used as a return type.

ArquintL:
This PR adds the possibility to declare built-in members that will be implicitly and unqualifiedly imported in Gobra files.
For caching reasons, only a single "builtin" package is permitted. However, built-in members can be split across multiple files of said package.
In particular, the `error` interface is added as an empty interface as we do not have support for strings yet.

In addition, this PR fixes some contextual bugs in the desugarer that have lead to crashes when encountering an interface located in an imported package. @Felalolf please have a look at commit 23ddbabfda056b4d07f713e1a9f8d368794869dc.

Fixes #158

### DIFF
--- a/src/main/resources/README.md
+++ b/src/main/resources/README.md
@@ -1,6 +1,10 @@
 # Resources
 This folder contains resources that are included in the Gobra jar file.
 
+## Built-Ins
+The `.gobra` files in the `builtin` folder form a package and thus need to have the same package clause and be located in the top-level directory (i.e. subdirectories are ignored).
+Members declared in these files will be available without any import in the file being verified and can be accessed without qualifier.
+
 ## Stubs
 The `.gobra` files contained in the `stubs` folder are particularly noteworthy:
 Gobra uses them when resolving package imports.
--- a/src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
@@ -10,6 +10,7 @@ import org.bitbucket.inkytonik.kiama.relation.Relation
 import org.bitbucket.inkytonik.kiama.util.Entity
 import org.bitbucket.inkytonik.kiama.util.Messaging.{Messages, message}
 import viper.gobra.ast.frontend._
+import viper.gobra.frontend.PackageResolver.{AbstractImport, BuiltInImport, RegularImport}
 import viper.gobra.frontend.info.base.BuiltInMemberTag
 import viper.gobra.frontend.info.base.BuiltInMemberTag.{BuiltInMPredicateTag, BuiltInMethodTag}
 import viper.gobra.frontend.{PackageResolver, Parser}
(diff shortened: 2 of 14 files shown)


src/main/resources/README.md
```diff
@@ -1,6 +1,10 @@
 # Resources
 This folder contains resources that are included in the Gobra jar file.
 
+## Built-Ins
+The `.gobra` files in the `builtin` folder form a package and thus need to have the same package clause and be located in the top-level directory (i.e. subdirectories are ignored).
+Members declared in these files will be available without any import in the file being verified and can be accessed without qualifier.
+
```
jcp19, line 7: To add to what @ArquintL said, it is worth noting that this is the same approach as in the Go compiler/std library. All definitions in the `builtin` package  are available without an explicit import and do not require qualification.

jcp19 (approved):
It looks good! Eager to have it merged 😁

src/main/resources/README.md
ArquintL: However, I couldn't find a source that the Go compiler actually handles these built-ins like we do it here. The documentation of the `builtin` package [here](https://golang.org/pkg/builtin/) only mentions that `builtin` is a package for documentation purposes

jcp19:
I went ahead and added support for `panic`, according to the definition in the Go compiler. For now, a panic is modeled as an unrecoverable error, because there is no support for defer statements, and thus, a call to `recover` may not produce any effect:
```
// The recover built-in function allows a program to manage behavior of a
// panicking goroutine. Executing a call to recover inside a deferred
// function (but not any function called by it) stops the panicking sequence
// by restoring normal execution and retrieves the error value passed to the
// call of panic. If recover is called outside the deferred function it will
// not stop a panicking sequence. In this case, or when the goroutine is not
// panicking, or if the argument supplied to panic was nil, recover returns
// nil. Thus the return value from recover reports whether the goroutine is
// panicking.
func recover() interface{}
```
I opted to not include `recover` for now, as it is pretty much meaningless in the current state of Gobra.

src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
```diff
@@ -279,32 +279,43 @@ trait MemberResolution { this: TypeInfoImpl =>
...
-          case _ => UnknownEntity()
-        }
-      }
+      id =>
+        // Go is weird in the sense that it let's you redeclare built-in identifiers such as "error" but won't complain
+        // about the redeclaration. If the redeclaration happens in the same package, the redeclaration is
+        // used (instead of the built-in one). If the redeclaration happens in an imported package Go's behavior is not
+        // fully clear since "error" is not exported. However, we give higher precedence to the built-in one in Gobra
+        // to approximate Go's behavior.
+        tryUnqualifiedBuiltInPackageLookup(id).getOrElse(tryUnqualifiedRegularPackageLookup(id))
```
Felalolf, line 296: What would happen if something is defined in the builtin package and a user defined pacakge? I'd guess that with the current setup the the user defined entity is overshadowed by builtin entity.

Felalolf (commented):
One comment, everything else seems fine

src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
ArquintL: There are two cases to distinguish:
- 1. you try to shadow a builtin member in the current package. You can find this case in `error-simple2.gobra`. Gobra and Go will use the redeclared member (Gobra uses it as this function is only invoked as a fallback if entity resolution fails)
- 2. you import a package that shadows a builtin member: as far as I can tell, all builtin members start with a lower case letter so they are ignored in Go. As Gobra does not care yet about an identifier's first letter, we give higher precendence to the builtin member (this is happening on this code line). Imho this mimics Go's behavior the best

Felalolf (approved):

PR #216, opened 2021-02-15
Removed warnings in main code

Felalolf:
As the title states, the main scala files should not produce warnings anymore.
From now on, we should use `@unused` where necesarry.


ArquintL (approved):
Cool!

jcp19 (approved):

PR #218, opened 2021-02-17
Continue string parsing

jcp19:
This PR continues the work by @ArquintL on the `string-parsing` branch from almost a year ago. In particular, it re-uses its work on adapting the parser and extends the type system and encoding. The string encoding keeps track of the length of a string, allowing Gobra to verify properties related to string lengths.


src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -221,7 +222,7 @@ object Parser {
       val r = s"($finalTokenRequiringSemicolon)((?:$ignoreComments|$ignoreWhitespace)*)$$".r
       // group(1) contains the finalTokenRequiringSemicolon after which a semicolon should be inserted
       // group(2) contains the line's remainder after finalTokenRequiringSemicolon
-      r.replaceAllIn(line, m => m.group(1) ++ ";" ++ m.group(2))
+      r.replaceAllIn(line, m => StringEscapeUtils.escapeJava(m.group(1) ++ ";" ++ m.group(2)))
```
Felalolf, line 225: what is this doing?

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -313,8 +313,10 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
         (n, exprOrTypeType(n.left), exprOrTypeType(n.right)) match {
           case (_: PEquals | _: PUnequals, l, r) => comparableTypes.errors(l, r)(n)
           case (_: PAnd | _: POr, l, r) => assignableTo.errors(l, AssertionT)(n) ++ assignableTo.errors(r, AssertionT)(n)
+          case (_: PLess | _: PAtMost | _: PGreater | _: PAtLeast, l, r) if l == StringT && r == StringT => noMessages
```
Felalolf, line 316: Are these operations defined on strings?

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -329,9 +331,9 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
 
     case PLength(op) => isExpr(op).out ++ {
       exprType(op) match {
-        case _: ArrayT | _: SliceT => noMessages
+        case _: ArrayT | _: SliceT | StringT => noMessages
         case _: SequenceT => isPureExpr(op)
-        case typ => error(op, s"expected an array, sequence or slice type, but got $typ")
+        case typ => error(op, s"expected an array, string, sequence or slice type, but got $typ")
```
Felalolf, line 336: I am just addig this comment to an arbitrary line of code. In Go, can you write `x := &str[i]` for a string `str`? If yes, then this should be forbidden by gobra.

src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
```diff
@@ -0,0 +1,168 @@
...
+    * [ len(s: string) ] -> strLen([s])
+    * [ (s1: string) + (s2: string) ] -> strConcat([ s1 ], [ s2 ])
+    */
+  override def expr(ctx: Context): in.Expr ==> CodeWriter[vpr.Exp] = {
+
+    def goE(x: in.Expr): CodeWriter[vpr.Exp] = ctx.expr.translate(x)(ctx)
+
+    default(super.expr(ctx)) {
+      case (e: in.DfltVal) :: ctx.String() / Exclusive =>
+        val encodedStr = stringBeginning + "Default"
```
Felalolf, line 50: Which string is actually the default? Then this should be used instead of "Default". Since your makeFunc results in unique functions, you get an unsoundness otherwiese, e.g. `var x string; assert x == "Default"` would verify.

src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
```diff
@@ -0,0 +1,168 @@
...
+    * Encodes a string variable as an vpr.Int
+    */
+  override def variable(ctx: Context): in.BodyVar ==> vpr.LocalVarDecl = {
+    case v :: ctx.String() / Exclusive if typ(ctx).isDefinedAt(v.typ) =>
+      val (pos, info, errT) = v.vprMeta
+      vpr.LocalVarDecl(v.id, vpr.Int)(pos, info, errT)
+    case v :: ctx.String() / Shared if typ(ctx).isDefinedAt(v.typ) =>
+      val (pos, info, errT) = v.vprMeta
+      vpr.LocalVarDecl(v.id, vpr.Ref)(pos, info, errT)
+  }
```
Felalolf, line 77: this is the same as the default implementation, isn't it? So you can just delete this.

src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
```diff
@@ -0,0 +1,168 @@
...
+
+  /**
+    * Generates
+    *   function strConcat(l: Int, r: Int): Int
+    * where l and r are string ids
+    */
+  private val concatFuncName: String = "strConcat"
+  private val concatFunc: vpr.DomainFunc = vpr.DomainFunc(
+    name = concatFuncName,
+    formalArgs = Seq(vpr.LocalVarDecl(Names.freshName, vpr.Int)(), vpr.LocalVarDecl(Names.freshName, vpr.Int)()),
```
Felalolf, line 118: Here, you do not have to use freshName. You should avoid using freshName if possible.

Felalolf: same for many of the other functions. I am not repeating it.

Felalolf (changes requested):
There are now three wanring when compiling the main files. Please resolve them. We should maintain the zero warnings.

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -886,6 +892,10 @@ case class BoolLit(b: Boolean)(val info: Source.Parser.Info) extends Lit {
   override def typ: Type = BoolT(Addressability.literal)
 }
 
+case class StringLit(s: String)(val info: Source.Parser.Info) extends Lit {
```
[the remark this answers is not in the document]
jcp19: Do you mean making `StringLit` a subtype of `StringOperation`? If so, I do not think that should be done because
(1) even though `StringLit` and `StringOperation` have the same type, they get their addressability from to different fields of the `Addressability` object: `literal` and `rValue`, respectively.
(2) for consistency: `IntLit` does not extend `IntOperation`. `BoolLit` also does not extend `BoolOperation`

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
jcp19: Yes, they perform lexicographic comparisons

src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
jcp19: I do not think that this is unsound. The reason is that, for every string other than the default, the suffix of the unique function is the hexadecimal representation of the string. This serves the purpose of having injective suffixes with valid characters (if I just added the original string value as a suffix, it could contain spaces or chars that are not acceptable in identifiers). Because "Default" is not the hexadecimal representation of any string, there should not be a problem.

src/main/scala/viper/gobra/ast/internal/Program.scala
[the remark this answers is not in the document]
Felalolf: ok

src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
Felalolf: But then you can show the following: `var s string; assert s != ""`

jcp19: I realized that there is a issue here though. Given the current way we encode things, the string literal "" and DfltVal will lead to different encodings even though they are meant to be the same. As such, the following verifies even though it shouldn't:
```go
    a4 := ""
    var a5 string
    assert a4 != a5
```
This has been fixed in the latest commit

src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
jcp19: Yes, you are right.

src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
jcp19: My bad. Fixed in the latest commit

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
jcp19: I did not add support for indexing strings for now. As such, there is no need to explicitly forbid this.

build.sbt
```diff
@@ -38,13 +38,21 @@ lazy val gobra = (project in file("."))
     libraryDependencies += "org.fusesource.jansi" % "jansi" % "1.17.1", // For colouring Logback output
     // libraryDependencies += "org.typelevel" %% "cats-core" % "1.6.0", // cats
     libraryDependencies += "org.apache.commons" % "commons-lang3" % "3.9", // for SystemUtils
+    libraryDependencies += "org.apache.commons" % "commons-text" % "1.9", // for escaping strings in parser preprocessor
+    libraryDependencies += "commons-codec" % "commons-codec" % "1.15", // for obtaining the hex encoding of a string
```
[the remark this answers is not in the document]
jcp19: Both projects are protected by an Apache v2 license. Given that we do not redistribute them in source-code form, my understanding is that this license only requires us to provide a copy of their license when redistributing it in binary form (e.g. GobraIDE, binary releases of Gobra)

https://github.com/apache/commons-text/blob/master/LICENSE.txt
https://github.com/apache/commons-codec/blob/master/LICENSE.txt

src/main/scala/viper/gobra/frontend/Parser.scala
jcp19: According to the [doc](https://commons.apache.org/proper/commons-text/javadocs/api-release/org/apache/commons/text/StringEscapeUtils.html#escapeJava(java.lang.String)), this escapes the special chars in a string according to the java string rules e.g.
```
 input string: He didn't say, "Stop!"
 output string: He didn't say, \"Stop!\"
```

Without this modification, the following line of Gobra code would cause an error during the preprocess step
```go
s3 := "\""
```
Error:
```
[info] Gobra has found 1 error(s):
[info] 	<13:10> strings0.gobra:13:10:error: '}' expected but '"' found
[info] 	s3 := """;
```

Felalolf (approved):

src/main/scala/viper/gobra/ast/internal/PrettyPrinter.scala
```diff
@@ -442,6 +443,7 @@ class DefaultPrettyPrinter extends PrettyPrinter with kiama.output.PrettyPrinter
 
   def showLit(l: Lit): Doc = l match {
     case IntLit(v, _) => v.toString
+    case StringLit(s) => s"${"\""}$s${"\""}"
```
ArquintL, line 446: this seems quite complicated. Wouldn't `"\"" <> s <> "\""` work as well?

src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
```diff
@@ -0,0 +1,156 @@
...
+
+    /**
+      * The length of every string literal in the program is axiomatized as
+      *   axiom {
+      *     strLen(literal()) == X
+      *   }
+      * where `literal` is one of the generated unique domain functions and X is the length of the corresponding string
+      */
+    val lenAxioms = strLengths.toSeq.map {
+      case (str, len) =>
```
ArquintL, line 122: You wouldn't need to store `len` in the map as you could calculate it here for the given `str`. However, you do not have to change it as it's so minor

ArquintL (approved):
Looks good! It's surprisingly concise, sorry for taking so long to review it

src/main/scala/viper/gobra/ast/internal/PrettyPrinter.scala
jcp19: Yes, that is a good suggestion 👍

src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
jcp19: Yes, it is indeed wasteful. Fixed that in the last commit

Issue #219, opened 2021-02-18
ManagedFileSystem crashes when importing stubs


jcp19:
The following excerpt causes Gobra to crash with a **java.nio.file.FileSystemAlreadyExistsException** (stack trace shown at the end of the issue):
```go
package path

import "time"

func ExpTimeToDuration(expTime uint8) time.Duration {
	return (time.Duration(expTime) + 1) * time.Duration(expTimeUnit) * time.Second
}
```
I'm assuming at the moment that you are running on Gobra on the `add-more-stubs` branch or in any other branch that incorporated its changes (and, in particular, that contains the stubs for the time package).

Once again, this issue only occurs when Gobra is run on sbt in the terminal. Running it from an IDE will sidestep the issue.

Stack trace:
```
[error] Exception in thread "main" java.nio.file.FileSystemAlreadyExistsException
[error] 	at jdk.zipfs/jdk.nio.zipfs.ZipFileSystemProvider.newFileSystem(ZipFileSystemProvider.java:102)
[error] 	at java.base/java.nio.file.FileSystems.newFileSystem(FileSystems.java:337)
[error] 	at java.base/java.nio.file.FileSystems.newFileSystem(FileSystems.java:286)
[error] 	at viper.gobra.frontend.PackageResolver$ManagedFileSystem.<init>(PackageResolver.scala:311)
[error] 	at viper.gobra.frontend.PackageResolver$BaseJarResource.filesystem$lzycompute(PackageResolver.scala:285)
[error] 	at viper.gobra.frontend.PackageResolver$BaseJarResource.filesystem(PackageResolver.scala:285)
[error] 	at viper.gobra.frontend.PackageResolver$BaseJarResource.<init>(PackageResolver.scala:286)
[error] 	at viper.gobra.frontend.PackageResolver$.$anonfun$getResource$2(PackageResolver.scala:112)
[error] 	at scala.Option.flatMap(Option.scala:283)
[error] 	at viper.gobra.frontend.PackageResolver$.getResource(PackageResolver.scala:104)
[error] 	at viper.gobra.frontend.PackageResolver$.getBuiltInResource(PackageResolver.scala:89)
[error] 	at viper.gobra.frontend.PackageResolver$.getLookupPath(PackageResolver.scala:135)
[error] 	at viper.gobra.frontend.PackageResolver$.resolve(PackageResolver.scala:50)
[...]

Issue #184, opened 2021-01-19
Gobra crashes with programs that rely on the built-in stubs


jcp19:
When a program such as [this](https://github.com/viperproject/gobra/blob/add-sync-stubs/src/test/resources/regressions/features/defunc/mutex1.gobra) is verified in Gobra, it crashes with
```
[error] Exception in thread "main" java.lang.IllegalStateException: managed file system can no longer be retained when it is already closed
[error] 	at viper.gobra.frontend.PackageResolver$ManagedFileSystem.retain(PackageResolver.scala:286)
[error] 	at viper.gobra.frontend.PackageResolver$JarFileResource.<init>(PackageResolver.scala:269)
[error] 	at viper.gobra.frontend.PackageResolver$JarResource.resolve(PackageResolver.scala:235)
[error] 	at viper.gobra.frontend.PackageResolver$JarResource.resolve$(PackageResolver.scala:234)
[error] 	at viper.gobra.frontend.PackageResolver$BaseJarResource.resolve(PackageResolver.scala:253)
[error] 	at viper.gobra.frontend.PackageResolver$BaseJarResource.resolve(PackageResolver.scala:253)
[error] 	at viper.gobra.frontend.PackageResolver$.$anonfun$getLookupPath$1(PackageResolver.scala:112)
[error] 	at scala.collection.immutable.Vector1.map(Vector.scala:1872)
[error] 	at scala.collection.immutable.Vector1.map(Vector.scala:375)
[error] 	at viper.gobra.frontend.PackageResolver$.getLookupPath(PackageResolver.scala:112)
[error] 	at viper.gobra.frontend.PackageResolver$.resolve(PackageResolver.scala:40)
[error] 	at viper.gobra.frontend.info.implementation.resolution.MemberResolution.parseAndTypeCheck$1(MemberResolution.scala:178)
[error] 	at viper.gobra.frontend.info.implementation.resolution.MemberResolution.$anonfun$tryPackageLookup$12(MemberResolution.scala:214)
[error] 	at scala.Option.getOrElse(Option.scala:201)
[error] 	at viper.gobra.frontend.info.implementation.resolution.MemberResolution.getTypeChecker$1(MemberResolution.scala:214)
[error] 	at viper.gobra.frontend.info.implementation.resolution.MemberResolution.tryPackageLookup(MemberResolution.scala:218)
[...]

PR #223, opened 2021-02-25
Simplify package resolver

jcp19:
In a discussion with @ArquintL, we realized that the logic for keeping track of the number of references to a filesystem in the `ManagedFileSystem` class in the `PackageResolver.scala` file is unnecessarily complex and has led to bugs that are hard to debug (issues #184 and #219) and that are not caught in our CI tests.

Furthermore, the code in question is only used to access the contents of the generated Gobra .jar (more precisely the stubs for the builtin methods in `src/main/resources/builtin/` and standard library methods in `src/main/resources/stubs`). A simpler way to do things would be to keep these two file systems open at all times during the execution of Gobra. This has the additional advantage of avoiding a complex logic for keeping track of references. Unlike the current version of Gobra, this approach has the advantage that each mentioned filesystem is only opened once. 

This PR also fixes issue #219 that occurred when importing a stub which, itself, imported another package from our stubs. Such setting would lead to `FileSystems.newFileSystem` being called twice with the same URI which resulted in a `FileSystemAlreadyExistsException`.

Testing was done locally. In fact, It's actually not easily possible to write a test case as tests executed with sbt test access resources as regular files and thus do not trigger this part of Gobra that deals with resources located in the JAR file.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/PackageResolver.scala
+++ b/src/main/scala/viper/gobra/frontend/PackageResolver.scala
@@ -58,7 +58,7 @@ object PackageResolver {
     * Similar to `resolve` but returns the package name of the files instead of the files themselves.
     * The returned package name can be used as qualifier for the implicitly qualified import.
     *
-    * @param n           implicitely qualified import for which a qualifier should be resolved
+    * @param n implicitely qualified import for which a qualifier should be resolved
     * @param includeDirs list of directories that will be used for package resolution before falling back to $GOPATH
     * @return qualifier with which members of the imported package can be accessed (right) or an error message (left)
     *         if no directory could be found or the directory contains input files having different package clauses
--- a/src/main/scala/viper/gobra/frontend/PackageResolver.scala
+++ b/src/main/scala/viper/gobra/frontend/PackageResolver.scala
@@ -7,8 +7,7 @@
 package viper.gobra.frontend
 
 import java.io.{Closeable, File, InputStream}
-import java.net.URI
-import java.nio.file.{FileSystems, Files, Path, Paths}
+import java.nio.file.{FileSystem, FileSystemAlreadyExistsException, FileSystems, Files, Path, Paths}
 import java.util.Collections
 
 import org.apache.commons.io.FilenameUtils
@@ -59,7 +58,7 @@ object PackageResolver {
     * Similar to `resolve` but returns the package name of the files instead of the files themselves.
     * The returned package name can be used as qualifier for the implicitly qualified import.
     *
-    * @param n implicitely qualified import for which a qualifier should be resolved
+    * @param n           implicitely qualified import for which a qualifier should be resolved
     * @param includeDirs list of directories that will be used for package resolution before falling back to $GOPATH
     * @return qualifier with which members of the imported package can be accessed (right) or an error message (left)
     *         if no directory could be found or the directory contains input files having different package clauses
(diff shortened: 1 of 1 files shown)


src/main/scala/viper/gobra/frontend/PackageResolver.scala
```diff
@@ -109,7 +95,13 @@ object PackageResolver {
...
       resource <- resourceUri.getScheme match {
         case s if s == fileUriScheme => Some(FileResource(Paths.get(resourceUri).toFile))
-        case s if s == jarUriScheme => Some(BaseJarResource(resourceUri, path))
+        case s if s == jarUriScheme =>
+          val fs = try {
+            FileSystems.newFileSystem(resourceUri, Collections.emptyMap[String, Any]())
+          } catch {
+            case _: FileSystemAlreadyExistsException => FileSystems.getFileSystem(resourceUri)
+          }
+          Some(JarResource(fs, path))
```
jcp19, line 104: Instead of relying on the exceptions thrown by the methods in `FileSystems`, I first tried to keep track of open file systems in the `PackageResolver` object with a `Map[URI, FileSystem]`. However, this would still lead to a `FileSystemAlreadyExistsException` when instantiating a FS with the URI of the `src/main/resources/builtin` folder. It would work OK with the URI from `src/main/resources/stubs` though. I assume that the first filesystem is initialized before package resolution.

ArquintL: Are both calls to `FileSystems.newFileSystem` with `src/main/resources/builtin` being made by the code listed in this class? I cannot get my head around why the map does not work

ArquintL (approved):
Looks good to me. I still don't understand why the map solution would not work but this version definitely works as a quick fix

src/main/scala/viper/gobra/frontend/PackageResolver.scala
jcp19: Yes, in my experiments with Map, this was the only place where I both created a new FileSystem and updated the corresponding entry in the Map. The first instantiation of a new FileSystem with the URI for `src/main/resources/builtin` would crash.

ArquintL: Might it be that Java complains about having two filesystems for the same JAR? I mean they point to distinct folders but have a common root (i.e. `resources`). Might be force trying in the future to use the same filesystem for `builtin` and `stubs`.

ArquintL:
I just noticed that your current suggestions does not close the filesystem

src/main/scala/viper/gobra/frontend/PackageResolver.scala
jcp19: I believe that the URIs for `builtin` and `stubs` are different. As such, I would be surprised if that were the case

ArquintL: They are, however maybe Java maps both of them to the same file system and thus throws an exception

jcp19:
> I just noticed that your current suggestions does not close the filesystem eventually

I chose to explicitly not close the filesystem because choosing a point in time to do it is not trivially easy. Assuming that a file system is open somewhere else as it may be the case for the builtin file, choosing a point to close it may cause some other component that is accessing the filesystem to throw an exception
We could close the files manually at a late point in the program by having a call to close() somewhere in after the parsing and type checking phase but it may not be clear to someone who is just looking through the code base why that is done at that point. As such, I opted to rely on the JVM runtime to close the file system at the end of the program

src/main/scala/viper/gobra/frontend/PackageResolver.scala
jcp19: Yes that might be the case indeed!

Issue #225, opened 2021-03-03
Parser error when CRLF is used as line separator


jcp19:
A recent PR from me introduced a bug in the parser when CRLF is used as the line separator (as in files created in Windows). For example, the following code leads to a parser error:
```go
package main

ghost
func f()
```

Error:
```
[info] Gobra has found 1 error(s):
[info] 	<3:1> quicksort.gobra:3:1:error: 'implements' expected but 'g' found
[info] ghost;r
```

I have an idea of what causes this bug and I should open a PR soon.

PR #226, opened 2021-03-03
Fix issue 225

jcp19:
Fixes issue #225. The parsing bug was caused by the introduction of the following line in a previous commit
```scala
r.replaceAllIn(line, m => StringEscapeUtils.escapeJava(m.group(1) ++ ";" ++ m.group(2)))
```
My understanding is that this causes `CR` characters to be escaped when they appear at the end of a line (e.g. in files created on a Windows machine).

This PR solves the issue by splitting lines on "\r\n|\n" whereas they were previously split only on "\n", which removes the occurrences of "\r\n" before parsing the contents of string literals.

### DIFF
--- a/.github/license-check/config.json
+++ b/.github/license-check/config.json
@@ -53,7 +53,14 @@
       "project/project/**",
       "project/target/**",
       "target/**",
-      "src/main/**"
+      "src/main/**",
+      "src/test/resources/regressions/issues/windows-line-ending/**"
+    ],
+    "license": "./.github/license-check/headers/CC0.txt"
+  },
+  {
+    "include": [
+      "src/test/resources/regressions/issues/windows-line-ending/*.gobra"
     ],
     "license": "./.github/license-check/headers/CC0.txt"
   },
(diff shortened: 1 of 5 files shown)


ArquintL (approved):
Looks good to me. Maybe add a comment to the test case that this includes \r\n to test the parser‘s behavior as it is invisible. In addition, please look at the header - for some reason, the checker is not happy

ArquintL:
> In addition, please look at the header - for some reason, the checker is not happy

could it be that the license checker is not smart enough either to accept \r\n? ^^

jcp19:
> > In addition, please look at the header - for some reason, the checker is not happy
> 
> could it be that the license checker is not smart enough either to accept \r\n? ^^

That may be the case. In that case, I would need to change the license checker to either ignore the char `\r` or to accept a list of licenses instead of a single license and then add a copy of each license with the \r\n line ending

ArquintL:
Sounds good. I totally agree that having just a few files with Windows line ending make sense to run Gobra on them. The reason why we have no/only few such files is that Git typically takes care of this and changes the line endings on Windows before committing.

PR #227, opened 2021-03-07
Add support for the Permission type

jcp19:
Added support for the `perm` type. Besides, changed the sync stubs and the parallel examples to use the new perm type instead of the old hack where two integer arguments were used in place one single perm argument. Whenever possible, I tried to make the interpretation of perm expressions similar to that of Viper. For example, integer expressions that are subexpressions of a `perm` expressions are interpreted in the same way as integers, and thus, the following code verifies:

```go
requires acc(i, 0)
ensures acc(i, (1/2)/4)
func test9(i *int) { } 
```

This is analogous to the following Viper excerpt:
```dafny
field i: Int

method test9(o: Ref)
  requires acc(o.i, 0/1)
  ensures acc(o.i, (1/2)/4) {}
```


src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1324,6 +1329,31 @@ object Desugar {
...
+
+          case PUnequals(left, right) if info.typOfExprOrType(left) == PermissionT || info.typOfExprOrType(right) == PermissionT =>
+            violation(left.isInstanceOf[PExpression], s"Expected an expression but got $left instead.")
+            violation(right.isInstanceOf[PExpression], s"Expected an expression but got $right instead.")
+            for {
+              l <- permissionD(ctx)(left.asInstanceOf[PExpression])
+              r <- permissionD(ctx)(right.asInstanceOf[PExpression])
+            } yield in.UneqCmp(l, r)(src)
+
+          case PLess(left, right) if info.typ(left) == PermissionT || info.typ(right) == PermissionT =>
```
Felalolf, line 1348: I think it would be better to make the distinction only once at the desugaring of PLess and not once for every possibility.

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -818,6 +819,7 @@ case class PFullPerm() extends PPermission
 case class PNoPerm() extends PPermission
 case class PFractionalPerm(left: PExpression, right: PExpression) extends PPermission with PBinaryGhostExp
 case class PWildcardPerm() extends PPermission
+case class PEpsilonPerm() extends PPermission
```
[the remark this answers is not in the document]
jcp19: These are the infinitesimal "counting" permission from Chalice. At the moment, epsilon permissions are not supported by the backends, even though they exist in the silver ast. If there are no plans to add support for them in the Viper backends, I will remove this node.

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -403,6 +403,18 @@ case class FullPerm(info: Source.Parser.Info) extends Permission
 case class NoPerm(info: Source.Parser.Info) extends Permission
 case class FractionalPerm(left: Expr, right: Expr)(val info: Source.Parser.Info) extends Permission
 case class WildcardPerm(info: Source.Parser.Info) extends Permission
+case class EpsilonPerm(info: Source.Parser.Info) extends Permission
+case class PermMinus(exp: Expr)(val info: Source.Parser.Info) extends Permission
+case class PermAdd(left: Expr, right: Expr)(val info: Source.Parser.Info) extends BinaryExpr("+") with Permission
+case class PermSub(left: Expr, right: Expr)(val info: Source.Parser.Info) extends BinaryExpr("-") with Permission
+case class PermMul(left: Expr, right: Expr)(val info: Source.Parser.Info) extends BinaryExpr("*") with Permission
+case class IntPermMul(left: Expr, right: Expr)(val info: Source.Parser.Info) extends BinaryExpr("*") with Permission
```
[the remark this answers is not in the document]
jcp19: In practice, I did not distinguish between the two. Furthermore, an `IntPermMul` is never generated. However, I opted to include it anyway for the reason that, if we change the frontend in the future in such a way that this distinction makes sense, than we would already have the corresponding node. I do not mind removing it though

src/main/scala/viper/gobra/frontend/info/implementation/property/TypeMerging.scala
```diff
@@ -21,8 +21,11 @@ trait TypeMerging extends BaseProperty { this: TypeInfoImpl =>
       case (Single(lst), Single(rst)) =>
         if (identicalTypes(lst, rst)) Some(lst) else {
           (lst, rst) match {
-            case (a, IntT(config.typeBounds.UntypedConst)) if underlyingType(a).isInstanceOf[IntT] => Some(a)
-            case (IntT(config.typeBounds.UntypedConst), b) if underlyingType(b).isInstanceOf[IntT] => Some(b)
+            case (a, UNTYPED_INT_CONST) if underlyingType(a).isInstanceOf[IntT] => Some(a)
+            case (UNTYPED_INT_CONST, b) if underlyingType(b).isInstanceOf[IntT] => Some(b)
+            case (PermissionT, PermissionT) => Some(PermissionT)
```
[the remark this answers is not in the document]
jcp19: Yes, you are right! Did not notice this

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
```diff
@@ -280,8 +284,13 @@ trait GhostExprTyping extends BaseTyping { this: TypeInfoImpl =>
 
       case _: PBoolLit | _: PIntLit | _: PNilLit | _: PStringLit => true
 
+      // Might change at some point
       case n: PInvoke => (exprOrType(n.base), resolve(n)) match {
-        case (Right(_), Some(_: ap.Conversion)) => false // Might change at some point
+        case (Right(_), Some(c: ap.Conversion)) =>
+          c.typ match {
+            case PPermissionType() => n.args.map(go).forall(identity)
```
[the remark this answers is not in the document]
jcp19: my bad, that is obvious

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1273,6 +1276,8 @@ object Desugar {
             case p => Violation.violation(s"encountered unexpected pattern: $p")
           }
 
+          case exp: PExpression if info.typ(exp) == PermissionT && !ignorePerm => permissionD(ctx)(exp)
```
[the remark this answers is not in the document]
jcp19: Permissions are desugared using the `permissionD` method in order to avoid adding a case split in every expression where a perm value may occur (PDiv, PConversion, PNegation, etc). However, in cases where the desugaring of a perm expression is done in the same way as for every other type (e.g. desugaring a NameOperand or a function call that yields a value of perm type), we can use the already existing desugarer instead of adding these cases to `permissionD`.
Thus, the `ignorePerm` argument skips over this case and desugarers the expression according to the following cases in the `exprD` method. Without this argument, the method would loop forever.

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -2491,15 +2522,39 @@ object Desugar {
...
+          case (PermissionT, IntT(_)) => for { vl <- permissionD(ctx)(l); vr <- goE(r) } yield in.PermDiv(vl, vr)(src)
+          case (IntT(_), IntT(_)) => for { vl <- goE(l); vr <- goE(r) } yield in.FractionalPerm(vl, vr)(src)
+          case err => violation(s"This case should be unreachable, but got $err")
+        }
+        case PNegation(exp) => for {e <- permissionD(ctx)(exp)} yield in.PermMinus(e)(src)
+        case PAdd(l, r) => for { vl <- permissionD(ctx)(l); vr <- permissionD(ctx)(r) } yield in.PermAdd(vl, vr)(src)
+        case PSub(l, r) => for { vl <- permissionD(ctx)(l); vr <- permissionD(ctx)(r) } yield in.PermSub(vl, vr)(src)
+        case PMul(l, r) =>
+          (info.typ(l), info.typ(r)) match {
+            case (IntT(_), PermissionT) | (PermissionT, IntT(_)) | (PermissionT, PermissionT) =>
```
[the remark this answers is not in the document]
jcp19: I introduced it here for some reason but it was not meant to be committed, thanks for noticing

src/main/scala/viper/gobra/reporting/DefaultErrorBackTranslator.scala
```diff
@@ -71,6 +71,8 @@ object DefaultErrorBackTranslator {
         SeqIndexExceedsLengthError(node, index)
       case vprrea.SeqIndexNegative(Source(node), Source(index)) =>
         SeqIndexNegativeError(node, index)
+      case vprrea.DivisionByZero(info) =>
+        DivisionByZeroReason(Source.unapply(info))
```
jcp19, line 75: Answering https://github.com/viperproject/gobra/pull/227/files#r589612593

In the particular file that I mention in my comment (src/test/resources/regressions/features/fractional_permissions/fields/fail3.gobra), Gobra would crash with a match error if, instead of `case vprrea.DivisionByZero(info) =>`, I would have written the cases `vprrea.DivisionByZero(Source(info)) =>` and `vprrea.DivisionByZero(Syntethized(info)) =>`. I assume that it may be indeed a bug where the source information is not recorded but I did not look too much at this because it seems to be an unrelated issue.

src/main/scala/viper/gobra/reporting/VerifierError.scala
```diff
@@ -210,6 +210,14 @@ case class SeqIndexNegativeError(node: Source.Verifier.Info, index: Source.Verif
   override def message: String = s"Index ${index.origin.tag.trim} into ${node.origin.tag.trim} might be negative"
 }
 
+// João, 06/03/2021: unlike the other subtypes of VerificationErrorReason, DivisionByZeroReason has an Optional argument.
+// This has to do with the fact that, in our tests, there are cases where a division by zero occurs but we cannot retrieve
+// a corresponding Source.Verifier.Info. E.g. src/test/resources/regressions/features/fractional_permissions/fields/fail3.gobra
```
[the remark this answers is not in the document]
jcp19: Please take a look at https://github.com/viperproject/gobra/pull/227/files#r589708184

src/main/scala/viper/gobra/translator/encodings/PermissionEncoding.scala
```diff
@@ -44,8 +44,15 @@ class PermissionEncoding extends LeafTypeEncoding {
...
-      case fp@ in.FractionalPerm(l, r) => for {vl <- goE(l); vr <- goE(r)} yield withSrc(vpr.FractionalPerm(vl, vr), fp)
       case wp: in.WildcardPerm => unit(withSrc(vpr.WildcardPerm(), wp))
+      case ep: in.EpsilonPerm => unit(withSrc(vpr.EpsilonPerm(), ep))
+      case co@ in.Conversion(_, exp) => for { e <- goE(exp) } yield withSrc(Function.untupled(e.withMeta), co)
+      case pm@ in.PermMinus(exp) => for { e <- goE(exp) } yield withSrc(vpr.PermMinus(e), pm)
+      case fp@ in.FractionalPerm(l, r) => for {vl <- goE(l); vr <- goE(r)} yield withSrc(vpr.FractionalPerm(vl, vr), fp)
+      case pa@ in.PermAdd(l, r) => for {vl <- goE(l); vr <- goE(r)} yield withSrc(vpr.PermAdd(vl, vr), pa)
+      case ps@ in.PermSub(l, r) => for {vl <- goE(l); vr <- goE(r)} yield withSrc(vpr.PermSub(vl, vr), ps)
+      case pm@ in.PermMul(l, r) => for {vl <- goE(l); vr <- goE(r)} yield withSrc(vpr.PermMul(vl, vr), pm)
+      case im@ in.IntPermMul(l, r) => for {vl <- goE(l); vr <- goE(r)} yield withSrc(vpr.IntPermMul(vl, vr), im)
```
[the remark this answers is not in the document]
jcp19: I could in principle remove `IntPermMul` altogether from the language without much consequence, given that something like `2 * 1/3` is desugared into `2/1 * 1/3`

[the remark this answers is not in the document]
jcp19: I removed it in the last commit

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -316,10 +316,15 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
+              (l == PermissionT && r.isInstanceOf[IntT]) || (l.isInstanceOf[IntT] && r == PermissionT) => noMessages
           case (_: PLess | _: PAtMost | _: PGreater | _: PAtLeast, l, r) =>
             assignableTo.errors(l, UNTYPED_INT_CONST)(n) ++ assignableTo.errors(r, UNTYPED_INT_CONST)(n)
           case (_: PAdd, l, r) if l == StringT && r == StringT => noMessages
+          case (_: PAdd | _: PSub | _: PMul | _: PMod | _: PDiv, l, r) if l == PermissionT || r == PermissionT ||
+            getTypeFromCtxt(n.asInstanceOf[PNumExpression], mustBeUntypedInt = false).contains(PermissionT) =>
+              assignableTo.errors(l, PermissionT)(n) ++ assignableTo.errors(r, PermissionT)(n)
           case (_: PAdd | _: PSub | _: PMul | _: PMod | _: PDiv, l, r) =>
             assignableTo.errors(l, UNTYPED_INT_CONST)(n) ++ assignableTo.errors(r, UNTYPED_INT_CONST)(n) ++
               numExprWithinTypeBounds(n.asInstanceOf[PNumExpression])
```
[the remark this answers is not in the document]
jcp19: This was indeed too complicated. I simplified it in the last commit. I did not introduce the functions you suggested given that they would be called only once (unless I am overlooking other places where they could simplify the existent implementation).

src/main/scala/viper/gobra/reporting/VerifierError.scala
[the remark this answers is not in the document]
Felalolf: I see, yeah, then just file a bug.

src/main/scala/viper/gobra/frontend/Desugar.scala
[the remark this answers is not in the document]
jcp19: I removed the ignorePerm in the latest commits

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1325,33 +1322,69 @@ object Desugar {
...
-              l <- go(left);
+              l <- go(left)
               r <- go(right)
-              res = if (info.typ(left) == StringT && info.typ(right) == StringT) {
-                in.Concat(l, r)(src)
-              } else {
-                in.Add(l, r)(src)
+              res = (info.typ(left), info.typ(right)) match {
+                case (StringT, StringT) => in.Concat(l, r)(src)
+                case _ => in.Add(l, r)(src)
```
ArquintL, line 1387: `left` or `right` cannot be of type `PermissionT` as these cases are handled (with an early return) by `maybePermissionD`, can they?

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -1207,19 +1208,14 @@ object Parser {
...
-      "write" ^^^ PFullPerm() |
-      "none" ^^^ PNoPerm() |
-      "_" ^^^ PWildcardPerm()
-
-    lazy val fractionalPermission: Parser[PFractionalPerm] =
-      expression into {
-        case d@PDiv(left, right) => success(PFractionalPerm(left, right).at(d))
-        case e => failure(s"expected a fractional permission amount expressed as a division but got $e")
-      }
+      "writePerm" ^^^ PFullPerm() |
```
ArquintL, line 1217: Did we decide to change these keywords? If so, I think you'll need to update the list of keywards accordingly in this file

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -704,9 +709,43 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
+          // The following case infers the type of an literal expression when it occurs inside a composite literal.
+          // For example, it infers that the expression `1/2` in `seq[perm]{ 1/2 }` has type perm. Notice that the whole
+          // expression would be parsed as
+          //   PCompositeLit(
+          //     PSequenceType(PPermissionType()),
+          //     PLiteralValue(Vector(
+          //       PKeyedElement(
+          //         None,
+          //         PExpCompositeVal(PDiv(PIntLit(BigInt(1)), PIntLit(BigInt(2))))))))
+          case comp if comp.isInstanceOf[PCompositeVal] => comp match {
```
ArquintL, line 722: why didn't you use `case comp: PCompositeVal => ...`?

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
```diff
@@ -151,6 +154,7 @@ trait GhostExprTyping extends BaseTyping { this: TypeInfoImpl =>
 
     case expr: PPermission => expr match {
       case PFullPerm() => noMessages
+      case PWildcardPerm() => noMessages
       case PNoPerm() => noMessages
       case fp@ PFractionalPerm(left, right) =>
```
ArquintL, line 159: Is `PFractionalPerm` still needed or can we delete the AST node?

src/test/resources/regressions/features/fractional_permissions/fields/fail3.gobra
```diff
@@ -9,10 +9,10 @@ type cell struct{
 
 const divisor = 0
 
-//:: ExpectedOutput(type_error)
+//:: ExpectedOutput(contract_not_well_formed)
 requires acc(&x.val, 1/0) && x.val == 42
```
ArquintL, line 13: ah here's a division by zero. Is this well-formedness error coming from Silicon? I don't understand why it now behaves differently...

ArquintL (approved):
Overall looks good to me. I didn't look too much at the test cases and left a few minor comments

src/main/scala/viper/gobra/frontend/Desugar.scala
jcp19: If either `left` or `right` are of type `PermissionT`, the type-checker/inference assigns the type `PermissionT` to the `PAdd` expression (assuming that the expression is well-typed). As such, those cases will be handled in the early return statement, before the match block.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
jcp19: out of distraction 😅

src/test/resources/regressions/features/fractional_permissions/fields/fail3.gobra
jcp19: Yes, this error comes from silicon. I believe that the reason why it did not work before is that the error was not back-translated to an error at the Gobra level.

src/main/scala/viper/gobra/frontend/Parser.scala
jcp19: I did not discuss this with any one. However, because we can have `perm` expressions in any place where a "regular" expression may occur, this gives rise to ambiguities when we use `none` for representing permission `0`, e.g.
```go
whatAmI := none
```
In this case, the type of `whatAmI` is ambiguous: it can either be `option[???]` or `perm`. Instead of complicating the type system to infer a default type when it cannot be obtained from the context, I opted to change the keyword from `none` to `noPerm`. The change from `write` to `writePerm` was done for consistency.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -719,21 +719,18 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
-                    case PCompositeLit(typ, _) => typ match {
-                      case PSequenceType(elem) => Some(typeSymbType(elem))
-                      case PSetType(elem) => Some(typeSymbType(elem))
-                      case PMultisetType(elem) => Some(typeSymbType(elem))
-                      case PSliceType(elem) => Some(typeSymbType(elem))
-                      case PArrayType(_, elem) => Some(typeSymbType(elem))
-                      case _ => None // conservative choice
-                    }
-                    case _ => None
+                case tree.parent(litValue: PLiteralValue) => litValue match {
```
ArquintL, line 726: I was tempted to say something about these casts but thought it would lead to too many nestings. Great that you've addressed them as well

jcp19: Yes, your commented maid me realize that I overlooked the fit and finish in this section of code

src/main/scala/viper/gobra/translator/encodings/PermissionEncoding.scala
```diff
@@ -44,8 +44,20 @@ class PermissionEncoding extends LeafTypeEncoding {
       case (e: in.DfltVal) :: ctx.Perm() / Exclusive => unit(withSrc(vpr.NoPerm(), e))
       case fp: in.FullPerm => unit(withSrc(vpr.FullPerm(), fp))
       case np: in.NoPerm => unit(withSrc(vpr.NoPerm(), np))
-      case fp@ in.FractionalPerm(l, r) => for {vl <- goE(l); vr <- goE(r)} yield withSrc(vpr.FractionalPerm(vl, vr), fp)
       case wp: in.WildcardPerm => unit(withSrc(vpr.WildcardPerm(), wp))
+      case co@ in.Conversion(_, exp) => for { e <- goE(exp) } yield withSrc(Function.untupled(e.withMeta), co)
```
Felalolf, line 48: I am bit confused. Is this the general conversion node?
If yes, then I am surprised that tests using conversions currently work. The FinalTypeEncoding should check that nodes are supported at most once, but this case seems to match on all conversion nodes, not just the conversion that result in an expression of type perm.

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -2491,15 +2542,38 @@ object Desugar {
...
+      exp match {
+        case n: PInvoke => Some(info.resolve(n) match {
+          case Some(_: ap.Conversion) =>
+            for {
+              // the well-definedness checker ensures that there is exactly one argument
+              arg <- permissionD(ctx)(n.args.head)
+            } yield in.Conversion(in.PermissionT(Addressability.conversionResult), arg)(src)
+          // this case cannot be None to avoid cycles when desugaring function calls that return permissions
+          case _ => invokeD(ctx)(exp)
+        })
```
Felalolf, line 2562: why is this case needed? Is this not just covered by exprD?

jcp19: this is the leftover of a previous try to remove the ignorePerm argument, before I moved the `if` statement at the beginning of `exprD`. I removed it in the latest commit

src/main/scala/viper/gobra/reporting/VerifierError.scala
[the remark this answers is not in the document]
jcp19: Done: issue #231

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -2491,15 +2542,37 @@ object Desugar {
...
-        case PWildcardPerm() => unit(in.WildcardPerm(src))
+      exp match {
+        case n: PInvoke => info.resolve(n) match {
+          case Some(_: ap.Conversion) =>
+            Some(for {
+              // the well-definedness checker ensures that there is exactly one argument
+              arg <- permissionD(ctx)(n.args.head)
+            } yield in.Conversion(in.PermissionT(Addressability.conversionResult), arg)(src))
+          case _ => None
+        }
```
Felalolf, line 2561: I also meant that I do not understand why this case is required at all?

Felalolf: could you try removing this case and then trying the case, I think it would still work.

jcp19: I actually tried it without the special case and it crashed
```
[error] Exception in thread "main" java.util.concurrent.ExecutionException: java.util.concurrent.ExecutionException: java.lang.AssertionError: assertion failed: Expected both operands to be of the same sort, but found Perm (Z) and Int (1 / 2).
[error] 	at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122)
[error] 	at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:191)
[error] 	at viper.silicon.Silicon.verify(Silicon.scala:199)
[error] 	at viper.silicon.Silicon.verify(Silicon.scala:157)
[error] 	at viper.gobra.backend.Silicon.$anonfun$verify$1(Silicon.scala:27)
[error] 	at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:672)
[error] 	at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:431)
[error] 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
[error] 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
[error] 	at java.base/java.lang.Thread.run(Thread.java:834)
[error] Caused by: java.util.concurrent.ExecutionException: java.lang.AssertionError: assertion failed: Expected both operands to be of the same sort, but found Perm (Z) and Int (1 / 2).
[error] 	at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122)
[error] 	at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:191)
[error] 	at viper.silicon.verifier.DefaultMasterVerifier.$anonfun$verify$9(DefaultMasterVerifier.scala:240)
[error] 	at scala.collection.immutable.List.flatMap(List.scala:293)
[error] 	at scala.collection.immutable.List.flatMap(List.scala:79)
[error] 	at viper.silicon.verifier.DefaultMasterVerifier.verify(DefaultMasterVerifier.scala:240)
[error] 	at viper.silicon.Silicon.viper$silicon$Silicon$$runVerifier(Silicon.scala:235)
[error] 	at viper.silicon.Silicon$$anon$1.call(Silicon.scala:193)
[error] 	at viper.silicon.Silicon$$anon$1.call(Silicon.scala:192)
[error] 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[error] 	... 3 more
[error] Caused by: java.lang.AssertionError: assertion failed: Expected both operands to be of the same sort, but found Perm (Z) and Int (1 / 2).
[error] 	at scala.Predef$.assert(Predef.scala:280)
```

Felalolf (approved):

Issue #190, opened 2021-01-23
Gobra crashes with default values of imported struct types


jcp19:
If we have two files `fileA.gobra` and `fileB.gobra` in packages `A` and `B`, and `fileA.gobra` imports `B` and uses the default value of a struct type `T` defined in `B`, e.g.
- `fileA.gobra`
```go
package A

import "B"

requires t != nil
requires *t == B.T{}
func nonInitialized(t *B.T)
```

- `fileB.gobra`
```go
package B

type T struct {
    t1 int
    t2 int32
}
```

then Gobra crashes with the following trace:
```
[error] Exception in thread "main" org.bitbucket.inkytonik.kiama.relation.NodeNotInTreeException: node not in tree: int32
[error] 	at org.bitbucket.inkytonik.kiama.relation.TreeRelation.apply(TreeRelation.scala:53)
[error] 	at viper.gobra.frontend.info.implementation.typing.BaseTyping.children(BaseTyping.scala:50)
[error] 	at viper.gobra.frontend.info.implementation.typing.BaseTyping.children$(BaseTyping.scala:49)
[error] 	at viper.gobra.frontend.info.implementation.TypeInfoImpl.children(TypeInfoImpl.scala:22)
[error] 	at viper.gobra.frontend.info.implementation.typing.BaseTyping.childrenWellDefined(BaseTyping.scala:52)
[error] 	at viper.gobra.frontend.info.implementation.typing.BaseTyping.childrenWellDefined$(BaseTyping.scala:52)
[error] 	at viper.gobra.frontend.info.implementation.TypeInfoImpl.childrenWellDefined(TypeInfoImpl.scala:22)
[error] 	at viper.gobra.frontend.info.implementation.typing.BaseTyping$$anon$1.safe(BaseTyping.scala:68)
[error] 	at viper.gobra.frontend.info.implementation.typing.BaseTyping$$anon$1.safe(BaseTyping.scala:66)
[error] 	at viper.gobra.util.Safety.apply(Computation.scala:31)
[error] 	at viper.gobra.util.Safety.apply$(Computation.scala:31)
```

PR #230, opened 2021-03-08
Add more stubs

jcp19:
This PR adds stubs for some files in packages `encoding/binary`, `net`, `strconv`, `strings` and `time`. The definitions in these files are required for verifying the SCION dataplane.

When deciding what to include in the stubs, I tried to follow these guidelines:
- I opted to include the headers for every exported member in the files, even if those members are severely under-specified. Sometimes, this was not possible (e.g. higher order functions are omitted/commented because there is no support for the function type). 
- I included the definition of non-exported packages in cases where they are useful to verify parts of the implementation of other exported members
- I included the full uncommented body of members whenever they could be fully verified. If they could not for some "small reason", I would still include the commented body and an annotation starting with `(joao)` stating the reason why it fails.
- Most often, I include comments from the original code to serve as documentation but I made them shorter. For a full description, one should look at the original sources.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/implementation/property/Convertibility.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/property/Convertibility.scala
@@ -6,7 +6,7 @@
 
 package viper.gobra.frontend.info.implementation.property
 
-import viper.gobra.frontend.info.base.Type.{DeclaredT, IntT, PointerT, Single, Type}
+import viper.gobra.frontend.info.base.Type.{DeclaredT, IntT, PointerT, Single, SliceT, StringT, Type}
 import viper.gobra.frontend.info.implementation.TypeInfoImpl
 
 trait Convertibility extends BaseProperty { this: TypeInfoImpl =>
@@ -19,6 +19,7 @@ trait Convertibility extends BaseProperty { this: TypeInfoImpl =>
     case (Single(lst), Single(rst)) => (lst, rst) match {
       case (left, right) if assignableTo(left, right) => true
       case (IntT(_), IntT(_)) => true
+      case (SliceT(IntT(config.typeBounds.Byte)), StringT) => true
       case (left, right) => (underlyingType(left), underlyingType(right)) match {
         case (l, r) if identicalTypes(l, r) => true
         case (IntT(_), IntT(_)) => true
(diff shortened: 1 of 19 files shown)


jcp19:
Test `stubs1.gobra` likely fails for the same reason as issue #190 (it also produces a `NodeNotInTreeException`)

src/main/scala/viper/gobra/frontend/info/implementation/property/Convertibility.scala
```diff
@@ -19,9 +19,11 @@ trait Convertibility extends BaseProperty { this: TypeInfoImpl =>
     case (Single(lst), Single(rst)) => (lst, rst) match {
       case (left, right) if assignableTo(left, right) => true
       case (IntT(_), IntT(_)) => true
+      case (SliceT(IntT(config.typeBounds.Byte)), StringT) => true
```
Felalolf, line 22: Is this case supported by the encoding? If yes, could you write a testcase that uses this case.

Felalolf (commented):
I will ot look at all the stubs for this PR. Over the next weeks, I will go thorugh them step by step and file issues if I see problems.

Felalolf (approved):

src/main/scala/viper/gobra/frontend/info/implementation/property/Convertibility.scala
jcp19: I reported this in issue #241

ArquintL: Does it hold in general that a slice of bytes can be assigned to a string (even without explicitly using `string(...)`)?
If not, it seems to me that it would make more sense to relax the inputs to the `string(...)` function

ArquintL (approved):
Sorry that it took me so long. I did not look in detail at all the stubs but it's definitely an impressive amount of stubs. I've one minor question

src/main/scala/viper/gobra/frontend/info/implementation/property/Convertibility.scala
jcp19: In Go, a []byte can be converted to a string using `string(...)`. Without the explicit conversion, it leads to a type error. Gobra follows the same approach here.

PR #233, opened 2021-03-10
Fix issue 232

Felalolf:
Fixes the issue (twice)


src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1826,16 +1826,17 @@ object Desugar {
     def registerInterface(t: Type.InterfaceT, dT: in.InterfaceT): Unit = {
       Violation.violation(t.decl.embedded.isEmpty, "embeddings in interfaces are currently not supported")
 
-      if (!registeredInterfaces.contains(dT.name)) {
+      if (!registeredInterfaces.contains(dT.name) && info == t.context.getTypeInfo) {
```
jcp19, line 1829: I assume that this is the core of the fix. In a sense, you only register the interface in the package where it is defined instead of the first desugared package that has a value of the interface type in question right?

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1826,16 +1826,17 @@ object Desugar {
...
         registeredInterfaces += dT.name
 
         val itfT = dT.withAddressability(Addressability.Exclusive)
+        val xInfo = t.context.getTypeInfo
```
jcp19, line 1833: Why don't you use `info` instead of declaring `xInfo`? At this point, isn't info always equal to `t.context.getTypeInfo`?

jcp19 (approved):
I have two minor comments but, overall, it looks good to me. Thank you for taking a look at this :)

src/main/scala/viper/gobra/frontend/Desugar.scala
Felalolf: Yes. Before it was translated for every package where it is used. This would also be fine, but it triggered another bug. That is why I made the other changes too. Now it works without this line, but I added it to avoid redundant computations.

src/main/scala/viper/gobra/frontend/Desugar.scala
Felalolf: We should write robust code. Whenever nodes from a different context are desugared, the context has to be adapted. A lot of the bugs that we have right now are cause because this was not done properly. A lot of the design decisions I made for Gobra are done to improve rebustness. In the PRs, I often complain about code that is correct, but not robust, which is a difference. Right now, we have a lot of code that was added quickly and is not robust, increasing the amount of bugs we have greatly.

Felalolf: To add to this, when I write "this is buggy" it usually does not mean that I am certain that there is bug, but that the code is not robust enough and might lead to bugs later on.

ArquintL: I agree that the use of `info` is a common source of bugs (in the type checker but even more often in the desugarer). As the local `info` is the default and often works, it's very easy to forget to correctly use the info of an imported package. However, I currently cannot think of a better architecture to solve this issue

ArquintL (approved):

src/main/scala/viper/gobra/frontend/Desugar.scala
Felalolf: We could just remove the info field and force that the info argument is given as an argument. I might do something like this soonish.

ArquintL: I see that's indeed a changes that should not cause an entire change of the architecture. My other spontaneous idea was to merge all info instances somehow that the desugarer no longer needs to worry which one to use

PR #237, opened 2021-03-14
Add support variadic functions

jcp19:
Add support for variadic functions like specified in the go language spec. There are two decisions I made which are open to discussion:
1) I chose not to add any new node in the internal language representation because this is feature is only syntactic sugar. As such:
    - `func f(s ...int)` is desugared into `func f(s []int)`
    - `f(s...)` is desugared into `f(s)`
    - `f(x1, ..., xn)` is desugared into `f([]T{ x1, ... , xn })`, where `T` is the type of the provided elements

2) I opted to disallow variadic pure functions. The reason for this is that calling such a function may cause a slice to be allocated e.g.
    ```go
    pure f(a ...int)
    func f() {
         f(1, 2, 3) // allocates slice []int{1, 2, 3} here, expression is not pure
    }
    ```


src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
```diff
@@ -23,15 +23,38 @@ trait Assignability extends BaseProperty { this: TypeInfoImpl =>
...
       }
   }
 
+  lazy val variadicAssignableTo: Property[(Vector[Type], Vector[Type])] = createProperty[(Vector[Type], Vector[Type])] {
+    case (right, left) =>
+      StrictAssignModi(left.size, right.size) match {
+        case AssignMode.Variadic => left.lastOption match {
+          case Some(VariadicT(elem)) =>
+            val dummyFill = UnknownType
+            // left.init corresponds to the parameter list on the left except for teh variadic type
```
Felalolf, line 50: spelling "the"

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
```diff
@@ -361,6 +361,8 @@ trait GhostExprTyping extends BaseTyping { this: TypeInfoImpl =>
       // Others
       case PReceive(_) => false
 
+      case PUnpackSlice(s) => isPure(s)(strong) // may change in the future
```
Felalolf, line 364: Is it not just `go(s)`? Unpacking does not change the purity of the expression.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GhostWellDef.scala
```diff
@@ -133,6 +133,8 @@ trait GhostWellDef { this: TypeInfoImpl =>
       "ghost error: make expressions may not contain ghost expressions",
       args exists (x => !noGhostPropagationFromChildren(x))
     )
+
+    case PUnpackSlice(_) => noMessages
```
Felalolf, line 137: I think the case can just be added to the list at line 100.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostMemberTyping.scala
```diff
@@ -71,4 +78,14 @@ trait GhostMemberTyping extends BaseTyping { this: TypeInfoImpl =>
...
   private def isPurePostcondition(spec: PFunctionSpec): Messages = spec.posts flatMap isPureExpr
+
+  private def nonVariadicArguments(args: Vector[PParameter]): Messages = args.flatMap {
+    p: PParameter => error(p, s"Pure members cannot have variadic arguments, but got $p", p.typ.isInstanceOf[PVariadicType])
+  }
+
+  private def wellDefVariadicArgs(args: Vector[PParameter]): Messages =
+    args.dropRight(1).flatMap {
+      p => error(p, s"Only the last argument can be variadic, got $p instead", p.typ.isInstanceOf[PVariadicType])
+    }
```
Felalolf, line 89: Why is this not checked in the member typing (for non ghost)?

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostMemberTyping.scala
```diff
@@ -47,8 +51,11 @@ trait GhostMemberTyping extends BaseTyping { this: TypeInfoImpl =>
     if (member.spec.isPure) {
       isSingleResultArg(member) ++
         isSinglePureReturnExpr(member) ++
-        isPurePostcondition(member.spec)
-    } else noMessages
+        isPurePostcondition(member.spec) ++
+        nonVariadicArguments(member.args)
+    } else {
+      wellDefVariadicArgs(member.args)
```
Felalolf, line 57: I would check this in the GhostMiscTyping

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1560,6 +1602,7 @@ object Desugar {
               }
           }
 
+          case PUnpackSlice(slice) => exprD(ctx)(slice)
```
Felalolf, line 1605: You could change the case to `case PUnpackSlice(slice) => slice` and then define the addressability of PUnpackSlice as the addressability of the argument.

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1051,20 +1051,62 @@ object Desugar {
...
+                // corresponds to the case where an unpacked slice is already passed as an argument
+                res
+              case Some(in.TupleT(_, _)) if len == 1 && parameterCount == 1 =>
+                // supports chaining function calls with variadic functions of one argument
+                val argsMap: Map[BigInt, in.Expr] =
+                  res.last.asInstanceOf[in.Tuple].args.zipWithIndex.map {
+                    case (arg, index) => BigInt(index) -> arg
+                  }.toMap
+                Vector(in.SliceLit(variadicInTyp, argsMap)(src))
+              case _ if len >= parameterCount =>
```
Felalolf, line 1093: I think just using `len >= parameterCount - 1` deals with both cases.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostMemberTyping.scala
```diff
@@ -33,8 +34,11 @@ trait GhostMemberTyping extends BaseTyping { this: TypeInfoImpl =>
...
```
Felalolf, line 40: I would check this in MemberTyping. Also, maybe the isPure function has to be updated for calls.

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1051,20 +1051,62 @@ object Desugar {
...
+                val argsMap: Map[BigInt, in.Expr] =
+                  res.drop(parameterCount-1).zipWithIndex.map {
+                    case (arg, index) => BigInt(index) -> arg
+                  }.toMap
+                res.take(parameterCount-1) :+ in.SliceLit(variadicInTyp, argsMap)(src)
+              case _ if len == parameterCount - 1 =>
+                // variadic argument not passed
+                res :+ in.NilLit(in.SliceT(variadicInTyp, Addressability.nil))(src)
```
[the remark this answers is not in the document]
jcp19: No, this follows the go spec more closely:

> If f is variadic with a final parameter p of type ...T, then within f the type of p is equivalent to type []T. If f is invoked with no actual arguments for p, the value passed to p is nil

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
jcp19: `go(s)` is indeed better, even though they are equivalent

ArquintL (approved):

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -1065,11 +1070,22 @@ object Parser {
...
-    lazy val parameterDecl: Parser[Vector[PParameter]] =
-      ghostParameter |
-      rep1sep(idnDef, ",") ~ typ ^^ { case ids ~ t =>
-        ids map (id => PNamedParameter(id, t.copy).at(id): PParameter)
-      } |  typ ^^ (t => Vector(PUnnamedParameter(t).at(t)))
+    lazy val parameterDecl: Parser[Vector[PParameter]] = {
+      val namedParam = rep1sep(idnDef, ",") ~ "...".? ~ typ ^^ {
+        case ids ~ variadicOpt ~ t =>
+          ids map { id =>
+            val typ = if (variadicOpt.isDefined) PVariadicType(t.copy) else t.copy
```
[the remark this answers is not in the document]
jcp19: Yes. However, I think that it is better to check for this at the type level (the current solution). That way, we can give the user a better error message

src/main/scala/viper/gobra/frontend/Desugar.scala
jcp19: I did not quite get what you mean. Can you elaborate on this?

src/main/scala/viper/gobra/frontend/Desugar.scala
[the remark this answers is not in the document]
Felalolf: wow, okay

src/main/scala/viper/gobra/frontend/Desugar.scala
jcp19: I would still need to do a case split to generate `nil` when `len == parameterCount - 1`. I don't think that I gain anything from merging the two

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostMemberTyping.scala
jcp19: changed in the last commit

jcp19: The reason I did this is that `wellDefActualMember` ends up calling the `wellDefIfPureMethod` and `wellDefIfPureFunction` which are both defined in the `GhostExprTyping.scala`. As such, it just seemed easier to extend the code in that file. I have changed it in the last commit for this particular function

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostMemberTyping.scala
jcp19: | Also, maybe the isPure function has to be updated for calls.

Why? It seems to me that the `isPure` results should be the same, meaning that calls to non-pure methods are not pure and calls to pure methods cannot have variadic arguments. My point here is that the presence of variadic parameters does not affect the purity of a method call. Am I missing anything?

src/main/scala/viper/gobra/frontend/Desugar.scala
Felalolf: My comment was for the case that you can join both cases, which you explained to not be the case.

src/main/scala/viper/gobra/frontend/info/implementation/typing/MemberTyping.scala
```diff
@@ -27,4 +27,14 @@ trait MemberTyping extends BaseTyping { this: TypeInfoImpl =>
       idenNumMsgs ++ constExprMsgs
     case s: PActualStatement => wellDefStmt(s).out
   }
+
+  private[typing] def wellDefBothMember(member: PMember): Messages = member match {
+    case m: PWithParam => wellDefVariadicArgs(m.args)
+    case _ => noMessages
+  }
```
Felalolf, line 34: This seems a bit confusing to me. How do I determine whether I add a check here or in wellDefActualMember? I would just add the additional check to FunctionDecls and MethodDecls.

jcp19: My idea was to have a single method that is applied to both kinds of Members (actual and ghost) in the MemberTyping.scala file. The alternative would be to have these checks in `wellDefActualMember` and in `wellDefGhostMember`, which is the way I had before and which you advised against (because these methods are defined in `GhostMemberTyping`)

Felalolf (approved):

Issue #37, opened 2020-06-19
Disallow use of ghost members in non-ghost code


viper-admin:
> Created by **@arquintl** on 2020-06-19 09:45

I’ve observed that the following code snippet successfully verifies with Gobra. However, in my opinion this should only work if `h` is marked as a ghost variable.

```
package main

type Rectangle struct {
    Width, Height int
}

ghost
requires acc(r.Height)
ensures acc(r.Height)
pure func GetHeight(r *Rectangle) (res int) {
    return r.Height
}

func main() {
    r! := Rectangle{Width: 2, Height: 5}
    h := GetHeight(&r)
    assert h == GetHeight(&r) && h == 5
}
```

‌

Felalolf:
Here is another faulty example from Peter.


[Exercise5.txt](https://github.com/viperproject/gobra/files/5306429/Exercise5.txt)

Felalolf:
Here is another example from Peter.
[anotherExample.txt](https://github.com/viperproject/gobra/files/5306458/anotherExample.txt)

ArquintL:
This issue might have been fixed by #239

Issue #39, opened 2020-07-07
Consistent Support of Old Expressions


viper-admin:
> Created by **@Felalolf** on 2020-07-07 16:10
> Last updated on 2020-07-09 12:49

Currently, the support of old has some flaws. Given labeled old expressions, the following Snippet verifies correctly:

```
package main

func test() {

  x1  := 5
  x2! := 5

  ghost L: // I added the label in the Viper file manually.

  x1 = 8
  x2 = 8

  assert old[L](x1) == 8 && old[L](x2) == 5
}
```

We should unify the behaviour of old on variables. I propose that we treat variables as entities that are affected by old, i.e. the labeled old expressions should evaluate to 5.

We can introduce further the concept of logical variables \(e.g. reusing ghost variables\) that have the same behaviour as variables in Viper.

Furthermore, we have to add support for labeled statements and labeled old expressions as shown above.

Felalolf:
Fixed the issue in PR #239

PR #239, opened 2021-03-17
Labels and labeled old expressions

Felalolf:
Adds labels and labeled old expressions.
In Go, lables have a different namespace than variables. Therefore, I added an additional name resolution.
For labeled old expressions to work, I also had to change how variables are initialized. We now have an explicit initialization statement. Furthermor, I changed how blocks are handled in the desugarer.


src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -734,11 +729,33 @@ object Desugar {
...
       val src: Meta = meta(stmt)
 
-      // Generates a fresh variable if `idn` is a wildcard or returns the variable
-      // associated with idn otherwise
-      def getVar(idn: PIdnNode)(t: in.Type): in.AssignableVar = idn match {
-        case _: PWildcard => freshExclusiveVar(t.withAddressability(Addressability.Exclusive))(src)
-        case x => assignableVarD(ctx)(x)
+      /**
+        * Desugars the left side of an assignment.
+        * If the argument is a variable use, Left is returned, containing the desugared variable.
```
ArquintL, line 734: This function does not return `Either`

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -459,11 +459,13 @@ object Parser {
...
         block |
         simpleStmt |
+        labeledStmt |
+        expressionStmt |
         emptyStmt
 
 
     lazy val simpleStmt: Parser[PSimpleStmt] =
-      sendStmt | assignmentWithOp | assignment | shortVarDecl | expressionStmt
+      sendStmt | assignmentWithOp | assignment | shortVarDecl // expressionStmt is parsed separately
```
ArquintL, line 468: was this change necessary due to trying to parse `labelDef` as an expression?

src/main/scala/viper/gobra/translator/Names.scala
```diff
@@ -10,6 +10,7 @@ import viper.silver.{ast => vpr}
 
 object Names {
   def returnLabel: String = "returnLabel"
+  def labeledVar(base: String, label: String): String = s"${base}_${label}L"
```
ArquintL, line 13: what is a labeled variable? resp. for what is it used?

ArquintL (approved):
Looks good, I only have a few smaller remarks

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -734,11 +729,33 @@ object Desugar {
...
+              _ <- declare(v)
+              _ <- write(in.Initialization(v)(src))
+            } yield v
+
+          case unk: PIdnUnk if info.isDef(unk) =>
+            val v = assignableVarD(ctx)(idn).asInstanceOf[in.LocalVar]
+            for {
+              _ <- declare(v)
+              _ <- write(in.Initialization(v)(src))
+            } yield v
```
jcp19, line 752: These cases have the same body, maybe you can merge them

src/main/scala/viper/gobra/frontend/info/base/SymbolTable.scala
```diff
@@ -245,4 +238,12 @@ object SymbolTable extends Environments[Entity] {
...
   case class BuiltInMPredicate(tag: BuiltInMPredicateTag, rep: PNode, context: ExternalTypeInfo) extends BuiltInGhostEntity with BuiltInMethodLike with GhostTypeMember
 
+
+
+
+  /**
+    * Label
+    */
+
+  case class Label(decl: PLabeledStmt, ghost: Boolean) extends Entity with Product
```
jcp19, line 248: What's the purpose of having a `ghost` flag here? Shouldn't a Label be always non-ghost, even if the labeled statement is ghost?

jcp19: I think I see why now. Is it because a label may occur inside a ghost function or block, in which case you want to ensure that regular control flow does not jump to a ghost statement?

jcp19 (approved):

jcp19:
Fixes #37 and #39

jcp19:
For some reason, I cannot link issues #37 and #39 so that they are closed automatically when we merge this PR.

src/main/scala/viper/gobra/frontend/Parser.scala
Felalolf: yes

src/main/scala/viper/gobra/frontend/info/base/SymbolTable.scala
Felalolf: yes

jcp19:
This is fine by me, you can merge it

PR #242, opened 2021-03-21
Fix issues with contracts in stubs not detected in tests

jcp19:
I realized that I introduced some errors in a late commit to the `add-more-stubs` branch. More precisely, the pre-conditions of methods `GobEncode`, `MarshalJSON`, and `MarshalText` in the file `time.gobra` mention the return values. This PR fixes that.

### DIFF
--- a/src/main/resources/stubs/time/time.gobra
+++ b/src/main/resources/stubs/time/time.gobra
@@ -238,7 +238,6 @@ ensures forall i int :: 0 <= i && i < len(data) ==> acc(&data[i], p)
 func (t *Time) UnmarshalBinary(data []byte, ghost p perm) error
 
 // GobEncode implements the gob.GobEncoder interface.
-requires forall i int :: 0 <= i && i < len(res) ==> acc(&res[i])
 ensures forall i int :: 0 <= i && i < len(res) ==> acc(&res[i])
 func (t Time) GobEncode() (res []byte, error)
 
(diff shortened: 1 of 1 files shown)


ArquintL (approved):

src/main/resources/stubs/time/time.gobra
```diff
@@ -238,7 +238,6 @@ ensures forall i int :: 0 <= i && i < len(data) ==> acc(&data[i], p)
 func (t *Time) UnmarshalBinary(data []byte, ghost p perm) error
 
 // GobEncode implements the gob.GobEncoder interface.
-requires forall i int :: 0 <= i && i < len(res) ==> acc(&res[i])
 ensures forall i int :: 0 <= i && i < len(res) ==> acc(&res[i])
 func (t Time) GobEncode() (res []byte, error)
```
ArquintL, line 242: Out of curiosity: why are some methods defined on `Time` and others on `*Time`?

jcp19: I believe that is a common idiom in Go. When you want to access the fields of a receiver `t` but you don't want to modify them, you use `Time` and this creates a copy of the structure, otherwise you pass `t` as a `*Time`.

ArquintL: True, this is a nice way to express immutability, but is Go clever enough to not copy the struct at runtime? Otherwise, it seems to be quite a high price just to express that something is immutable

jcp19: I don't know about that :/ However, I do not think we should be concerned about it. In fact, this is the way that it is implemented in the Gobra std library

jcp19: I would however assume that Go indeed makes a copy of the structure, at least when you modify the fields of the struct (for some reason, you may want to modify the argument in the scope of your method without modifying the original value that was passed)

PR #248, opened 2021-04-13
Add maps

jcp19:
This PR introduces maps `map` and mathematical maps `mmath` (corresponding to the viper maps) to Gobra. Furthermore, it fixes some minor issues I found during my work on Verified SCION.

Besides the map syntax from Go, maps in Gobra have two ghost operations, represented by the keywords `keySet` and `valueSet`. They are useful to express properties about the content of maps at the specification level. There is also the update functionality for `mmath`s, with the syntax `m[x = y]`.

It should be stressed that, unlike slices, maps are not thread-safe, i.e. do not support concurrent modification. As such, reading the length of a map requires read permissions to the map. This has an effect on the design of permission models for maps: modifying a map requires permission to the whole map, not to `&map[key]` (once again, unlike slices)

**EDIT**: at the moment, maps do not support keys with ghost fields (this is checked in the codebase). I will add them in a separate PR, after ghost fields are added to the concrete AST, so that I can test the solution. I opened a PR so that I do not forget about it ([link](https://github.com/viperproject/gobra/issues/278#issue-875378002))


src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -1118,6 +1142,26 @@ case class SliceT(elems : Type, addressability: Addressability) extends PrettyTy
...
+  * The (composite) type of maps from type `keys` to type `values`.
+  */
+case class MapT(keys: Type, values: Type, addressability: Addressability) extends Type {
+  def hasGhostField(k: Type): Boolean = k match {
+    case StructT(_, fields, _) => fields exists (_.ghost)
+    case _ => false
+  }
+  // this check must be done here instead of at the type system level because the concrete AST does not support
+  // ghost fields yet
+  require(!hasGhostField(keys))
```
Felalolf, line 1155:
```suggestion
  require(!hasGhostField(keys))
  
```

src/main/scala/viper/gobra/ast/internal/theory/TypeHead.scala
```diff
@@ -36,6 +37,7 @@ object TypeHead {
   case object SeqHD extends TypeHead
   case object SetHD extends TypeHead
   case object MSetHD extends TypeHead
+  case object MMapHD extends TypeHead
```
Felalolf, line 40: I think that `MMapHD` is a bad name. I would use `MathMapHD` or something similar.

src/main/scala/viper/gobra/ast/internal/theory/TypeHead.scala
```diff
@@ -108,10 +114,12 @@ object TypeHead {
     case t: StructTExpr => t.fields.map(_._2)
     case t: ArrayTExpr => Vector(t.elems)
     case t: SliceTExpr => Vector(t.elems)
+    case t: MapTExpr => Vector(t.elems)
     case _: PermTExpr => Vector.empty
     case t: SequenceTExpr => Vector(t.elems)
     case t: SetTExpr => Vector(t.elems)
     case t: MultisetTExpr => Vector(t.elems)
+    case t: MathMapTExpr => Vector(t.elems)
```
Felalolf, line 122: These two are wrong. There should be two children.

src/main/scala/viper/gobra/ast/internal/utility/GobraStrategy.scala
```diff
@@ -128,7 +128,7 @@ object GobraStrategy {
       case (_: Capacity, Seq(exp: Expr)) => Capacity(exp)(meta)
       case (_: RangeSequence, Seq(low: Expr, high: Expr)) => RangeSequence(low, high)(meta)
       case (_: SequenceAppend, Seq(l: Expr, r: Expr)) => SequenceAppend(l, r)(meta)
-      case (_: SequenceUpdate, Seq(base: Expr, left: Expr, right: Expr)) => SequenceUpdate(base, left, right)(meta)
+      case (_: GhostCollectionUpdate, Seq(base: Expr, left: Expr, right: Expr)) => GhostCollectionUpdate(base, left, right)(meta)
```
Felalolf, line 131: I think a lot of nodes are missing.

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -968,6 +976,9 @@ object Parser {
     lazy val multisetType : Parser[PMultisetType] =
       "mset" ~> ("[" ~> typ <~ "]") ^^ PMultisetType
 
+    lazy val mmapType: Parser[PMathematicalMapType] =
+      ("mmap" ~> ("[" ~> typ <~ "]")) ~ typ ^^ PMathematicalMapType
```
Felalolf, line 980: I do not like the name mmap

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -748,6 +750,12 @@ object Parser {
     lazy val cap : Parser[PCapacity] =
       "cap" ~> ("(" ~> expression <~ ")") ^^ PCapacity
 
+    lazy val keys : Parser[PMapKeys] =
+      "keySet" ~> ("(" ~> expression <~ ")") ^^ PMapKeys
+
+    lazy val values : Parser[PMapValues] =
+      "valueSet" ~> ("(" ~> expression <~ ")") ^^ PMapValues
+
```
Felalolf, line 758: Why did you pick these names and not the names used by e.g. Viper?

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
```diff
@@ -273,4 +282,16 @@ trait Assignability extends BaseProperty { this: TypeInfoImpl =>
...
   }
+
+  private def isDefinedType(t: Type): Boolean = {
+    // All of the following are defined types (https://golang.org/ref/spec#Predeclared_identifiers):
+    //   bool byte complex64 complex128 error float32 float64
+    //   int int8 int16 int32 int64 rune string
+    //   uint uint8 uint16 uint32 uint64 uintptr
+    t match {
+      // should be extended as new types are added to the language
+      case IntT(_) | BooleanT | DeclaredT(_, _) | StringT => true
```
Felalolf, line 293:
```suggestion
      case _: IntT | BooleanT | _: DeclaredT | StringT => true
```

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
```diff
@@ -273,4 +282,16 @@ trait Assignability extends BaseProperty { this: TypeInfoImpl =>
       case (_, i) => BigInt(i)
     }
   }
+
+  private def isDefinedType(t: Type): Boolean = {
```
Felalolf, line 286: This  function should be moved to UnderlyingType.scala

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -378,9 +381,13 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
                 error(m, s"len larger than cap in make($typ)", maybeLen.isDefined && maybeCap.isDefined && maybeLen.get > maybeCap.get)
             }
 
-        case _: PChannelType | _: PMapType =>
+        case _: PChannelType =>
           error(m, s"too many arguments passed to make($typ)", args.length > 1)
 
+        case PMapType(k, _) =>
+          error(m, s"too many arguments passed to make($typ)", args.length > 1) ++
+            error(m, s"key type $k is not comparable", !comparableType(symbType(k)))
```
Felalolf, line 389:
```suggestion
            error(m, s"key type $k is not comparable", !comparableType(symbType(k))) // TODO: add check that type does not contain ghost
```

Felalolf (commented):
I am not done yet, but here are some comments

src/main/scala/viper/gobra/frontend/Parser.scala
jcp19: What would you suggest instead?

Felalolf: I do not have a goog suggestion.

src/main/scala/viper/gobra/frontend/Parser.scala
jcp19: I was trying to avoid overloading `range`. However, I believe that can be changed to the `domain` and `range` from Viper easily

src/main/scala/viper/gobra/translator/encodings/maps/MapEncoding.scala
```diff
@@ -0,0 +1,361 @@
...
+  private val getMapFunc: vpr.DomainFunc = vpr.DomainFunc(
+    name = getMapFuncName,
+    formalArgs = Seq(vpr.LocalVarDecl("id", vpr.Int)()),
+    typ = vpr.MapType(keyParam, valueParam),
+  )(domainName = domainName)
+
+  /**
+    * Field of the corresponding map id
+    */
+  private def underlyingMapField(ctx: Context): vpr.Field = ctx.field.field(in.IntT(Exclusive))(ctx)
```
Felalolf, line 304: you should just return the field directly, `vpr.Field(someName, vpr.Int)`

src/main/scala/viper/gobra/translator/encodings/maps/MapEncoding.scala
```diff
@@ -0,0 +1,361 @@
...
+            vIdx <- ctx.expr.translate(idx)(ctx)
+            _ <- local(vRes)
+
+            correspondingMapM <- getCorrespondingMap(m, keys, values)(ctx)
+
+            // cannot be replaced by `getCorrespondingMap(vRes, goT(k), goT(v)`, there is no field access to the underlyingMapField
+            correspondingMapRes = vpr.DomainFuncApp(
+              func = getMapFunc,
+              args = Seq(vRes.localVar),
+              typVarMap = Map(keyParam -> goT(keys), valueParam -> goT(values)))(pos, info, errT)
```
Felalolf, line 242:
```suggestion
              typVarMap = Map(keyParam -> goT(keys), valueParam -> goT(values))
            )(pos, info, errT)
```

src/main/scala/viper/gobra/translator/encodings/maps/MapEncoding.scala
```diff
@@ -0,0 +1,361 @@
...
+    }
+  }
+
+  /**
+    * Encodes the allocation of a new map
+    *  [r := make(map[T1]T2, n)] ->
+    *    asserts 0 <= [n]
+    *    var a Ref := new(underlyingMapField)
+    *    inhales len(m) == 0, where m is the underlying map of a
+    *    [r] := a
```
Felalolf, line 137: description of safe map lookup is missing.

src/main/scala/viper/gobra/translator/encodings/maps/MapEncoding.scala
```diff
@@ -0,0 +1,361 @@
...
+        }
+
+        seqn(
+          for {
+            checks <- sequence(runtimeCheck)
+            _ <- if (checks.length == 1) write(checks(0)) else unit(())
+            _ <- errorT {
+              case e@err.ExhaleFailed(Source(info), _, _) if checks.nonEmpty && e.causedBy(checks(0)) =>
+                MapMakePreconditionError(info)
+            }
```
Felalolf, line 161: the error transformer should only be added if checks are acutally happening.

src/main/scala/viper/gobra/translator/encodings/maps/MapEncoding.scala
```diff
@@ -0,0 +1,361 @@
...
+
+            mapVar = in.LocalVar(Names.freshName, t.withAddressability(Exclusive))(makeStmt.info)
+            mapVarVpr = ctx.typeEncoding.variable(ctx)(mapVar)
+            _ <- local(mapVarVpr)
+
+            zeroLit = vpr.IntLit(BigInt(0))(pos, info, errT)
+            correspondingMap <- getCorrespondingMap(mapVar, keys, values)(ctx)
+            _ <- write(vpr.NewStmt(mapVarVpr.localVar, Seq(underlyingMapField(ctx)))(pos, info, errT))
+            _ <- write(vpr.Inhale(vpr.EqCmp(vpr.MapCardinality(correspondingMap)(pos, info, errT), zeroLit)(pos, info, errT))(pos, info, errT))
+            ass <- ctx.typeEncoding.assignment(ctx)(in.Assignee.Var(target), mapVar, makeStmt)
```
Felalolf, line 171: why are you not just assuming that it is the empty map?

src/main/scala/viper/gobra/translator/encodings/maps/MapEncoding.scala
jcp19: I did this to avoid introducing a new name at the viper level. Still, I can do that

src/main/scala/viper/gobra/translator/encodings/maps/MathematicalMapEncoding.scala
```diff
@@ -0,0 +1,118 @@
...
+  /**
+    * Encodes an assignment.
+    * The first and second argument is the left-hand side and right-hand side, respectively.
+    * [(e: mmap[K]V)[idx] = rhs] -> [ e = e[idx := rhs] ]
+    */
+  override def assignment(ctx: Context): (in.Assignee, in.Expr, in.Node) ==> CodeWriter[vpr.Stmt] = default(super.assignment(ctx)){
+    case (in.Assignee(in.IndexedExp(base :: ctx.MathematicalMap(_, _), idx) :: _ / Exclusive), rhs, src) =>
+      for {
+        isCompKey <- MapEncoding.checkKeyComparability(idx)(ctx)
+        _ <- assert(isCompKey, comparabilityErrorT) // key must be comparable
```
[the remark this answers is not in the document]
jcp19: I think it makes sense to be able to only perform lookups if the keys are comparable, even for ghost code. This makes it consistent with the actual maps. Besides, if we remove this check, we may face more difficult challenges in the future, e.g. if we introduce function types and use them as keys of a ghost map

src/main/scala/viper/gobra/translator/encodings/maps/MathematicalMapEncoding.scala
```diff
@@ -0,0 +1,118 @@
...
+              v <- goE(value)
+            } yield vpr.Maplet(k, v)(pos, info, errT)
+          })
+        } yield vpr.ExplicitMap(mapletList)(pos, info, errT)
+
+      case n@ in.IndexedExp(e :: ctx.MathematicalMap(_, _), idx) =>
+        val (pos, info, errT) = n.vprMeta
+        for {
+          isCompKey <- MapEncoding.checkKeyComparability(idx)(ctx)
+          _ <- assert(isCompKey, isCompKey, comparabilityErrorT)(ctx)
```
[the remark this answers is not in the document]
jcp19: Check previous comment

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -492,6 +499,7 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
       case (VariadicT(elem), IntT(_)) => elem
       case (MapT(key, elem), indexT) if assignableTo(indexT, key) =>
         InternalSingleMulti(elem, InternalTupleT(Vector(elem, BooleanT)))
+      case (MathMapT(key, elem), indexT) if assignableTo(indexT, key) => elem
```
ArquintL, line 502: Why is `InternalSingleMulti` not used here analogously to `MapT`? Isn't this required for safe map lookup?

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
```diff
@@ -374,7 +403,7 @@ trait GhostExprTyping extends BaseTyping { this: TypeInfoImpl =>
     res
   }
 
-  private def isPureSeqUpdClause(clause : PSequenceUpdateClause) : Boolean =
+  private def isPureSeqUpdClause(clause : PGhostCollectionUpdateClause) : Boolean =
```
ArquintL, line 406: adapt function name to param type

src/main/scala/viper/gobra/translator/encodings/maps/MapEncoding.scala
```diff
@@ -0,0 +1,364 @@
...
+          for {
+            nVpr <- goE(n)
+            runtimeCheckExp = vpr.LeCmp(vpr.IntLit(0)(pos, info, errT), nVpr)(pos, info, errT)
+          } yield vpr.Exhale(runtimeCheckExp)(pos, info, errT)
+        }
+
+        seqn(
+          for {
+            checks <- sequence(runtimeCheck)
+            _ <- if (checks.length == 1) write(checks(0)) else unit(())
```
ArquintL, line 157: checking that length is 1 seems a bit fragile to me

src/main/scala/viper/gobra/translator/encodings/maps/MapEncoding.scala
```diff
@@ -0,0 +1,364 @@
...
+    lookupExp match {
+      case l@in.IndexedExp(exp :: ctx.Map(keys, values), idx) =>
+        for {
+          vIdx <- goE(idx)
+          isComp <- MapEncoding.checkKeyComparability(idx)(ctx)
+          vDflt <- goE(in.DfltVal(values)(l.info))
+          correspondingMap <- getCorrespondingMap(exp, keys, values)(ctx)
+          containsExp = goMapContains(correspondingMap, vIdx)(l)
+          lookupRes = withSrc(vpr.CondExp(containsExp, withSrc(vpr.MapLookup(correspondingMap, vIdx), l), vDflt), l)
+          lookupResCheckComp <- assert(isComp, lookupRes, comparabilityErrorT)(ctx)
```
ArquintL, line 346: what does this assert do here? Why do we need to check comparability of the key? Can this not be checked statically?

src/main/scala/viper/gobra/translator/encodings/maps/MathematicalMapEncoding.scala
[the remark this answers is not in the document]
ArquintL: Why does `isCompKey` occur twice as argument?

src/test/resources/regressions/features/maps/maps-simple1.gobra
```diff
@@ -0,0 +1,114 @@
...
+requires acc(m, _)
+func test5(m map[int]int) int {
+	return m[0]
+}
+
+func test6() {
+	m := make(map[int]int, 1)
+	assert acc(m) && len(m) == 0
+
+	// multi assignement supported
```
ArquintL, line 38:
```suggestion
	// multi assignment supported
```

src/test/resources/regressions/features/maps/maps-simple1.gobra
```diff
@@ -0,0 +1,114 @@
...
+	v1, ok1 := m[3]
+	assert !ok1 && v1 == 0
+
+	// single assignement supported
```
ArquintL, line 42:
```suggestion
	// single assignment supported
```

ArquintL (approved):
Looks mostly good to me. I had a couple of questions and a few very minor suggestions

src/main/scala/viper/gobra/ast/internal/utility/GobraStrategy.scala
jcp19: For some reason, IntelliJ stopped warning me of inexhaustive matches. Thanks for noticing!

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1783,15 +1794,6 @@ object Desugar {
       }
     }
 
-    def compositeLitToObject(lit : in.CompositeLit) : in.CompositeObject = lit match {
```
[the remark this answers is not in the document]
jcp19: I don't know exactly what was its use before. However, this code seems to not be used/dead.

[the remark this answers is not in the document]
jcp19: PS: I also checked with @Felalolf before removing it

src/main/scala/viper/gobra/translator/encodings/maps/MapEncoding.scala
jcp19: No, at least not fully statically. If the static type of the key is an interface type, this check must be done at runtime

src/main/scala/viper/gobra/translator/encodings/maps/MathematicalMapEncoding.scala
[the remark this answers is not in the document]
jcp19: this is just a hack so that I can fulfill the new signature of `assert`:
```scala
    /* Can be used in expressions. */
    def assert(cond: vpr.Exp, exp: vpr.Exp, reasonT: (Source.Verifier.Info, ErrorReason) => VerificationError)(ctx: Context): Writer[vpr.Exp] = {
      // In the future, this might do something more sophisticated
      val (res, errT) = ctx.condition.assert(cond, exp, reasonT)
      errorT(errT).map(_ => res)
    }
```
that Felix introduced in a previous PR. Introducing a variant of this method that can be used in expressions and that only expects the condition to check and the error transformer introduces ambiguities when calling such method due to ambiguous references to overloaded definition

src/main/scala/viper/gobra/ast/internal/utility/GobraStrategy.scala
jcp19: Should be complete now

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -782,6 +776,33 @@ case class MultisetConversion(expr : Expr)(val info: Source.Parser.Info) extends
...
 
+/* ** Mathematical Map expressions */
+
+/**
+  * Represents a mathematical map literal "mmap[`keys`]`values` { k_0: e_0, ..., k_n: e_n }",
+  * where `keyExprs` constitutes the vector "k_0, ..., k_n" of members of `keys` type
+  * and `valExprs` constitutes the vector "e_0, ..., e_n" of members,
+  * which should all be of type `values`.
+  */
+case class MathMapLit(keys : Type, values : Type, entries : Map[Expr, Expr])(val info : Source.Parser.Info) extends CompositeLit {
```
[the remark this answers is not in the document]
jcp19: You are totally right. I changed it to a sequence of pairs. This observation also made me realize that Go does not define an evaluation order for components of composite literals, which invalidates some assumptions that we make in Gobra (#279).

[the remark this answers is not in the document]
jcp19: I also realized that Go treats repeated keys in maps as an error when they can be evaluated at compile time and I introduced a corresponding check. Furthermore, I introduced a dynamic check that throws an error when the runtime value of two expressions used as keys in a map literal are equal. This way, we circumvent the need to handle undefined behavior.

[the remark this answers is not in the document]
Felalolf: We do not have to check whether constant keys overlap at compile time. We can deal with that at verification time. We do not have to mirror the Go compiler.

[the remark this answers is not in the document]
Felalolf: For instance, in theory, we could do it like Verifast and do the type checks at verification time, as well. What we care about are runtime errors. I.e. if Go throws an exception if a map literal uses the same key twice, then we also have to throw an exception, but if the Go compiler just complain because it likes to complain a lot, then we do not have to care about that in my opinion.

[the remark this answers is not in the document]
jcp19: Sure, that sounds reasonable but it is also good that Gobra fails fast when trivial properties are not met instead of possibly waiting a long time for verification to terminate with an error. I will keep that in mind in the future. Nonetheless, I already added the static checks to Gobra and it seems unecessary to remove them.

src/main/scala/viper/gobra/translator/encodings/maps/MathematicalMapEncoding.scala
```diff
@@ -0,0 +1,123 @@
...
+            assert(checkAllDiffKeys, vpr.ExplicitMap(mapletList)(pos, info, errT), repeatedKeyErrorT)(ctx)
+          } else {
+            unit(vpr.EmptyMap(vprKeyT, vprValT)(pos, info, errT))
+          }
+        } yield mmapVal
+
+      case n@ in.IndexedExp(e :: ctx.MathematicalMap(_, _), idx) =>
+        val (pos, info, errT) = n.vprMeta
+        for {
+          isCompKey <- checkKeyComparability(idx)(ctx)
```
Felalolf, line 96: You are inconsistent with these checks. 

First, I still do not think that keys have to be comparable for mathematical maps. Even if we compare functions, there is always an identity at the verification level. For instance, sets also do not require that the elemnts are comparable. Otherwise, we are unnecessarily restricted in what we can express in Gobra. For instance, we could not have a map from slices to some ghost state, e.g. the number of occurrences of a certain element. 

Second, if you check that keys are comparabile then these checks have to be consistent. You also need to check that the key used in an update expression is comparable and you have to check that the keys used in a literal are comparable. Both of these are missing. (But again, I would just remove the check here).

src/main/scala/viper/gobra/translator/encodings/maps/MapEncoding.scala
```diff
@@ -0,0 +1,396 @@
...
+            mapVar = in.LocalVar(Names.freshName, t.withAddressability(Exclusive))(makeStmt.info)
+            mapVarVpr = ctx.typeEncoding.variable(ctx)(mapVar)
+            _ <- local(mapVarVpr)
+
+            correspondingMap <- getCorrespondingMap(mapVar, keys, values)(ctx)
+            _ <- write(vpr.NewStmt(mapVarVpr.localVar, Seq(underlyingMapField))(pos, info, errT))
+            _ <- write(vpr.Inhale(vpr.EqCmp(correspondingMap, vpr.EmptyMap(goT(keys), goT(values))(pos, info, errT))(pos, info, errT))(pos, info, errT))
+            ass <- ctx.typeEncoding.assignment(ctx)(in.Assignee.Var(target), mapVar, makeStmt)
+          } yield ass
+        )
```
Felalolf, line 194: Do you not check that the keys are distinct and that the keys are comparable?

src/main/scala/viper/gobra/translator/encodings/maps/MapEncoding.scala
```diff
@@ -0,0 +1,396 @@
...
+          correspondingMap <- getCorrespondingMap(res, keys, values)(ctx)
+          // inhale acc(res.underlyingMapField)
+          _ <- write(
+            vpr.Inhale(
+              vpr.FieldAccessPredicate(
+                vpr.FieldAccess(vRes.localVar, underlyingMapField)(pos, info, errT),
+                vpr.FullPerm()(pos, info, errT))(pos, info, errT))(pos, info, errT))
+          // inhale getCorrespondingMap(res) == underlyingMap; recall that underlyingMap == ExplicitMap(mapletList)
+          _ <- write(vpr.Inhale(vpr.EqCmp(underlyingMap, correspondingMap)(pos, info, errT))(pos, info, errT))
+        } yield vRes.localVar
```
Felalolf, line 129: Are you checking that the keys are comparable?

Felalolf (commented):
I mixed up one issue. I did not mean the make statement, but the map literal.

src/test/resources/regressions/features/maps/maps-fail2.gobra
```diff
@@ -0,0 +1,48 @@
...
+	y := 1
+	// we treat map literals with repeated keys as bugs because it
+	// makes use of undefined behaviour in Go
+	//:: ExpectedOutput(assert_error:repeated_map_key_reason)
+	m := map[int]int{1: x, y: 2}
+}
+
+func test6() {
+	//:: ExpectedOutput(assert_error:key_not_comparable_reason)
+	var m1 interface{} = make(map[int]int)
```
Felalolf, line 46:
```suggestion
	var m1 interface{} = make(map[int]int)
	//:: ExpectedOutput(assert_error:key_not_comparable_reason)
```

Felalolf (approved):
Seems good. There is one very small issue as far as I can see.

src/main/scala/viper/gobra/ast/internal/Program.scala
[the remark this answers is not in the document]
ArquintL: It's fine to keep it like that. I second @Felalolf that we could save us some implementation work if we have to do a check at verification time anyways to not implement it statically as well

ArquintL (approved):
Looks good to me. I think there is only one or two open questions by Felix and/or me

PR #250, opened 2021-04-18
Add support for bitwise operators

jcp19:
This PR introduces the following changes:
- Adds support for bitwise operators and extends the constant expression evaluator to handle them. I also added support for integer type converions in the constant evaluator
- Adds support for binary, octal and hexadecimal integer literals
- Changes the cardinality operator for sets from `|exp|` to `len(exp)`. This makes it more consistent with the rest of the language and avoids a parsing ambiguity with the `|` bitwise operator
- Fix some bugs in the type system. A noteworthy change is that Gobra now rejects expressions like `uint(1) + (-1)` which it previously accepted.


src/main/scala/viper/gobra/ast/internal/PrettyPrinter.scala
```diff
@@ -472,7 +473,14 @@ class DefaultPrettyPrinter extends PrettyPrinter with kiama.output.PrettyPrinter
       braces(space <> showExprList(exprs) <> (if (exprs.nonEmpty) space else emptyDoc))
 
   def showLit(l: Lit): Doc = l match {
-    case IntLit(v, _) => v.toString
+    case IntLit(lit, _, base) =>
+      val prefix = base match {
+        case Binary => "0b"
+        case Octal => "0o"
```
ArquintL, line 479: this is inconsistent with the frontend pretty printer

ArquintL:
What is the cardinality operator for multisets now?

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -865,7 +857,7 @@ sealed trait BoolOperation extends Expr {
 }
 
 sealed trait IntOperation extends Expr {
-  override val typ: Type = IntT(Addressability.rValue)
+  override def typ: Type = IntT(Addressability.rValue)
```
ArquintL, line 860: why is this change necessary?

src/main/scala/viper/gobra/ast/internal/utility/Nodes.scala
```diff
@@ -139,6 +138,7 @@ object Nodes {
         case OptionSome(exp) => Seq(exp)
         case OptionGet(exp) => Seq(exp)
         case Negation(operand) => Seq(operand)
+        case BitNeg(operand) => Seq(operand)
```
ArquintL, line 141: `BitNeg` is missing in `GobraStrategy.scala`. What about all other newly introduced internal nodes?

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -703,12 +709,18 @@ object Parser {
       precedence5 ~ ("++" ~> precedence6) ^^ PSequenceAppend |
         precedence5 ~ ("+" ~> precedence6) ^^ PAdd |
         precedence5 ~ ("-" ~> precedence6) ^^ PSub |
+        precedence5 ~ (not("||") ~> "|" ~> precedence6) ^^ PBitOr |
```
ArquintL, line 712: interesting rule, I did not think about this to resolve certain ambiguities!

src/main/scala/viper/gobra/frontend/info/implementation/property/ConstantEvaluation.scala
```diff
@@ -64,7 +66,27 @@ trait ConstantEvaluation { this: TypeInfoImpl =>
...
+      }
+      case PBitNegation(op) =>
+        // Not sufficient to do `intConstantEval(op) map (_.unary_~)`, produces wrong results for unsigned int values
+        exprType(op) match {
+          case IntT(t) =>
+            val constEval = intConstantEval(op)
+            constEval map { constValue =>
+              t match {
+                case UnboundedInteger | _: Signed => ~constValue
+                case u: Unsigned => BigInt(~constValue.toShort) mod (u.upper + 1)
```
ArquintL, line 85: why do you call `toShort`? Doesn't this result in a loss of precision?

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -333,13 +342,60 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
             case _ => assignableTo.errors(l, UNTYPED_INT_CONST)(n) ++ assignableTo.errors(r, UNTYPED_INT_CONST)(n)
           }
           case (_: PAdd, StringT, StringT) => noMessages
-          case (_: PAdd | _: PSub | _: PMul | _: PMod | _: PDiv, l, r) =>
-            if (l == PermissionT || r == PermissionT || getTypeFromCtxt(n.asInstanceOf[PNumExpression]).contains(PermissionT)) {
+          case (_: PAdd | _: PSub | _: PMul | _: PMod | _: PDiv | _: PBitAnd | _: PBitOr | _: PBitXor | _: PBitClear, l, r) =>
+            if (l == PermissionT || r == PermissionT || getTypeFromCtxt(n).contains(PermissionT)) {
```
ArquintL, line 346: Do we effectively allow bitwise operations with permissions on lhs and rhs?

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -333,13 +342,60 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
+              lIsInteger ++ rIsInteger ++ typesAreMergeable ++ exprWithinBounds
             }
+          case (_: PShiftLeft, l, r) =>
+            val integerOperands = assignableTo.errors(l, UNTYPED_INT_CONST)(n) ++ assignableTo.errors(r, UNTYPED_INT_CONST)(n)
+            if (integerOperands.isEmpty) {
+              intConstantEval(n.right.asInstanceOf[PExpression]) match {
+                case Some(v) =>
+                  // The Go compiler checks that the RHS of (<<) is non-negative and, at most, the size
+                  // of the type of the left operand (or 512 if there is an untyped const on the left)
+                  val lowerBound = error(n.right, s"constant ${n.right} overflows uint", v < 0)
```
ArquintL, line 377: Is this indeed called `overflow`? A negative left shift sounds to me more like a right shift with non-negative value

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -333,13 +342,60 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
+                  val nBits = exprOrTypeType(n.left) match {
+                    case IntT(t: BoundedIntegerKind) => t.nbits
+                    case IntT(UnboundedInteger) => MAX_SHIFT
+                    case t => violation(s"unexpected type $t")
+                  }
+                  val upperBound = error(n.right, s"shift count ${n.right} large for type ${exprOrTypeType(n.left)}", v > nBits)
```
ArquintL, line 383:
```suggestion
                  val upperBound = error(n.right, s"shift count ${n.right} too large for type ${exprOrTypeType(n.left)}", v > nBits)
```

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -333,13 +342,60 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
+                case None => noMessages
+              }
+            } else integerOperands
+          case (_: PShiftRight, l, r) =>
+            val integerOperands = assignableTo.errors(l, UNTYPED_INT_CONST)(n) ++ assignableTo.errors(r, UNTYPED_INT_CONST)(n)
+            if (integerOperands.isEmpty) {
+              (intConstantEval(n.right.asInstanceOf[PExpression]) match {
+                case Some(v) =>
+                  // The Go compiler only checks that the RHS of (>>) is non-negative
+                  error(n, s"constant $r overflows uint", v < 0)
```
ArquintL, line 395: same here, I would simply create an error message stating that the constant is negative

src/main/scala/viper/gobra/reporting/VerifierError.scala
```diff
@@ -246,6 +246,11 @@ case class MapMakePreconditionError(info: Source.Verifier.Info) extends Verifica
   override def localMessage: String = s"The provided length to ${info.origin.tag.trim} might be negative"
 }
 
+case class ShiftPreconditionError(info: Source.Verifier.Info) extends VerificationError {
+  override def localId: String = "shift_precondition_error"
+  override def localMessage: String = s"The shift count in ${info.origin.tag.trim} might be negative"
```
ArquintL, line 251: Is `count` a common terminology? I would otherwise prefer `amount`
```suggestion
  override def localMessage: String = s"The shift amount in ${info.origin.tag.trim} might be negative"
```

src/main/scala/viper/gobra/translator/encodings/IntEncoding.scala
```diff
@@ -48,6 +64,103 @@ class IntEncoding extends LeafTypeEncoding {
...
+    col.addMember(bitwiseOr)
+    col.addMember(bitwiseXor)
+    col.addMember(bitClear)
+    col.addMember(shiftLeft)
+    col.addMember(shiftRight)
+    col.addMember(bitwiseNegation)
+  }
+
+  /* Bitwise Operations */
+  private val bitwiseAnd: vpr.Function =
```
ArquintL, line 89: A few comments to this file:
- all bitwise operations are simply uninterpreted function applications, right? Thus, connecting to bit vectors comes at a later point in time?
- you could change your bitwise operations to be `lazy val` and keep track whether they are actually invoked such that only used functions get emitted

ArquintL (approved):
Looks overall good to me. I have a couple of questions and/or remarks

src/main/scala/viper/gobra/translator/encodings/IntEncoding.scala
jcp19: > What is the cardinality operator for multisets now?

It is the same as the regular sets: `len(exp)` where `exp`

src/main/scala/viper/gobra/frontend/info/implementation/resolution/AmbiguityResolution.scala
```diff
@@ -99,14 +100,15 @@ trait AmbiguityResolution { this: TypeInfoImpl =>
 
     case n: PInvoke =>
       exprOrType(n.base) match {
-        case Right(t) => Some(ap.Conversion(t, n.args))
+        case Right(t) if n.args.length == 1 => Some(ap.Conversion(t, n.args.head))
```
Felalolf, line 103: why did you do this change?

src/main/scala/viper/gobra/ast/internal/PrettyPrinter.scala
jcp19: should be fixed in the last commit

src/main/scala/viper/gobra/ast/internal/Program.scala
jcp19: The goal here was to make `typ` in the class `BinaryIntExpr` also a `def`. The reason is that the binary bitwise operators are subtypes of `BinaryIntExpr`. If `typ` is a val, then on the initialization of an instance of `ShiftRight`, the `typ` field would also be initialized. Because the types of both operands cannot be merged, this would trigger a violation. Even when I overrode that field, I got that same violation. Changing it to `def` solved the issue

src/main/scala/viper/gobra/ast/internal/utility/Nodes.scala
jcp19: Good catch, I often forget about these because `Node` is not a sealed trait and thus, the compiler does not complain about missing cases.

src/main/scala/viper/gobra/frontend/info/implementation/property/ConstantEvaluation.scala
jcp19: surprisingly, I did not manage to trigger a case where this caused loss of precision. I have however changed the code to be less fragile in this regard

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
jcp19: This was initially my idea but then I opted to not do it - I don't think it makes sense to have bitwise operators in permissions because they do not have a standard bit representation. However, I forgot to change this.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
jcp19: The Go compiler does indeed report this as an overflow. For example, in this program, 
```go
package main

import (
	"fmt"
)

func main() {
	fmt.Println(1 << -1)
}
```
Go reports the following erros:
```
./prog.go:8:16: constant -1 overflows uint
./prog.go:8:16: invalid negative shift count: -1
```

The reason is that Go expects the right operand to be of uint type and considers that negative right operands overflow uint

src/main/scala/viper/gobra/reporting/VerifierError.scala
jcp19: Count is the default terminology of the Go compiler. For example,
```go
package main

const N = 1 << 513
```
leads to the compiler error
```
./prog.go:3:13: shift count too large: 513
```

src/main/scala/viper/gobra/translator/encodings/IntEncoding.scala
jcp19: That is indeed a good idea

src/main/scala/viper/gobra/frontend/info/implementation/resolution/AmbiguityResolution.scala
jcp19: According to the Go spec,
> An explicit conversion is an expression of the form T(x) where T is a type and x is an expression that can be converted to type T.

In other words, a conversion must have exactly one argument. Before, `ap.Conversion` expected an argument of type `Vector[Expr]` and we had a second check somewhere that it had exactly one value. I don't see a point in having all this complexity. Instead, we can just make invalid instances of ap.Conversion unrepresentable

src/main/scala/viper/gobra/ast/internal/utility/Nodes.scala
ArquintL: I do also always forget them as well. We should either check whether we need this serialization & deserialization at all and then either remove it or write unit tests that consistently trigger them

src/main/scala/viper/gobra/frontend/info/implementation/property/ConstantEvaluation.scala
ArquintL: But why does it has to be a short? Elsewhere in this file, you're for example using `toLong`

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
ArquintL: I see, thanks for the explanation!

src/main/scala/viper/gobra/reporting/VerifierError.scala
ArquintL: Okay, then ignore my comment

src/main/scala/viper/gobra/frontend/info/implementation/property/ConstantEvaluation.scala
jcp19: Long story short, it does not need to be cast actually, `mod` guarantees the correct result without it. However, in my initial version (before using `mod`), I did need a cast to be sure that I produced the correct result. More precisely, I wanted to invert the exact number of bits that the bitwise representation has. I forgot to remove it when I started using `mod`. I removed this complexity in my last commit.

ArquintL (approved):

PR #251, opened 2021-04-18
Fix issue 134

Felalolf:
Fixes how fields are desugared.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Desugar.scala
+++ b/src/main/scala/viper/gobra/frontend/Desugar.scala
@@ -19,6 +19,7 @@ import viper.gobra.translator.Names
 import viper.gobra.util.Violation.violation
 import viper.gobra.util.{DesugarWriter, Violation}
 
+import scala.annotation.unused
 import scala.collection.Iterable
 import scala.reflect.ClassTag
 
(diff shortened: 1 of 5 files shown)


src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -2843,7 +2844,7 @@ object Desugar {
     def variable(n: String, s: PScope, context: ExternalTypeInfo): String = name(VARIABLE_PREFIX)(n, s, context)
     def global  (n: String, context: ExternalTypeInfo): String = nameWithoutScope(GLOBAL_PREFIX)(n, context)
     def typ     (n: String, context: ExternalTypeInfo): String = nameWithoutScope(TYPE_PREFIX)(n, context)
-    def field   (n: String, s: StructT): String = nameWithoutScope(s"$FIELD_PREFIX${struct(s)}")(n, s.context)
+    def field   (n: String, @unused s: StructT): String = s"$n$FIELD_PREFIX" // Field names must preserve their equality from the Go level
```
jcp19, line 2847: Would this pose any problem if two different structs implemented a field with same name but different types?

Felalolf: no, the actual field names are only relevant for the encoding of dynamic types.

jcp19 (approved):

ArquintL (approved):

PR #252, opened 2021-04-18
Fix issue 249

Felalolf:
Adds toString methods to (parse tree) types and symbol table entries.

### DIFF
--- a/src/main/scala/viper/gobra/ast/frontend/Ast.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/Ast.scala
@@ -26,6 +26,7 @@ sealed trait PNode extends Product {
   def pretty(prettyPrinter: PrettyPrinter = PNode.defaultPrettyPrinter): String = prettyPrinter.format(this)
 
   lazy val formatted: String = pretty()
+  lazy val formattedShort: String = pretty(PNode.shortPrettyPrinter)
 
   override def toString: String = formatted
 }
@@ -33,6 +34,7 @@ sealed trait PNode extends Product {
 object PNode {
   type PPkg = String
   val defaultPrettyPrinter = new DefaultPrettyPrinter
+  val shortPrettyPrinter = new ShortPrettyPrinter
 }
 
 sealed trait PScope extends PNode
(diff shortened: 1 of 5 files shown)


src/main/scala/viper/gobra/frontend/info/base/Type.scala
```diff
@@ -84,34 +92,38 @@ object Type {
...
-  case class InternalTupleT(ts: Vector[Type]) extends Type
+  case class InternalTupleT(ts: Vector[Type]) extends PrettyType(s"(${ts.mkString(",")})")
 
   case class InternalSingleMulti(sin: Type, mul: InternalTupleT) extends Type
 
-  case class ImportT(decl: PImport) extends Type
+  case class ImportT(decl: PImport) extends PrettyType(decl.formatted)
 
-  case object SortT extends Type
+  case object SortT extends PrettyType("sort")
```
jcp19, line 108: I believe that `"Type"` is preferable to `"sort"` in order to be consistent with the Go docs (e.g. https://golang.org/src/builtin/builtin.go?s=6524:6544#L160)

jcp19 (approved):

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -26,13 +26,15 @@ sealed trait PNode extends Product {
   def pretty(prettyPrinter: PrettyPrinter = PNode.defaultPrettyPrinter): String = prettyPrinter.format(this)
 
   lazy val formatted: String = pretty()
+  lazy val formattedShort: String = pretty(PNode.shortPrettyPrinter)
```
jcp19, line 29: Do we need a new prettyPrinter? wouldn't the "normal" pretty printer work given that we are only interested in printing the types correctly?

Felalolf: I use the short printer for the symbol table entries

src/main/scala/viper/gobra/frontend/info/base/Type.scala
Felalolf: While I think that "Type" is a bad name, I agree that picking the same name as Go is preferrable.

ArquintL (approved):

PR #254, opened 2021-04-18
Fix issue 253

Felalolf:
What I intended to solve with the `assert` function from the ViperWriter is more complicated than I had originally thought. We would need some kind of sophisticated dependency analysis to not violate any reordering constraints.
For now, I simplified the solution at the price of some performance. The assert function now always generates an additional function.

Additionally, I adapted the equality and string component such that they only generate a domain if necessary.

This PR only fixes issue 253. Somehow issue number 250, 251, 252 were skipped.


src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
```diff
@@ -29,10 +29,15 @@ class StringEncoding extends LeafTypeEncoding {
     * Translates a type into a Viper type.
     */
   override def typ(ctx: Context): in.Type ==> vpr.Type = {
-    case ctx.String() / Exclusive => vpr.Int
+    case ctx.String() / Exclusive => stringType
     case ctx.String() / Shared => vpr.Ref
   }
 
+  private lazy val stringType: vpr.Type = {
```
jcp19, line 36: Good call, I should have done this from the beginning

src/test/resources/regressions/issues/000253.gobra
```diff
@@ -0,0 +1,15 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package issue250
```
jcp19, line 4: nitpick: the package should be `issue253`

src/main/scala/viper/gobra/translator/util/ViperWriter.scala
```diff
@@ -366,33 +339,54 @@ object ViperWriter {
...
 
+    /* Emits Viper statements. */
     def write(stmts: vpr.Stmt*): Writer[Unit] =
       create(stmts.toVector.map(Statement), ())
 
+    /* Emits Viper statements. */
     def assume(cond: vpr.Exp*): Writer[Unit] =
       create(cond.toVector.map(Assumption), ())
 
+    /* Can be used in expressions. */
```
jcp19, line 350: :+1:

jcp19 (approved):

ArquintL:
You do not see issues 250, 251, 252 as GitHub uses the same counter for issues and PRs

ArquintL (approved):

PR #255, opened 2021-04-18
Fix crash of Gobra on every exception

Felalolf:
Added catch to CLI Gobra instance such that sbt does not have to be restartet on every exception.
For the artifact, we should think about what the output on an exception should be exactly.


jcp19 (approved):

ArquintL (approved):
This does not affect Gobra-IDE because Gobra Server directly uses a `Gobra` instance

PR #259, opened 2021-04-20
Example using Shared Variables

ArquintL:
Adds `parallel_search_replace_shared.gobra`, which is a modification of `parallel_search_replace.gobra`. It uses shared instead of exclusive variables in the `worker` function.

- `parallel_search_replace.gobra` takes 60-66sec on my machine
- `parallel_search_replace_shared.gobra` takes 110-121sec on my machine


jcp19 (approved):

jcp19:
@ArquintL I suggest that we either close this PR or merge it. Any preference?

ArquintL:
It's an additional example so I'm more in favor of keeping it

PR #260, opened 2021-04-22
No allocation in ghost code

Felalolf:
I added a ghost slice and adapted the exsting files accordingly.

I created an additional branch because I expected that I will have to adapt test cases, which fortunately was not the case.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GhostWellDef.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GhostWellDef.scala
@@ -95,7 +95,12 @@ trait GhostWellDef { this: TypeInfoImpl =>
 
   private def exprGhostSeparation(expr: PExpression): Messages = expr match {
     case _: PGhostExpression => noMessages
-    case e if enclosingGhostContext(e) => noMessages
+    case e if enclosingGhostContext(e) =>
+      e match {
+        case PMake(_: PGhostSliceType, _) => noMessages
+        case _: PMake | _: PNew | PReference(_: PCompositeLit) => error(e, "Allocating memory within ghost code is forbidden")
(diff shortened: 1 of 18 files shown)


src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GhostWellDef.scala
```diff
@@ -95,7 +95,12 @@ trait GhostWellDef { this: TypeInfoImpl =>
 
   private def exprGhostSeparation(expr: PExpression): Messages = expr match {
     case _: PGhostExpression => noMessages
-    case e if enclosingGhostContext(e) => noMessages
+    case e if enclosingGhostContext(e) =>
+      e match {
+        case PMake(_: PGhostSliceType, _) => noMessages
+        case _: PMake | _: PNew | PReference(_: PCompositeLit) => error(e, "Allocating memory within ghost code is forbidden")
```
jcp19, line 101: in principle, we could allow allocation of ghost memory like ghost slices inside ghost code right? (not that it matters for now)

jcp19 (approved):

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GhostWellDef.scala
Felalolf: that is what is happening at line 100

jcp19: Nvm, that's true

ArquintL (approved):
Looks good to me. Did I understand it correctly that reading a non-ghost slice is fine in ghost code but just the allocation is problematic which is now checked?

Felalolf:
Yes

PR #261, opened 2021-04-22
Interface more checks

Felalolf:
This is not yet completely done, but I would want to merge it in to avoid a PR that is too big later on.

I added a lot of consistency checks. Furthermore, I added that types can be parsed as part of an equality. Lastly, I improved error messages.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Parser.scala
+++ b/src/main/scala/viper/gobra/frontend/Parser.scala
@@ -687,8 +687,8 @@ object Parser {
         precedence4
 
     lazy val precedence4: PackratParser[PExpression] = /* Left-associative */
-      (predeclaredTypeSeparate | precedence4 | typ) ~ ("==" ~> (predeclaredTypeSeparate | precedence4P1 | typ)) ^^ PEquals |
-          (predeclaredTypeSeparate | precedence4 | typ) ~ ("!=" ~> (predeclaredTypeSeparate | precedence4P1 | typ)) ^^ PUnequals |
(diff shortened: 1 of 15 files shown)


src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -1036,6 +1049,26 @@ object Parser {
...
+        exactWord("int64") ~ not("(" | ".") ^^^ PInt64Type() |
+        // unsigned integer types
+        exactWord("byte") ~ not("(" | ".") ^^^ PByte() |
+        exactWord("uint") ~ not("(" | ".") ^^^ PUIntType() |
+        exactWord("uint8") ~ not("(" | ".") ^^^ PUInt8Type() |
+        exactWord("uint16") ~ not("(" | ".") ^^^ PUInt16Type() |
+        exactWord("uint32") ~ not("(" | ".") ^^^ PUInt32Type() |
+        exactWord("uint64") ~ not("(" | ".") ^^^ PUInt64Type() |
+        exactWord("uintptr") ~ not("(" | ".") ^^^ PUIntPtr()
+
```
jcp19, line 1071: wouldn't the following work instead?
```scala
lazy val predeclaredTypeSeparate: Parser[PPredeclaredType] = predeclaredType <~ not("(" | ".")
```

Felalolf: Yes, but I had the impression that my version has a better performance.

ArquintL (approved):

PR #262, opened 2021-04-25
More interface checks

Felalolf:
This is the last bigger pull request before the deadline. 
It adds more checks, improves the error messages, and makes improves our examples.

I test might still fail, but only because of an incorrect ExpectedOutput annotation.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Parser.scala
+++ b/src/main/scala/viper/gobra/frontend/Parser.scala
@@ -1223,13 +1223,23 @@ object Parser {
       }
 
     lazy val implementationProof: Parser[PImplementationProof] =
-      (typ <~ "implements") ~ typ ~ ("{" ~> (methodImplementationProof <~ eos).* <~ "}").? ^^ {
-        case subT ~ superT ~ memberProofOpt => PImplementationProof(subT, superT, memberProofOpt.getOrElse(Vector.empty))
+      (typ <~ "implements") ~ typ ~ ("{" ~> (implementationProofPredicateAlias <~ eos).* ~ (methodImplementationProof <~ eos).* <~ "}").? ^^ {
+        case subT ~ superT ~ Some(predAlias ~ memberProof) =>PImplementationProof(subT, superT, predAlias, memberProof)
+        case subT ~ superT ~ None => PImplementationProof(subT, superT, Vector.empty, Vector.empty)
       }
 
+    lazy val implementationProofPredicateAlias: Parser[PImplementationProofPredicateAlias] =
+      ("pred" ~> idnUse <~ ":=") ~ (selection | namedOperand) ^^ PImplementationProofPredicateAlias
+
     lazy val methodImplementationProof: Parser[PMethodImplementationProof] =
-      "pure".? ~ receiver ~ idnUse ~ signature ~ blockWithBodyParameterInfo.? ^^ {
+      "pure".? ~ nonLocalReceiver ~ idnUse ~ signature ~ blockWithBodyParameterInfo.? ^^ {
         case spec ~ recv ~ name ~ sig ~ body => PMethodImplementationProof(name, recv, sig._1, sig._2, spec.isDefined, body)
+      }
+
+    lazy val nonLocalReceiver: PackratParser[PParameter] =
+      "(" ~> idnDef.? ~ typ <~ ")" ^^ {
+        case None ~ t => PUnnamedParameter(t)
+        case Some(id) ~ t => PNamedParameter(id, t)
       }
 
     lazy val predicateBody: Parser[Option[PExpression]] =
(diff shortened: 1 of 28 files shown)


src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -1223,15 +1223,25 @@ object Parser {
...
     lazy val implementationProof: Parser[PImplementationProof] =
-      (typ <~ "implements") ~ typ ~ ("{" ~> (methodImplementationProof <~ eos).* <~ "}").? ^^ {
-        case subT ~ superT ~ memberProofOpt => PImplementationProof(subT, superT, memberProofOpt.getOrElse(Vector.empty))
+      (typ <~ "implements") ~ typ ~ ("{" ~> (implementationProofPredicateAlias <~ eos).* ~ (methodImplementationProof <~ eos).* <~ "}").? ^^ {
+        case subT ~ superT ~ Some(predAlias ~ memberProof) =>PImplementationProof(subT, superT, predAlias, memberProof)
+        case subT ~ superT ~ None => PImplementationProof(subT, superT, Vector.empty, Vector.empty)
       }
 
+    lazy val implementationProofPredicateAlias: Parser[PImplementationProofPredicateAlias] =
+      ("pred" ~> idnUse <~ ":=") ~ (selection | namedOperand) ^^ PImplementationProofPredicateAlias
```
jcp19, line 1232: this syntax dissonates a bit from the rest of the language because (1) the type annotation occurs before the identifier (2) allows a type annotation in the `:=` syntax, which is not supported otherwise. I do NOT oppose this syntax strongly because this is a very particular case. However, I would suggest instead allowing the short var declaration syntax without the type annotation `perm` (wouldn't we be able to parse this case unambiguously given that every assignment in an implementation proof must be to a perm alias?)

jcp19: PS: `pred` is not exactly a type annotation, but I hope that my point is still valid

src/test/resources/regressions/examples/spec_errors/visitor_pattern.gobra
```diff
@@ -148,8 +137,6 @@ pure func (self *evaluator) visitAddition(a *addition, ghost u int) int {
 
 
 (*evaluator) implements visitor {
-  pure (self *evaluator) visitVariable(v *variable, ghost u int) int { return self.visitVariable(v, u) }
-  pure (self *evaluator) visitConstant(c *constant, ghost u int) int { return self.visitConstant(c, u) }
   //:: ExpectedOutput(precondition_error)
   pure (self *evaluator) visitAddition(c *addition, ghost u int) int { return self.visitAddition(c, u) }
```
jcp19, line 141: Does this new approach support partial implementation proofs? E.g. if the implementation of `visitAddition` was correct, would the current proof inference generate proofs for the methods `visitVariable` and `visitConstant` and use the proof you provided for `visitAddition`?

src/test/resources/regressions/examples/visitor_pattern.gobra
```diff
@@ -97,12 +88,6 @@ pure func (self *childrenCounter) visitAddition(a *addition, ghost u int) int {
   return unfolding acc(a.mem(), 1/u) in (a.left.accept(self, u) + a.right.accept(self, u))
 }
 
-(*childrenCounter) implements visitor {
-  pure (self *childrenCounter) visitVariable(v *variable, ghost u int) int { return self.visitVariable(v, u) }
-  pure (self *childrenCounter) visitConstant(c *constant, ghost u int) int { return self.visitConstant(c, u) }
-  pure (self *childrenCounter) visitAddition(c *addition, ghost u int) int { return self.visitAddition(c, u) }
-}
-
```
jcp19, line 105: Nice! 😁

jcp19 (approved):
I took a brief glance at this. I did not find anything wrong but I have some questions and observations as comments

src/test/resources/regressions/examples/spec_errors/visitor_pattern.gobra
Felalolf: This is not a new approach. I just added that the desugarer actually generates the proof obligations. Yes, it can generated partial proofs.

ArquintL (approved):

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostMemberTyping.scala
```diff
@@ -85,4 +107,80 @@ trait GhostMemberTyping extends BaseTyping { this: TypeInfoImpl =>
...
+    }
+  }
+
+  /**
+    * Depends on which packages are loaded. Only call at the end of type checking.
+    * Either returns a set of errors caused by missing implementation proofs
+    * or a set of implementation proofs that have to be generated.
+    **/
+  def wellImplementationProofs: Either[Messages, Vector[(Type, InterfaceT, MethodImpl, MethodSpec)]] = {
+    if (isMainContext && requiredImplements.nonEmpty) {
```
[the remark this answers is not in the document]
Felalolf: The function checks that all implementation proofs, including the ones from other packages are fine. Therefore, this function is executed only once by the main context.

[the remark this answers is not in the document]
ArquintL: I see thanks!

PR #263, opened 2021-04-26
Relative In-File Config Paths

ArquintL:
Changes behavior of In-File IncludeDir paths to be relative to the file in which they are instead of being relative to the location at which Gobra is executed.

### DIFF
--- a/src/test/scala/viper/gobra/BenchmarkTests.scala
+++ b/src/test/scala/viper/gobra/BenchmarkTests.scala
@@ -54,7 +54,7 @@ trait GobraFrontendForTesting extends Frontend {
   override def init(verifier: Verifier): Unit = () // ignore verifier argument as we reuse the Gobra / Parser / TypeChecker / etc. instances for all tests
 
   override def reset(files: Seq[Path]): Unit =
-    config = Some(Config(inputFiles = files.map(_.toFile).toVector, reporter = NoopReporter))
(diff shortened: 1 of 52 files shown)


src/test/scala/viper/gobra/BenchmarkTests.scala
```diff
@@ -47,14 +47,17 @@ trait BenchmarkTests extends StatisticalTestSuite {
 }
 
 trait GobraFrontendForTesting extends Frontend {
+  val z3PropertyName = "GOBRATESTS_Z3_EXE"
+  val z3Exe: Option[String] = Option(System.getProperty(z3PropertyName))
```
Felalolf, line 51: What is this?

Felalolf (commented):
I also checked it on Windows. I have one question, otherwise everthing looks fine.

src/test/scala/viper/gobra/BenchmarkTests.scala
ArquintL: This reads the `GOBRATESTS_Z3_EXE` property passed with `-DGOBRATESTS_Z3_EXE`

Felalolf (approved):

ArquintL:
I'll wait with merging this branch until @Felalolf branch on interfaces is merged as there are a few more in file configs to be adapted

Felalolf (approved):

PR #267, opened 2021-04-26
More robust error backtranslation.

Felalolf:
I made the error backtranslation more robust and fixed a bug concerning assertions. Before, the body of an assertion could be impure, which is forbidden now.


ArquintL (approved):
Looks good

PR #271, opened 2021-04-27
Fixes Built-In Methods and MPredicates

ArquintL:
Caching of built-in method and mpredicate proxies did not take the receiver into account. Thus, using the same member with different receiver types in the same program lead to wrong proxies and missing encoded members.
In addition, this PR fixes a couple of "not in tree" errors in the desugarer and adds test cases covering all cases in which no longer `info` but the correct context is used in the desugarer.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Desugar.scala
+++ b/src/main/scala/viper/gobra/frontend/Desugar.scala
@@ -243,64 +243,65 @@ object Desugar {
 
     // proxies to built-in members
     def methodProxy(tag: BuiltInMethodTag, recv: in.Type, args: Vector[in.Type])(src: Meta): in.MethodProxy = {
-      val inRecv = recv.withAddressability(Addressability.inParameter)
-      val inArgs = args.map(_.withAddressability(Addressability.inParameter))
-
-      def create(): in.BuiltInMethod = {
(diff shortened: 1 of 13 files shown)


ArquintL:
@MartinClochard This PR fixes your channel example. The example itself is not part of this PR

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -210,97 +210,98 @@ object Desugar {
...
         builtInMembers += (tag, args) -> member
         member
       }
 
-      val filteredMembers: Map[(BuiltInMemberTag, Vector[in.Type]), M] = builtInMembers.collect {
-        case ((t, as), m: M@unchecked) => (t, as) -> m
+      val filteredMembers: Map[(T, Vector[in.Type]), M] = builtInMembers.collect {
+        case ((t: T@unchecked, as), m: M@unchecked) => (t, as) -> m
       }
       filteredMembers.getOrElse((tag, args), genAndStore())
```
Felalolf, line 306: This part looks a bit whack, but due to circumstances it is fine.

Felalolf (approved):

jcp19 (approved):

PR #273, opened 2021-04-28
just the tutorial examples

Felalolf:
I think that we should merge the tutorial examples as soon as possible in order to check the docker image appropriately.
We can either just merge the add-tutorial branch or this one. I do not have a preference. If add-tutorial is merged before this branch, then just close the PR and delete the branch.


Felalolf:
Also, feel free to change the name of the folder

PR #275, opened 2021-04-28
Improve artifact

Felalolf:
We later need to change the GOBRA_COMMIT arg to a tag.


artifact/content/tutorial.md
```diff
@@ -0,0 +1,681 @@
...
+(*counter) implements stream {
+  ...
+}
+```
+
+- Explained in [the section on interfaces](#interfaces)
+
+## Basic Annotations
+
+All examples shown in this tutorial can be found [here](../src/test/resources/regressions/examples/tutorial-examples/). We start with a simple function `sum` computing the sum of the integers from 0 to `n`. The table below shows a Go implementation (left) and a corresponding annotated version (right).
```
jcp19, line 110: now that the `tutorial-examples` is in the master branch, we can put a link to it. If we don't change it for this particular copy of the tutorial, the link will not work
```suggestion
All examples shown in this tutorial can be found [here](https://github.com/viperproject/gobra/tree/master/src/test/resources/regressions/examples/tutorial-examples). We start with a simple function `sum` computing the sum of the integers from 0 to `n`. The table below shows a Go implementation (left) and a corresponding annotated version (right).
```

ArquintL: Should this be eventually changed to the `artifact1` tag as well?

ArquintL (approved):
As this PR contains my commits, I'm obviously approving it ^^

artifact/content/tutorial.md
Felalolf: yes

PR #283, opened 2021-05-13
Variadic assign

Augl:
Add fix to allow variadic arguments to be assigned to slices.


src/test/resources/regressions/issues/000282.gobra
```diff
@@ -0,0 +1,6 @@
+package issue282
```
jcp19, line 1: Our `.gobra` regressions generally must start with the following license, otherwise our CI complains
```
```

jcp19 (approved):

PR #290, opened 2021-05-23
Fix selective import

jcp19:
NOT READY FOR REVIEW


jcp19:
Also, modify `TypeInfoImpl.scala` after merging the PR for the `preserves` keyword to include the nodes refered in preserves clauses.

```scala
private def addTransitiveDeps(r: SymbolTable.Regular): Unit = {
    def checkedNodesFromFuncSpec(spec: PFunctionSpec): Vector[PNode] =
      spec.pres ++ spec.preserves ++ spec.posts ++ (spec.pres ++ spec.preserves ++ spec.posts).flatMap(_.subNodes)
        // TODO: check why this causes node not in tree exception
        // (try { (spec.pres ++ spec.posts) flatMap allChildren } catch { case e => Seq.empty })
```

PR #293, opened 2021-06-06
Add support for type switch statements

### LINKED ISSUE #289: Switch statements for types not implemented
The following leads to a `NotImplementedException`
```go
switch e := e.(type) {
    ...
}
```

jcp19:
Fixes #289

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Desugar.scala
+++ b/src/main/scala/viper/gobra/frontend/Desugar.scala
@@ -1060,58 +1060,61 @@ object Desugar {
       }
     }
 
-    def switchCaseD(switchCase: PExprSwitchCase, scrutinee: in.LocalVar)(ctx: FunctionContext): Writer[(in.Expr, in.Stmt)] =
-      switchCase match {
-        case PExprSwitchCase(left, body) => for {
-          acceptExprs <- sequence(left.map(clause => exprD(ctx)(clause)))
-          acceptCond = acceptExprs.foldRight(in.BoolLit(b = false)(meta(left.head)) : in.Expr){
-            (expr, tail) => in.Or(in.EqCmp(scrutinee, expr)(expr.info), tail)(expr.info)
-          }
-          stmt <- stmtD(ctx)(body)
-        } yield (acceptCond, stmt)
-      }
+    def switchCaseD(switchCase: PExprSwitchCase, scrutinee: in.LocalVar)(ctx: FunctionContext): Writer[(in.Expr, in.Stmt)] = {
+      val left = switchCase.left
+      val body = switchCase.body
+      for {
+        acceptExprs <- sequence(left.map(clause => exprD(ctx)(clause)))
+        acceptCond = acceptExprs.foldRight(in.BoolLit(b = false)(meta(switchCase)): in.Expr){
+          (expr, tail) => in.Or(in.EqCmp(scrutinee, expr)(expr.info), tail)(expr.info)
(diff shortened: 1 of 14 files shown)


src/test/resources/regressions/features/switch/type-switch-simple1.gobra
```diff
@@ -0,0 +1,60 @@
...
+	x := 1
+	var y interface{} = nil 
+	switch y.(type) {
+		case int:
+			k = 2
+		case nil:
+			k = 1 
+	}
+	assert k == 1
+}
```
ArquintL, line 49: Another test case for `default` would be nice:

```suggestion
func h() {
	k := 0
	x := 1
	var y interface{} = nil 
	switch y.(type) {
		case int:
			k = 2
		case nil:
			k = 1 
	}
	assert k == 1
}

func i() {
	k := 0
	x := 1
	var y interface{} = nil 
	switch y.(type) {
		case int:
			k = 2
		case default:
			k = 1 
	}
	assert k == 1
}
```

ArquintL (approved):

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1030,21 +1030,91 @@ object Desugar {
...
+            newCtx
+          case (None, None) => ctx
+          case c => violation(s"This case should be unreachable, but got $c")
+        }
+
+        init = assign.map(v => in.SingleAss(in.Assignee.Var(v), in.TypeAssertion(scrutinee, v.typ)(v.info))(v.info)).toVector
+
+        stmt = blockD(context)(body) match {
+          case in.Block(decls, stmts) => in.Block(assign.toVector ++ decls, init ++ stmts)(meta(body))
+          case c => violation(s"This case should be unreachable, but got $c")
```
Felalolf, line 1114:
```suggestion
          case c => violation(s"Expected Block as result from blockD, but got $c")
```

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1030,21 +1030,91 @@ object Desugar {
...
+    }
+
+    def typeSwitchCaseD(switchCase: PTypeSwitchCase, scrutinee: in.LocalVar, bind: Option[PIdnDef])(ctx: FunctionContext): Writer[(in.Expr, in.Stmt)] = {
+      val left = switchCase.left
+      val body = switchCase.body
+      val metaCase = meta(switchCase)
+      for {
+        acceptExprs <- sequence(left.map {
+          case t: PType =>
+            for { e <- exprAndTypeAsExpr(ctx)(t) } yield in.EqCmp(in.TypeOf(scrutinee)(meta(t)), e)(metaCase)
```
Felalolf, line 1082: What if the type is an interface type?

jcp19: I misunderstood Go's language spec and thought that interface types where could not occur in case clauses. That should be fixed in my last commit.

Felalolf (changes requested):
Right now, the encoding is unsound due to the handling of interfaces.

A cast to an interface must always succeed. This is because we only encode the behavioral subtyping and not the syntactitc subtyping. If our encoding states that some type is not a behavioral subtype, even though Go considers it a subtype, then we have changed the control flow of the program in our encoding.

Felalolf (approved):

PR #300, opened 2021-06-10
Fixes #299

### LINKED ISSUE #299: Reduce restrictions of pure annotation
Right now, pure annotations always have to be written on the same line as the "func" keyword. We should permit the "pure" annotation to be at an arbitrary clause position (or at the beginning of the line of the function declaration).

Felalolf:
The "pure" annotation can now be at several positions in the specification.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Parser.scala
+++ b/src/main/scala/viper/gobra/frontend/Parser.scala
@@ -421,10 +421,29 @@ object Parser {
           PFunctionDecl(name, sig._1, sig._2, spec, body)
       }
 
-    lazy val functionSpec: Parser[PFunctionSpec] =
-      ("requires" ~> expression <~ eos).* ~ ("ensures" ~> expression <~ eos).* ~ "pure".? ^^ {
-        case pres ~ posts ~ isPure => PFunctionSpec(pres, posts, isPure.nonEmpty)
+
+
+    lazy val functionSpec: Parser[PFunctionSpec] = {
+
+      sealed trait FunctionSpecClause
+      case class RequiresClause(exp: PExpression) extends FunctionSpecClause
+      case class EnsuresClause(exp: PExpression) extends FunctionSpecClause
+      case object PureClause extends FunctionSpecClause
+
+      lazy val functSpecClause: Parser[FunctionSpecClause] = {
+        "requires" ~> expression <~ eos ^^ RequiresClause |
+        "ensures" ~> expression <~ eos ^^ EnsuresClause |
+        "pure" <~ eos ^^^ PureClause
       }
+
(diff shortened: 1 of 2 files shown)


src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -421,11 +421,30 @@ object Parser {
...
+      case class EnsuresClause(exp: PExpression) extends FunctionSpecClause
+      case object PureClause extends FunctionSpecClause
+
+      lazy val functSpecClause: Parser[FunctionSpecClause] = {
+        "requires" ~> expression <~ eos ^^ RequiresClause |
+        "ensures" ~> expression <~ eos ^^ EnsuresClause |
+        "pure" <~ eos ^^^ PureClause
       }
 
+      functSpecClause.* ~ "pure".? ^^ {
```
ArquintL, line 439: is it necessary to have here `"pure".?`? Shouldn't this already be covered by another `functSpecClause`?

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -421,11 +421,30 @@ object Parser {
...
+        case clauses ~ pure =>
+          val pres = clauses.collect{ case x: RequiresClause => x.exp }
+          val posts = clauses.collect{ case x: EnsuresClause => x.exp }
+          val isPure = pure.nonEmpty || clauses.contains(PureClause)
+          PFunctionSpec(pres, posts, isPure)
```
ArquintL, line 444: shouldn't you attach positional information to `PFunctionSpec` here?

ArquintL (approved):
Two comments, otherwise looks good

src/main/scala/viper/gobra/frontend/Parser.scala
Felalolf: that is done by the parser framework

src/main/scala/viper/gobra/frontend/Parser.scala
Felalolf: So, yes it is necessary.

jcp19 (approved):
Looks fine to me. The only awkward thing to me is that it allows functions with multiple pure clauses, but we can argue that the `pure` modifier is idempotent.

src/main/scala/viper/gobra/frontend/Parser.scala
ArquintL: I see, then maybe this might be a nicer solution: `repsep(functSpecClause, eos) <~ eos.?`

ArquintL:
@jcp19 maybe you can negative the effect of the pure keyword when having two :P

PR #303, opened 2021-06-17
Improve stub for encoding.binary

jcp19:
This PR uncomments the bodies of some methods in the `encoding.binary` stub - this can now be done because of the recently added support for bitwise operators. Furthermore, I changed the way `BigEndian` and `LittleEndian` are defined, which allows for calling methods defined for types `bigEndian` and `littleEndian` using the same syntax as in regular Go.

### DIFF
--- a/src/main/resources/stubs/encoding/binary/binary.gobra
+++ b/src/main/resources/stubs/encoding/binary/binary.gobra
@@ -73,6 +73,10 @@ type bigEndian int
 	(e littleEndian) PutUint64(b []byte, v uint64) {
 		e.PutUint64(b, v)
 	}
+
+	pure (e littleEndian) String() string {
+		return e.String()
+	}
 }
 
 (bigEndian) implements ByteOrder {
@@ -98,6 +102,10 @@ type bigEndian int
 
 	(e bigEndian) PutUint64(b []byte, v uint64) {
 		e.PutUint64(b, v)
+	}
+
+	pure (e bigEndian) String() string {
+		return e.String()
 	}
 }
 
--- a/src/main/resources/stubs/encoding/binary/binary.gobra
+++ b/src/main/resources/stubs/encoding/binary/binary.gobra
@@ -34,7 +34,7 @@ type ByteOrder interface {
 	ensures acc(&b[4]) && acc(&b[5]) && acc(&b[6]) && acc(&b[7])
 	PutUint64(b []byte, uint64)
 
-	String() string
+	pure String() string
 }
 
 // (joao): Original implementation of BigEndian and LittleEndian. Changed due to lack of support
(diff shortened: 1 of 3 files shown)


src/main/resources/stubs/encoding/binary/binary.gobra
```diff
@@ -44,106 +44,110 @@ type ByteOrder interface {
...
-type LittleEndian struct {
-	// TODO: remove field
-	x int // temporary solution to side-step problems with the encoding of interfaces with fields of finite cardinality
-}
-type BigEndian struct {
-	// TODO: remove field
-	x int // temporary solution to side-step problems with the encoding of interfaces with fields of finite cardinality
-}
+const LittleEndian littleEndian = 0;
+const BigEndian bigEndian = 0;
```
ArquintL, line 48: It is intentional that both have the same value?

src/main/resources/stubs/encoding/binary/binary.gobra
```diff
@@ -44,106 +44,110 @@ type ByteOrder interface {
 // type littleEndian struct{}
 // type bigEndian struct{}
```
ArquintL, line 45: Couldn't this be removed?

src/test/resources/regressions/features/stubs/stubs0.gobra
```diff
@@ -13,10 +13,10 @@ import (
 )
 
 // Checks that files from package "encoding/binary" are correctly parsed
-func testEncodingBinary(e binary.LittleEndian) {
+func testEncodingBinary() {
 	s := make([]byte, 2)
-	e.PutUint16(s, uint16(0))
-	res := e.String()
+	binary.LittleEndian.PutUint16(s, uint16(0))
```
ArquintL, line 18: Ah the constants get used as a receiver and you already get the different semantics just because of the type such that the constants' values do not matter

ArquintL (approved):
One minor question regarding lines 44-45

src/main/resources/stubs/encoding/binary/binary.gobra
jcp19: This is the original definition. I would like to keep this here for when we add support to global variables

src/test/resources/regressions/features/stubs/stubs0.gobra
jcp19: Exactly!

src/main/resources/stubs/encoding/binary/binary.gobra
jcp19: I did pick the same value because in the original implementation, both constants have the same value too (an instance of the empty struct).

src/main/resources/stubs/encoding/binary/binary.gobra
ArquintL: I'm fine with keeping the `var`s. I was more referring to the `type` declarations as you seem to use `int` instead of empty interfaces.
Why don't you use empty interfaces? Is this still because of the problematic encoding of fields of finite cardinality (as mentioned in the changeset below)?

jcp19: No, it is because empty interfaces are not valid types for constants in Go. Empty structs are an acceptable type but initializing a value of struct{} cannot be done statically and thus, the compiler rejects it

ArquintL: I see thanks

PR #304, opened 2021-07-01
Fix permission errors when accessing elements of a nil map

jcp19:
Currently, Gobra requires access to a map in order to perform a lookup. However, that rules out the possibility of performing a lookup in a nil map, which should always be possible (nil maps are treated as empty maps in Go). This PR addresses that and fixes a couple of other issues.


ArquintL (approved):
Looks good!

Issue #228, opened 2021-03-08
Incorrect type checking of PLen 


jcp19:
The following code compiles and runs successfully, printing the value "true"
```go
package main

import "fmt"

type S []int

func main() {
    var s S = S{1,2,3}
    fmt.Println(len(s) == 3)
}
```

Doing something analogous in Gobra, e.g. the following snippet, leads to a type error:
```go
package main

type S []int

func main() {
    var s S = S{1,2,3}
    assert len(s) == 3
}
```

Type error:
```
Gobra has found 1 error(s):
[info] 	<7:16> bug_slice.gobra:7:16:error: expected an array, string, sequence or slice type, but got DeclaredT(type S []int,viper.gobra.frontend.info.implementation.TypeInfoImpl@15993d33)
[info]     assert len(s) == 3;
[info]                ^
```

PR #305, opened 2021-07-01
Fix defined types functions

jcp19:
Currently, operations defined for slices, arrays and maps such as `len` and indexing are not available to elements of a type T whose underlying type is a slice, array or map type. Issue #228 shows such an example. This PR addresses that by making these operations available to defined types with the adequate underlying type.

To make this work in our current architecture, I had to pass extra parameters to constructors of some nodes in the internal language. For example, `in.IndexedExp` now receives as an extra parameter the underlying type of its base expression's type. Although this forces an implicit relation between the parameters that we cannot check, this is required because there is no typing information available to the constructor of `in.IndexedExp` and thus, without it, we could not return the correct expression type via a call to `typ`.


src/test/resources/regressions/issues/000228.gobra
```diff
@@ -0,0 +1,71 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package types
+
+type TSlice []int
```
ArquintL, line 6: maybe you could duplicate this test case but use type aliases instead of type declarations

ArquintL (approved):
Looks good to me, I only have one comment

src/test/resources/regressions/issues/000228.gobra
jcp19: Done! In the process of doing this, I found a bug which makes me doubt that it was ever possible to instantiate a value of an alias type ^^ this should be fixed now

ArquintL (approved):

PR #307, opened 2021-07-04
Improve specification of the net stub

jcp19:
This PR improves introduces a `mem()` predicate  in every interface defined in the `net.gobra` file. Furthermore, it adds two new files (`iprawsock.gobra` and `udpsock.gobra`) with definitions useful to verify scion. In these two files, the signal to noise ratio may be rather small, as I did not bother verifying members which are not immediately useful to my ends on verified scion or which rely on unsupported features of Gobra.


ArquintL (approved):
Looks good

Issue #308, opened 2021-07-04
Cannot define method predicates for non-struct types


jcp19:
The following snippet leads to an error in Gobra:
```go
type T int

func (t T)m() // definition of methods succeeds

pred (t T)n() // definition of method predicates fails
```

Error:
```
Gobra has found 1 error(s):
[info] 	<87:1> workingFile.gobra:87:1:error: property error: got T that is not a class type
[info] pred (t T)n(); 
[info] ^
```

PR #310, opened 2021-07-05
Fix issue 308

jcp19:
Fixes issue #308


ArquintL (approved):
That's a nice changeset!

PR #315, opened 2021-07-09
Implement outline keyword

Augl:
This PR introduces the outline keyword which allows implicit outlining of code into its own function (with specification) on the viper level, which reduces the verification burden.

```
func outline_declare() {
	ensures x == 1 && y == 2
	outline (
		x := 1
		y := 2
	)
	sum := x + y
	assert sum == 3
}
```

PR #317, opened 2021-07-12
VerifiedSCION preview

jcp19:
**This PR is not supposed to be reviewed and merged!** Instead, it keeps track of the version of Gobra that I am using to verify SCION's codebase. I never commit directly to this method. I only use it to merge features that are still experimental or under review.


jcp19:
This branch is now deprecated. It contains only the changes from #315, which is still unstable.

Issue #318, opened 2021-07-16
Kiama cycle when checking interface well-definedness


jcp19:
In the following example, kiama detects a cycle when checking the well-definedness of method `cmp`
```go
type T interface {
	pred Mem()

	requires Mem()
	requires t.Mem() // Gobra runs fine if this line is omitted
	cmp(t T) bool
}
```

This is caused by the precondition `requires t.Mem()` in the specification of `cmp`, which in turn causes Gobra to check the well definedness of the interface type.

Snippet from the stack trace:
```
[info] An unknown Exception was thrown.
[info] Cycle detected in attribute evaluation 'store' at interface {
[info]   requires Mem()
[info]   requires t.Mem()
[info]   cmp(t T) (bool)pred Mem()
[info] }
[info] java.lang.IllegalStateException: Cycle detected in attribute evaluation 'store' at interface {
[info]   requires Mem()
[info]   requires t.Mem()
[info]   cmp(t T) (bool)pred Mem()
[info] }
[info] 	at org.bitbucket.inkytonik.kiama.attribution.Attribute.reportCycle(Attribute.scala:24)
[info] 	at org.bitbucket.inkytonik.kiama.attribution.AttributionCore$CachedAttribute.apply(AttributionCore.scala:54)
[info] 	at viper.gobra.util.Memoization.apply(Computation.scala:23)
[info] 	at viper.gobra.util.Memoization.apply$(Computation.scala:23)
[info] 	at viper.gobra.frontend.info.implementation.typing.BaseTyping$$anon$1.apply(BaseTyping.scala:98)
[info] 	at viper.gobra.util.Validity.valid(Computation.scala:38)
[info] 	at viper.gobra.util.Validity.valid$(Computation.scala:38)
[info] 	at viper.gobra.frontend.info.implementation.typing.BaseTyping$$anon$1.valid(BaseTyping.scala:98)
[info] 	at viper.gobra.frontend.info.implementation.typing.BaseTyping$$anon$4.safe(BaseTyping.scala:138)
[info] 	at viper.gobra.util.Safety.apply(Computation.scala:31)
[info] 	at viper.gobra.util.Safety.apply$(Computation.scala:31)
[info] 	at viper.gobra.frontend.info.implementation.typing.BaseTyping$$anon$4.viper$gobra$util$Memoization$$super$apply(BaseTyping.scala:136)
[info] 	at viper.gobra.util.Memoization.$anonfun$store$1(Computation.scala:21)
```

PR #319, opened 2021-07-16
Fix issue #318

jcp19:
Fixes issue #318 by avoiding the checks for well definedness which occur in an interface definition.


Felalolf (approved):
I literally suggested the lines of code.

PR #322, opened 2021-07-19
Fixes #312

### LINKED ISSUE #312: Implementation Proof with Return Params
The encoding of postconditions uses the Viper variable names for return parameters from the interface's method declaration instead of the implementation proof. Thus, Silicon crashes with a no such element exception as the variable lookup fails.
Neither the generated nor manually provided implementation proof works.

The following program triggers this problem:
```

package pkg

type A struct {}

type I interface {
    ensures res == 42
    C() (res int)
}

ensures res == 42
func (a *A) C() (res int) {
    res = 42
    return
}

func client(a *A) {
  var i I = a
  res := i.C()
  assert res == 42
}
```

Felalolf:
Title. There was a bug that in implementation proofs, return parameters were not properly substituted in the specification that is copied from the interface method.


ArquintL (approved):
Looks good as well

jcp19 (approved):
Looks good to me

src/test/resources/regressions/issues/000312.gobra
```diff
@@ -0,0 +1,23 @@
...
+func (a *A) C() (res int) {
+    res = 42
+    return
+}
+
+func client(a *A) {
+  var i I = a
+  res := i.C()
+  assert res == 42
+}
```
Felalolf, line 23:
```suggestion
}

```

PR #324, opened 2021-07-22
Mathematical Maps, Sets, Multisets, Options, and Sequences with Declared Types

ArquintL:
Previously, declared types have not been allowed for the data structures mentioned in the title.
This PR removes these (artificial) restrictions and adds corresponding test cases checking that the encoding works properly.

### DIFF
--- a/src/test/resources/regressions/features/maps/maps-declared-type-simple1.gobra
+++ b/src/test/resources/regressions/features/maps/maps-declared-type-simple1.gobra
@@ -1,0 +1,24 @@
+package pkg
+
+type cell struct {
+		val int
+}
+
+func test1() {
+    m1 := map[int]cell { 0: cell{42} }
+    m2 := map[cell]int { cell{42}: 0 }
+    assert m1[0] == cell{42}
+    assert m2[cell{42}] == 0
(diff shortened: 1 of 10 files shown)


src/test/resources/regressions/features/maps/maps-declared-type-simple1.gobra
```diff
@@ -0,0 +1,27 @@
...
+
+func test2() {
+    c := &cell{42}
+    m1 := map[int]*cell { 0: c }
+    m2 := map[*cell]int { c: 0 }
+    assert m1[0] == c
+    assert m2[c] == 0
+    //:: ExpectedOutput(assert_error:assertion_error)
+    assert false
+}
```
jcp19, line 27: Just out of curiosity, is there any reason to add these tests for go maps, which are not mathematical types? At any rate, more tests cannot hurt :)

jcp19 (approved):

src/test/resources/regressions/features/maps/maps-declared-type-simple1.gobra
ArquintL: I only saw that both types of maps are in the same folder and basically translated the same unit test for both of them

Issue #246, opened 2021-03-30
Missing proof obligation for conversion of slices


jcp19:
The following verifies but it should not
```go
func m() {
	k := make([]byte, 1)
	exhale acc(&k[0])
	x := string(k)
}
```
The reason is that the conversion from a slice to a string should ensure that the method has read access to every element in the slice.

jcp19:
The other way around is also incomplete
```
func m() {
        s := "string"
	x := byte[](s)
        assert forall i int :: 0 <= i && i < len(s) ==> acc(&x[i]) // must hold
}
```

PR #328, opened 2021-07-30
Add support for effectful conversions

jcp19:
Adds support for conversions which might produce side effects (e.g. converting from `string` to `[]byte`). Furthermore, this fixes an unsoundness caused by issue #246.

### DIFF
--- a/src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
+++ b/src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
@@ -163,13 +163,22 @@ class StringEncoding extends LeafTypeEncoding {
       *   }
       * where `literal` is one of the generated unique domain functions and X is the length of the corresponding string
       */
-    val lenAxioms = encodedStrings.keys.toSeq.map { str =>
+    val litLenAxioms = encodedStrings.keys.toSeq.map { str =>
       vpr.AnonymousDomainAxiom {
         val encodedStr: vpr.Exp = vpr.DomainFuncApp(encodedStrings(str), Seq.empty, Map.empty)()
         val lenCall = vpr.DomainFuncApp(func = lenFunc, Seq(encodedStr), Map.empty)()
         vpr.EqCmp(lenCall, vpr.IntLit(BigInt(str.length))())()
       }(domainName = domainName)
     }
+
+    val lenAxiom = vpr.AnonymousDomainAxiom {
+      val qtfVar = vpr.LocalVarDecl("str", stringType)()
+      vpr.Forall(
+        variables = Seq(qtfVar),
+        triggers = Seq(),
+        exp = vpr.LeCmp(vpr.IntLit(0)(), vpr.DomainFuncApp(func = lenFunc, Seq(qtfVar.localVar), Map.empty)())()
+      )()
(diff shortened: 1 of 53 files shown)


src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
```diff
@@ -68,13 +75,61 @@ class StringEncoding extends LeafTypeEncoding {
...
+        // interesting properties
+        val (pos, info, errT) = conv.vprMeta
+
+        val sliceT = in.SliceT(in.IntT(Addressability.sliceElement, TypeBounds.Byte), Addressability.outParameter)
+        val slice = in.LocalVar(Names.freshName, sliceT)(conv.info)
+        val vprSlice = ctx.typeEncoding.variable(ctx)(slice)
+        val qtfVar = in.BoundVar("i", in.IntT(Addressability.boundVariable))(conv.info)
+        val post = in.SepForall(
+          vars = Vector(qtfVar),
+          triggers = Vector(),
```
ArquintL, line 110: could you use the indexed expression as a trigger?

src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
```diff
@@ -68,13 +75,61 @@ class StringEncoding extends LeafTypeEncoding {
...
+    }
+  }
+
+
+  /**
+    * Encodes the (effectul) conversion from a string to a []byte
+    * [ target = []byte(str) ] ->
+    *   [
+    *     var s []byte
+    *     inhale forall i Int :: 0 <= i && i < len(s) ==> acc(&s[i])
```
ArquintL, line 90: is it correct that you (currently) only take care of the permissions without restricting the resulting values in any way?

src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
```diff
@@ -209,4 +264,41 @@ class StringEncoding extends LeafTypeEncoding {
...
+  private val byteSliceToStrFuncGenerator: FunctionGenerator[Unit] = new FunctionGenerator[Unit] {
+    override def genFunction(@unused x: Unit)(ctx: Context): vpr.Function = {
+      val info = Source.Parser.Internal
+      val paramT = in.SliceT(in.IntT(Addressability.sliceElement, TypeBounds.Byte), Addressability.outParameter)
+      val param = in.Parameter.In("s", paramT)(info)
+      val res = in.Parameter.Out("res", in.StringT(Addressability.outParameter))(info)
+      val qtfVar = in.BoundVar("i", in.IntT(Addressability.boundVariable))(info)
+      val pre = in.SepForall(
+        vars = Vector(qtfVar),
+        triggers = Vector(),
```
ArquintL, line 282: same here

ArquintL (approved):
a few smaller remarks

src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
jcp19: Yes, that is the point. We can enrich the specification once we introduce more string operations (e.g. indexing strings)

Issue #334, opened 2021-08-16
`Error: Resource not accessible by integration` on PRs originating from forks


jcp19:
As described [here](https://github.community/t/github-actions-are-severely-limited-on-prs/18179), Github Actions that modify the original repository cannot be performed on PRs based on forks. This is a [well known problem](https://github.com/actions/first-interaction/issues/10). Because of that, our CI cannot successfully run on external contributions (e.g. PR #333)

PR #336, opened 2021-08-17
Fix 334

jcp19:
Fixes issue #334. Instead of using `LouisBrunner/checks-action@v1.1.2` to publish the RAM usage annotations, this PR uses Github's [workflow actions](https://docs.github.com/en/enterprise-server@2.22/actions/reference/workflow-commands-for-github-actions), thus circumventing the need for a GITHUB_TOKEN that has write access to the repository. Consequently, PRs originating from forks can now successfully run this step.

One unfortunate consequence of these changes is that all annotations are published either as warnings or errors; there is no way to output an "info" annotation via workflow actions that I am aware of (there is a way to output debug annotations but that has unintended consequences, as described in a comment in the `test.yml` file).

@ArquintL I opted not to have a distinction on the origin of the PR (fork or not). This way, we provide a consistent experience for all PRs and it greatly simplifies our script.

I tested these changes with a PR originating from my fork (#335).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -102,9 +102,9 @@ jobs:
       #   # run: pidstat 1 -r -p ALL --human -e java -Xss128m -XshowSettings:vm -cp target/scala-2.13/gobra-test.jar -Dlogback.configurationFile=conf/logback.xml -DGOBRATESTS_REGRESSIONS_DIR=src/test/resources/regressions org.scalatest.run viper.gobra.GobraTests
       #   working-directory: gobra
 
-#      - name: Execute all tests
-#        run: sbt test
-#        working-directory: gobra
+      - name: Execute all tests
+        run: sbt test
+        working-directory: gobra
 
       - name: Terminate pidstat
         if: ${{ always() }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -170,43 +170,43 @@ jobs:
       - name: Echo
         run: echo "the full name is ${{ github.event_name != 'pull_request' || (github.event_name == 'pull_request' && github.repository == github.event.pull_request.head.repo.full_name) }}"
 
-#      - name: Create annotations
-#      #TODO: proper comment
-#        if: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
-#        uses: LouisBrunner/checks-action@v1.1.2
-#        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -154,7 +163,7 @@ jobs:
           MAX_Z3_GB=$(convertKbToGb $MAX_Z3_KB)
           JAVA_LEVEL=$(getLevel $MAX_JAVA_GB ${{ env.JAVA_WARNING_THRESHOLD_GB }} ${{ env.JAVA_FAILURE_THRESHOLD_GB }})
           Z3_LEVEL=$(getLevel $MAX_Z3_GB ${{ env.Z3_WARNING_THRESHOLD_GB }} ${{ env.Z3_FAILURE_THRESHOLD_GB }})
-          if [[ "$JAVA_LEVEL" = "failure" || "$Z3_LEVEL" = "failure" ]]
+          if [[ "$JAVA_LEVEL" = "${{ env.FAILURE_LEVEL }}" || "$Z3_LEVEL" = "${{ env.FAILURE_LEVEL }}" ]]
           then
             CONCLUSION="failure"
```
ArquintL, line 168:
```suggestion
            CONCLUSION="${{ env.FAILURE_LEVEL }}"
```

.github/workflows/test.yml
```diff
@@ -169,35 +178,16 @@ jobs:
...
-              "end_line": 1,
-              "annotation_level": "${{ env.Z3_LEVEL }}"
-              }
-            ]
+        # Outputs the memory consumption of JAVA and Z3. The message format is described in
+        # https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions .
+        run: |
+          echo "::${{ env.JAVA_LEVEL }} file=.github/workflows/test.yml,line=1::Java used up to ${{ env.MAX_JAVA_GB }}GB of RAM"
+          echo "::${{ env.Z3_LEVEL }} file=.github/workflows/test.yml,line=1::Z3 used up to ${{ env.MAX_Z3_GB }}GB of RAM"
+          if [ $CONCLUSION = "failure" ]
```
ArquintL, line 186:
```suggestion
          if [ $CONCLUSION = "${{ env.FAILURE_LEVEL }}" ]
```

ArquintL (approved):
Looks good although I would be in favor of simply echoing the stats (without warning) if they are below the warning threshold

PR #348, opened 2021-10-19
Add docker image to package repository

jogasser:
Added a docker image, containing gobra, z3 and boogie to be used by the [Gobra GitHub Action](https://github.com/jogasser/gobra-action) and added a workflow publishing to the GitHub package repository

### DIFF
--- a/workflow-container/Dockerfile
+++ b/workflow-container/Dockerfile
@@ -76,7 +76,7 @@ RUN apt-get update; \
 	rm packages-microsoft-prod.deb && \
 	apt-get update && \
 	apt-get install -y dotnet-sdk-5.0 && \
-        apt-get remove -y apt-transport-https wget && apt-get autoremove -y
+      apt-get remove -y apt-transport-https wget && apt-get autoremove -y
 
 RUN dotnet tool install --global boogie
 
--- a/workflow-container/Dockerfile
+++ b/workflow-container/Dockerfile
@@ -1,0 +1,71 @@
+FROM openjdk:11.0.12-slim as build
+
+WORKDIR /build
+
+RUN apt-get update && \
+	apt-get install -y \
+	gnupg2 \
+	curl \
+	unzip \
+	git
+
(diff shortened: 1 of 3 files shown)


workflow-container/Dockerfile
```diff
@@ -0,0 +1,87 @@
...
+	wget https://packages.microsoft.com/config/debian/11/packages-microsoft-prod.deb -O packages-microsoft-prod.deb && \
+	dpkg -i packages-microsoft-prod.deb &&\
+	rm packages-microsoft-prod.deb && \
+	apt-get update && \
+	apt-get install -y dotnet-sdk-5.0 && \
+        apt-get remove -y apt-transport-https wget && apt-get autoremove -y
+
+RUN dotnet tool install --global boogie
+
+ENV BOOGIE_EXE /root/.dotnet/tools/boogie
```
ArquintL, line 77: nice to include Boogie! Did you check whether the setup is working? I'm only used to take the Boogie binary from the Viper tool releases ([here](https://github.com/viperproject/viper-ide/releases)), which in turn is based on our own Boogie compilation [here](https://github.com/viperproject/boogie-builder/releases) as we had to fix several things before compilation was actually possible. It might thus be better to use Boogie from there to make sure the same version is used against which Viper is tested. This technically also holds for Z3 although we ran some experiments and 4.8.7 seemed slightly better than 4.8.6 (as used by Viper).

ArquintL (dismissed):
Only some very minor comments. We should (in the Gobra team) have a separate discussion of how to maintain the dockerfiles and docker builds as we have now 3 similar ones

workflow-container/Dockerfile
```diff
@@ -0,0 +1,87 @@
+FROM openjdk:11.0.12-slim as build
```
ArquintL, line 1:
```suggestion
# This Source Code Form is subject to the terms of the Mozilla Public
# License, v. 2.0. If a copy of the MPL was not distributed with this
# file, You can obtain one at http://mozilla.org/MPL/2.0/.
#
# Copyright (c) 2011-2021 ETH Zurich.

FROM openjdk:11.0.12-slim as build
```

jcp19 (approved):

PR #354, opened 2021-10-27
Add support for floats

jcp19:
This PR adds support for the float types (`float32` and `float64`) and translates operations on floats to uninterpreted functions. It only adds support for the types and operations, literals will be added in a later PR.

This PR contributes to putting the master branch on par with the verified scion branch.


src/test/resources/regressions/features/floats/floats-fail-02.gobra
```diff
@@ -0,0 +1,22 @@
...
+}
+
+//:: ExpectedOutput(postcondition_error)
+ensures res != x + y || res != x - y || res != x * y || res != x / y
+func g(x float32, y float32) (res float32) {
+	return x + y
+}
+
+//:: ExpectedOutput(postcondition_error)
+ensures res == y
```
Felalolf, line 19: would `res == x` go through?

Felalolf: (I now looked at the encoding, and the answer should be no, but maybe add a test case, as it should not hold)

src/main/scala/viper/gobra/translator/encodings/FloatEncoding.scala
```diff
@@ -0,0 +1,297 @@
...
+    */
+  override def expr(ctx: Context): in.Expr ==> CodeWriter[vpr.Exp] = {
+
+    def goE(x: in.Expr): CodeWriter[vpr.Exp] = ctx.expr.translate(x)(ctx)
+
+    default(super.expr(ctx)) {
+      case (e: in.DfltVal) :: ctx.Float32() / Exclusive =>
+        unit(withSrc(vpr.FuncApp(func = defaultValue32, Seq()), e))
+      case (e: in.DfltVal) :: ctx.Float64() / Exclusive =>
+        unit(withSrc(vpr.FuncApp(func = defaultValue64, Seq()), e))
```
Felalolf, line 59: You could make the default value the conversion from 0 to the respective type (e.g. `float32(0)` ).

Felalolf (approved):

Felalolf (approved):
I forgot to write text. Please change the default value and add a test as described. Otherwise, everything looks good to me and can be merged afterwards.

src/main/scala/viper/gobra/translator/encodings/FloatEncoding.scala
```diff
@@ -0,0 +1,270 @@
...
+    * Encodes expressions as values that do not occupy some identifiable location in memory.
+    * X stands for either 32 or 64 below:
+    * [ dflt(x: floatX) ] -> [ floatX(0) ]
+    * [ (x: floatX) + (y: floatX) ] -> addFloatX([ x ], [ y ])
+    * [ (x: floatX) - (y: floatX) ] -> subFloatX([ x ], [ y ])
+    * [ (x: floatX) * (y: floatX) ] -> mulFloatX([ x ], [ y ])
+    * [ (x: floatX) / (y: floatX) ] -> divFloatX([ x ], [ y ])
+    * [ floatX(x: int) ] -> fromIntToX([ x ])
+    */
+  override def expr(ctx: Context): in.Expr ==> CodeWriter[vpr.Exp] = {
```
ArquintL, line 52: contains quite a bit of repetition. One could parameterize the `addFloat`, `subFloat`, etc. functions by the float size but I'm not sure whether that worth the code complexity

ArquintL (approved):

PR #358, opened 2021-11-01
Fix issue with checking assignability

jcp19:


Felalolf (approved):

Felalolf (commented):
Is the case for array correct?

PR #367, opened 2021-11-22
Switch from Paths to Sources

ArquintL:
This PR adapts Gobra to take a vector of `Source` instead of `Path` as input. This provides greater flexibility as a `Source` can either represent a file or simply some string. In particular, it allows Gobra Server to invoke Gobra with some (file)content that is not necessarily stored in local files.
Not however that this does not immediately enables Gobra Server to run on a remote machine as the entire import resolution mechanism still relies on the local filesystem. A proper abstraction would need to be implemented (e.g. with callbacks to query the directory structure and/or to retrieve files).

### DIFF
--- a/src/main/scala/viper/gobra/ast/frontend/Ast.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/Ast.scala
@@ -6,12 +6,11 @@
 
 package viper.gobra.ast.frontend
 
-import java.nio.file.Paths
 import org.bitbucket.inkytonik.kiama.rewriting.Rewritable
 import org.bitbucket.inkytonik.kiama.util.Messaging.Messages
 import org.bitbucket.inkytonik.kiama.util._
 import viper.gobra.ast.frontend.PNode.PPkg
-import viper.gobra.frontend.Source.FromFileSource
+import viper.gobra.frontend.Source.TransformableSource
 import viper.gobra.reporting.VerifierError
 import viper.gobra.util.{Decimal, NumBase}
 import viper.silver.ast.{LineColumnPosition, SourcePosition}
@@ -72,11 +71,7 @@ class PositionManager(val positions: Positions) extends Messaging(positions) {
   }
 
   def translate(start: Position, end: Position): SourcePosition = {
-    val path = start.source match {
-      case FileSource(filename, _) => Paths.get(filename)
-      case FromFileSource(path, _) => path
-      case StringSource(_, _) => null // StringSource stores some name but there is no guarantee that it corresponds to a valid file path
-    }
+    val path = start.source.toPath
     new SourcePosition(
       path,
       LineColumnPosition(start.line, start.column),
--- a/src/main/scala/viper/gobra/ast/frontend/Ast.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/Ast.scala
@@ -7,12 +7,11 @@
 package viper.gobra.ast.frontend
 
 import java.nio.file.Paths
-
 import org.bitbucket.inkytonik.kiama.rewriting.Rewritable
 import org.bitbucket.inkytonik.kiama.util.Messaging.Messages
 import org.bitbucket.inkytonik.kiama.util._
 import viper.gobra.ast.frontend.PNode.PPkg
-import viper.gobra.frontend.Parser.FromFileSource
+import viper.gobra.frontend.Source.FromFileSource
 import viper.gobra.reporting.VerifierError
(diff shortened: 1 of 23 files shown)


src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -129,37 +112,53 @@ object Parser {
...
-      }
+    def samePackage(programs: Vector[PProgram]): Either[Vector[ParserError], Vector[PProgram]] = {
+      require(programs.nonEmpty)
+      val pkgName = programs.head.packageClause.id.name
+      val differingPkgNameMsgs = programs.flatMap(p =>
+        error(
+          p.packageClause,
+          s"Files have differing package clauses, expected $pkgName but got ${p.packageClause.id.name}",
+          p.packageClause.id.name != pkgName))
+      println(s"samePackage msgs: ${differingPkgNameMsgs}")
```
jcp19: is this print supposed to be here or is this just a debugging device?

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -129,37 +112,53 @@ object Parser {
...
+    val res = for {
+      // check that each of the parsed programs has the same package clause. If not, the algorithm collecting all files
+      // of the same package has failed or users have entered an invalid collection of inputs
+      programs <- isErrorFree(parserResults)
+      programs <- samePackage(programs)
+      pkg <- makePackage(programs)
+    } yield pkg
+    // report potential errors:
+    res.left.map(errors => {
+      println(s"errors: ${errors}")
```
jcp19: same as before

jcp19 (approved):
LGTM, I just have a few minor comments

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -76,7 +75,7 @@ class PositionManager(val positions: Positions) extends Messaging(positions) {
     val path = start.source match {
       case FileSource(filename, _) => Paths.get(filename)
       case FromFileSource(path, _) => path
-      case _ => ???
+      case StringSource(_, _) => null // StringSource stores some name but there is no guarantee that it corresponds to a valid file path
```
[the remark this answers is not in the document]
ArquintL: I've left it at `???` and mentioned multiple ways how to potentially fix this in the future as I currently do not need it for supporting packages in the IDE

src/main/scala/viper/gobra/frontend/Source.scala
```diff
@@ -0,0 +1,74 @@
...
+      case FromFileSource(path, _) => path
+      case StringSource(_, _) =>
+        // StringSource stores some name but there is no guarantee that it corresponds to a valid file path
+        // there are the following ideas:
+        // - create a file adhoc if a filepath is really necessary (in the spirit of `StringSource.useAsFile`)
+        // - adapt Silver to not require a Path object but some abstraction
+        // - create an implementation implementing Path that simply stores the StringSource and leaves all functions
+        //    unimplemented. This works as long as nobody relies on any operations available for paths. It's also bad
+        //    style but at least any use fails in an understandable way (as opposed to simply returning `null` here)
+      ???
```
jcp19, line 40: This is a lot safer, thanks!

src/main/scala/viper/gobra/reporting/Message.scala
```diff
@@ -46,7 +46,12 @@ case class GobraOverallFailureMessage(verifier: String, result: VerifierResult)
     s"failure=${result.toString})"
 }
 
-case class GobraEntitySuccessMessage(verifier: String, concerning: Source.Verifier.Info) extends GobraVerificationResultMessage {
+sealed trait GobraEntityResultMessage extends GobraVerificationResultMessage {
```
jcp19, line 49: What is the purpose of storing `entity` here? I might have missed something but it does not seem to be used by any of its subcalsses

jcp19 (approved):

src/main/scala/viper/gobra/reporting/Message.scala
ArquintL: It allows Gobra Server to figure out to which entity a message belongs. In particular, it keeps track of the progress by having a list of entities extracted from the Viper AST and checking whether an entity message belongs to any of these

jcp19: I see, you use it to keep track of the progress in the GobraIDE

PR #368, opened 2021-11-22
Fix for Issue 363

### LINKED ISSUE #363: CI builds image for wrong branches
[This run](https://github.com/viperproject/gobra/runs/4214601369) is an example in which the image was deployed despite the component variables being correctly set. It seems that the condition for setting `SHOULD_DEPLOY` behalves differently than expected.

Afterwards, the [list of image tags](https://github.com/viperproject/gobra/pkgs/container/gobra) should be cleaned up.

ArquintL:
Fixes #363

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -211,6 +211,9 @@ jobs:
           echo "TEST=$(($IS_GOBRA_REPO == 1))" >> $GITHUB_ENV
           echo "SHOULD_DEPLOY=$SHOULD_DEPLOY" >> $GITHUB_ENV
 
+      - name: Dummy step
+        run: echo ${{ env.SHOULD_DEPLOY }}
+
       - name: Login to Github Packages
         if: env.SHOULD_DEPLOY == 'true'
         uses: docker/login-action@v1
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -197,12 +197,12 @@ jobs:
       - name: Decide whether image should be deployed or not
         run: |
           # note that these GitHub expressions return the value '1' for 'true' (as opposed to bash)
-          IS_GOBRA_REPO=${{ github.repository == 'viperproject/gobra' }}
-          IS_MASTER=${{ github.ref == 'refs/heads/master' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -201,7 +201,8 @@ jobs:
           IS_MASTER=${{ github.ref == 'refs/heads/master' }}
           IS_VERIFIED_SCION=${{ github.ref == 'refs/heads/verified-scion-preview-without-selective' }}
           IS_PROPER_TAG=${{ startsWith(github.ref, 'refs/tags/v') }}
-          if [[ $IS_GOBRA_REPO && ($IS_MASTER || $IS_VERIFIED_SCION || $IS_PROPER_TAG) ]]
+          # note that `$IS_GOBRA_REPO == 1` is the same as `!$IS_GOBRA_REPO` because value 1 is interpreted as false in bash
+          if [[ $IS_GOBRA_REPO == 1 && ($IS_MASTER == 1 || $IS_VERIFIED_SCION == 1 || $IS_PROPER_TAG == 1) ]]
```
jcp19, line 205: I think this does what we want but it is confusing to read - from reading the condition alone, it looks like the opposite of what we want because everything is negated. I think I would probably change the variable declarations above to
```bash
          # note that these GitHub expressions return the value '1' for 'true', where bash considers "0" to be 'true'.
          # as such, all conditions are negated to produce the expected result in bash
          IS_GOBRA_REPO=${{ github.repository != 'viperproject/gobra' }}
          IS_MASTER=${{ github.ref != 'refs/heads/master' }}
          IS_VERIFIED_SCION=${{ github.ref != 'refs/heads/verified-scion-preview-without-selective' }}
          IS_PROPER_TAG=${{ !startsWith(github.ref, 'refs/tags/v') }}
```
and then I would keep the if-condition as it was before.

Nonetheless, if you prefer it as it is now, fine by me :)

jcp19 (approved):

.github/workflows/test.yml
ArquintL: Haha that's weird too but maybe `IS_GOBRA_REPO=!${{ github.repository == 'viperproject/gobra' }}` is some middle ground with a comment on the `!`

jcp19: Yes, that is fine ^^

PR #373, opened 2021-12-06
Reintegrate viperserver backend

jogasser:
# Reintegrate viperserver backend
- readds viperserver as a dependency
- adds ViperServerWithSilicon(VSWITHSILICON) and ViperServerWithCarbon(VSWITHCARBON) as potential backends to choose from
- adds cacheFile config option to specify a cache file to be used by viperserver


src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -211,6 +214,13 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
     noshort = false
   )
 
+  val cacheFile: ScallopOption[String] = opt[String](
+    name = "cacheFile",
+    descr = "Cache file to be used, only usable if using viper server as a backend",
```
jcp19, line 219: Would shorten to "Cache file to be used by Viper Server".

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -295,6 +305,14 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
   }
   validateInput(input, module, include)
 
+  // cache file should only be usable when using viper server
+  validateOpt (backend, cacheFile) {
+    case (Some(ViperBackends.ViperServerWithSilicon), Some(_)) => Right()
+    case (Some(ViperBackends.ViperServerWithCarbon), Some(_)) => Right()
+    case (_, None) => Right()
+    case (_, Some(_)) => Left("Cache file option only works with viper server as a backend")
```
jcp19, line 313: Would change to "Cache file can only be specified when the backend uses Viper Server"

src/main/scala/viper/gobra/backend/ViperBackends.scala
```diff
@@ -26,12 +31,75 @@ object ViperBackends {
...
+
+  object ViperServerWithSilicon extends ViperBackend {
+    def create(exePaths: Vector[String], config: Config)(implicit executor: GobraExecutionContext): ViperServer = {
+      if(ViperServerBackend.executor == null) {
+        ViperServerBackend.setExecutor(executor)
+      }
+
+      if(ViperServerBackend.server == null) {
+        ViperServerBackend.initServer(config)
+      }
```
jcp19, line 79: I wonder if we should have these if statements surrounding the calls to `setExecutor` and `initServer`. This looks like a "silent failure" and might become a bit confusing for debugging later. Does it even make sense to not fail here?

src/main/scala/viper/gobra/backend/ViperBackends.scala
```diff
@@ -26,12 +31,75 @@ object ViperBackends {
...
+
+  object ViperServerWithCarbon extends ViperBackend {
+    def create(exePaths: Vector[String], config: Config)(implicit executor: GobraExecutionContext): ViperServer = {
+      if(ViperServerBackend.executor == null) {
+        ViperServerBackend.setExecutor(executor)
+      }
+
+      if(ViperServerBackend.server == null) {
+        ViperServerBackend.initServer(config)
+      }
```
jcp19, line 99: same as above

viper-toolchain-versions.sh
```diff
@@ -8,3 +8,4 @@
 SILVER_REF="f645539a04e76d70337d7b5b108e7f9a9de9bc7a"
 SILICON_REF="v.21.07-release"
 CARBON_REF="v.21.07-release"
+VIPERSERVER_REF="master"
```
jcp19, line 11: Is there any reason not to use version "v.21.07-release" like with the other dependencies?

src/test/scala/viper/gobra/GobraTests.scala
```diff
@@ -29,12 +30,17 @@ class GobraTests extends AbstractGobraTests with BeforeAndAfterAll {
...
 
   override def beforeAll(): Unit = {
+    executor = new DefaultGobraExecutionContext()
     gobraInstance = new Gobra()
   }
 
   override def afterAll(): Unit = {
+    executor.terminateAndAssertInexistanceOfTimeout()
+    ViperServerBackend.resetServer();
+    ViperServerBackend.resetExecutor();
```
jcp19, line 43: why do we do this here?

jcp19 (changes requested):
In general, looks good but I would add Viper Server as a dependency in our ReadMe, in the installation instructions

src/main/scala/viper/gobra/backend/ViperBackends.scala
jogasser: Yes, the goal here is to initialize the ViperServer environment if it wasn't previously. `create` gets called for each verification, if we do it like this, we can use a single ViperServer instance for multiple verification runs.

viper-toolchain-versions.sh
jogasser: There will be after https://github.com/viperproject/viperserver/pull/44 is merged. This then needs to be changed to the new commit

src/main/scala/viper/gobra/backend/ViperBackends.scala
jogasser: I updated it to better handle executor shutdowns

src/test/scala/viper/gobra/GobraTests.scala
jogasser: I actually updated `ViperBackends` such that we don't need this here

src/main/scala/viper/gobra/backend/ViperBackends.scala
jcp19: > Yes, the goal here is to initialize the ViperServer environment if it wasn't previously. create gets called for each verification, if we do it like this, we can use a single ViperServer instance for multiple verification runs.

I would add a comment in the code explaining exactly what you said

jcp19:
PS: we should adapt the [README](https://github.com/viperproject/gobra/blob/master/README.md) with the new dependency

src/test/scala/viper/gobra/GobraTests.scala
```diff
@@ -9,6 +9,7 @@ package viper.gobra
 import java.nio.file.Path
 import ch.qos.logback.classic.Level
 import org.scalatest.BeforeAndAfterAll
+import viper.gobra.backend.ViperBackends.ViperServerBackend
```
jcp19, line 12: This imports does not seem to be needed

src/main/scala/viper/gobra/backend/ViperServer.scala
```diff
@@ -0,0 +1,86 @@
...
+
+import scala.concurrent.{Await, Future, Promise}
+import viper.gobra.util.GobraExecutionContext
+import viper.server.core.{CarbonConfig, SiliconConfig, VerificationExecutionContext, ViperBackendConfig, ViperCoreServer, ViperServerBackendNotFoundException}
+
+import scala.concurrent.duration.Duration
+
+object ViperServer {
+
+  case object Result
```
jcp19, line 22: what is the purpose of this `Result` object? Is this even used?

jcp19 (approved):
LGTM

ArquintL:
@jogasser can you please update Silicon to commit [`22551b40509c381991a493da0108ea4c97fd602d`](https://github.com/viperproject/silicon/commit/22551b40509c381991a493da0108ea4c97fd602d)? Thanks!

jcp19 (commented):
Looks fine to me

PR #374, opened 2021-12-06
added stubs for As, Is, and Unwrap methods of errors package

LinoTelschow:
As Is:
- errors package is not supported

To Be:
- adds stubs for the errors.Is, errors.As, and errors.Unwrap methods

### DIFF
--- a/src/main/resources/stubs/errors/wrap.gobra
+++ b/src/main/resources/stubs/errors/wrap.gobra
@@ -2,6 +2,9 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
+// Signatures for the public declarations in file
+// https://github.com/golang/go/blob/master/src/errors/wrap.go
+
 package errors
 
 // (tlino) the methods aren't verified yet
--- a/src/main/resources/stubs/errors/wrap.gobra
+++ b/src/main/resources/stubs/errors/wrap.gobra
@@ -1,4 +1,4 @@
-// Copyright 2018 The Go Authors. All rights reserved.
+// Copyright 2009 The Go Authors. All rights reserved.
 // Use of this source code is governed by a BSD-style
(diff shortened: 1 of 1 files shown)


jcp19 (changes requested):

src/main/resources/stubs/errors/wrap.gobra
```diff
@@ -0,0 +1,112 @@
...
+
+// (tlino) the methods aren't verified yet
+// import (
+// 	"internal/reflectlite"
+// )
+
+// Unwrap returns the result of calling the Unwrap method on err, if err's
+// type contains an Unwrap method returning error.
+// Otherwise, Unwrap returns nil.
+decreases
```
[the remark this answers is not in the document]
LinoTelschow: sure!
I will adapt it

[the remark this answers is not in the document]
LinoTelschow: Done!

src/main/resources/stubs/errors/wrap.gobra
```diff
@@ -0,0 +1,112 @@
...
+// As(target) returns true. In the latter case, the As method is responsible for
+// setting target.
+//
+// An error type might provide an As method so it can be treated as if it were a
+// different error type.
+//
+// As panics if target is not a non-nil pointer to either a type that implements
+// error, or to any interface type.
+decreases _
+func As(err error, target interface{}) bool //{
```
jcp19, line 85: I would also add here a comment saying that, for now, we do not require any predicate to hold for `target` but at some point, we might add it to ensure that `target` can be safely read.

LinoTelschow: sure, I will change that

LinoTelschow: Done!

jcp19 (approved):

PR #376, opened 2021-12-13
added termination contracts to the strconv package

LinoTelschow:
As Is:
- termination of the methods in `strconv` isn't verified

To Be:
- verified termination of methods in `strconv`

### DIFF
--- a/src/main/resources/stubs/strconv/atoi.gobra
+++ b/src/main/resources/stubs/strconv/atoi.gobra
@@ -25,12 +25,14 @@ type NumError struct {
 requires p > 0
 requires acc(e, p)
 ensures acc(e, p)
+decreases _
 func (e *NumError) Error(ghost p perm) string /*{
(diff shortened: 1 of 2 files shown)


src/main/resources/stubs/strconv/atoi.gobra
```diff
@@ -25,12 +25,14 @@ type NumError struct {
 requires p > 0
 requires acc(e, p)
 ensures acc(e, p)
+decreases _
 func (e *NumError) Error(ghost p perm) string /*{
 	return "strconv." + e.Func + ": " + "parsing " + Quote(e.Num) + ": " + e.Err.Error()
 }*/
```
jcp19, line 31: right now, it is not sound to assume that this terminates becaure Gobra does not guarantee that the Error method in `error` does not terminate

LinoTelschow: ok, in that case I will remove this contract

jcp19 (approved):
LGTM

PR #377, opened 2021-12-20
ViperServer Backend Improvements

ArquintL:
Adapts ViperServer backend (reintegrated by @jogasser) to be usable by Gobra Server. In particular, Gobra Server creates a ViperCoreServer instance as part of its initialization. This PR enables Gobra Server to pass this server instance to either `ViperServerWithSilicon` or `ViperServerWithCarbon` instead of a new server instance being created.
Furthermore, `ViperVerifierConfig` has been removed as I've introduced it back then to route configuration options via Gobra to the backend but this does not seem to be necessary anymore and reduces complexity.

### DIFF
--- a/src/main/scala/viper/gobra/backend/Carbon.scala
+++ b/src/main/scala/viper/gobra/backend/Carbon.scala
@@ -16,7 +16,7 @@ import scala.concurrent.Future
 
 class Carbon(commandLineArguments: Seq[String]) extends ViperVerifier {
 
-  override def verify(programID: String, reporter:Reporter, program: Program)(executor: GobraExecutionContext): Future[VerificationResult] = {
+  override def verify(programID: String, reporter: Reporter, program: Program)(executor: GobraExecutionContext): Future[VerificationResult] = {
     // directly declaring the parameter implicit somehow does not work as the compiler is unable to spot the inheritance
     implicit val _executor: GobraExecutionContext = executor
     Future {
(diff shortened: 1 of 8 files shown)


src/main/scala/viper/gobra/backend/Carbon.scala
```diff
@@ -16,7 +16,7 @@ import scala.concurrent.Future
 
 class Carbon(commandLineArguments: Seq[String]) extends ViperVerifier {
 
-  def verify(programID: String, config: ViperVerifierConfig, reporter:Reporter, program: Program)(executor: GobraExecutionContext): Future[VerificationResult] = {
+  override def verify(programID: String, reporter:Reporter, program: Program)(executor: GobraExecutionContext): Future[VerificationResult] = {
```
jcp19, line 19: Nitpick: we should add a space between `reporter:` and `Reporter` to be consistent in our code style.

src/main/scala/viper/gobra/backend/ViperBackends.scala
```diff
@@ -40,72 +40,56 @@ object ViperBackends {
...
-      require(executor == null || executor.executorService.isShutdown)
-      executor = executionContext
-      // reset server since it depends on the execution context
-      resetServer()
-    }
-
-    def initServer(config: Config): Unit = {
-      require(server == null, "ViperCoreServer is already set.")
-      require(executor != null, "Executor is not set.")
+    /** should be implemented by classes implementing this abstract class */
```
jcp19, line 46: This comment could be shortened to `/* abstract method */`. 

Additionally, I would add a short comment to the methods in this class describing their intended use, specially in the abstract method (`getViperVerifierConfig`). This would be useful for future reference.

jcp19 (approved):
LGTM! I just have two minor comments

PR #378, opened 2021-12-20
Trusted Flag

jogasser:
- Adds a trusted flag to Gobra, which lets a user assume that the specification of a method or function holds without checking it. To achieve this, it ignores the body of the function or method
- Upgrades sbt to 1.5.7 to fix log4j CVE(see, https://www.scala-lang.org/blog-detail/2021/12/16/state-of-log4j-in-scala-ecosystem.html for more details)

### DIFF
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -44,6 +44,7 @@ import (
 ```go
 requires ... // preconditions 
 ensures  ... // postconditions 
+trusted      // trusted flag (optional)
 func min(xs []int) int {
   ... // function body (optional)
 }
@@ -52,6 +53,7 @@ func min(xs []int) int {
 ```go
 requires ... // preconditions 
 ensures  ... // postconditions 
+trusted      // trusted flag (optional)
 func (xs *list) contains(value int) bool {
   ... // method body (optional)
 }
@@ -59,7 +61,7 @@ func (xs *list) contains(value int) bool {
 - Functions and methods can be annotated with pre and postconditions. If omitted, they default to `true`.
 - Functions and methods are verified modularly. A call only uses the specification of the callee and cannot peek inside the body.
 - The body (including `{}`) may be omitted. In this case, Gobra assumes that the specification holds.
-
+- A function or method can be annotated with the `trusted` flag. Gobra then assumes the specification holds and ignores the body.
 
 #### Predicates
 
--- a/src/main/scala/viper/gobra/frontend/Parser.scala
+++ b/src/main/scala/viper/gobra/frontend/Parser.scala
@@ -447,7 +447,8 @@ object Parser {
         name <- "func" ~> idnDef
         sig  <- signature
         body <- if (spec.isTrusted) nestedCurlyBracketsConsumer else specOnlyParser(blockWithBodyParameterInfo)
-      } yield PFunctionDecl(name, sig._1, sig._2, spec, body).at(spec)
+        // Need to update the position range, since otherwise only the body bzw signature position is accounted for
+      } yield PFunctionDecl(name, sig._1, sig._2, spec, body).range(spec, if (body.isDefined) body.get._2 else sig._2)
 
     lazy val functionSpec: Parser[PFunctionSpec] = {
       sealed trait FunctionSpecClause
@@ -489,7 +490,8 @@ object Parser {
         name <- idnDef
         sig  <- signature
         body <- if (spec.isTrusted) nestedCurlyBracketsConsumer else specOnlyParser(blockWithBodyParameterInfo)
-      } yield PMethodDecl(name, rcv, sig._1, sig._2, spec, body).at(spec)
+        // Need to update the position range, since otherwise only the body bzw signature position is accounted for
+      } yield PMethodDecl(name, rcv, sig._1, sig._2, spec, body).range(spec, if (body.isDefined) body.get._2 else sig._2)
 
     /**
       * Statements
--- a/src/main/scala/viper/gobra/frontend/Parser.scala
+++ b/src/main/scala/viper/gobra/frontend/Parser.scala
@@ -442,13 +442,14 @@ object Parser {
       (idnDef <~ "=") ~ typ ^^ { case left ~ right => PTypeAlias(right, left)}
 
     lazy val functionDecl: Parser[PFunctionDecl] =
-      functionSpec() ~ ("func" ~> idnDef) ~ signature ~ specOnlyParser(blockWithBodyParameterInfo) ^^ {
-        case spec ~ name ~ sig ~ body => PFunctionDecl(name, sig._1, sig._2, spec, body)
-      } | functionSpec(withTrusted = true) ~ ("func" ~> idnDef) ~ signature ~ nestedCurlyBracketsConsumer.? ^^ {
-        case spec ~ name ~ sig ~ _ => PFunctionDecl(name, sig._1, sig._2, spec, None)
-      }
+      for {
+        spec <- functionSpec
+        name <- "func" ~> idnDef
+        sig  <- signature
+        body <- if (spec.isTrusted) nestedCurlyBracketsConsumer else specOnlyParser(blockWithBodyParameterInfo)
+      } yield PFunctionDecl(name, sig._1, sig._2, spec, body).at(spec)
 
-    def functionSpec(withTrusted: Boolean = false): Parser[PFunctionSpec] = {
+    lazy val functionSpec: Parser[PFunctionSpec] = {
       sealed trait FunctionSpecClause
       case class RequiresClause(exp: PExpression) extends FunctionSpecClause
       case class PreservesClause(exp: PExpression) extends FunctionSpecClause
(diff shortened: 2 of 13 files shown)


src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -464,14 +465,21 @@ object Parser {
         "pure" <~ eos ^^^ PureClause
       }
 
-      functSpecClause.* ~ "pure".? ^^ {
+      val specClauses = if(withTrusted) {
+        functSpecClause | "trusted" <~ eos ^^^ TrustedClause
+      } else {
+        functSpecClause
+      }
```
jcp19, line 472: the variable `specClauses` is unnecessary. We can just add `trusted" <~ eos ^^^ TrustedClause` to the definition of `functSpecClause`.

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -464,14 +465,21 @@ object Parser {
...
+
+      specClauses.* ~ "pure".? ^^ {
```
jcp19, line 474: we can revert the changes in this line after incorporating my previous suggestion.

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -442,19 +442,20 @@ object Parser {
       (idnDef <~ "=") ~ typ ^^ { case left ~ right => PTypeAlias(right, left)}
 
     lazy val functionDecl: Parser[PFunctionDecl] =
-      functionSpec ~ ("func" ~> idnDef) ~ signature ~ specOnlyParser(blockWithBodyParameterInfo) ^^ {
-        case spec ~ name ~ sig ~ body =>
-          PFunctionDecl(name, sig._1, sig._2, spec, body)
+      functionSpec() ~ ("func" ~> idnDef) ~ signature ~ specOnlyParser(blockWithBodyParameterInfo) ^^ {
+        case spec ~ name ~ sig ~ body => PFunctionDecl(name, sig._1, sig._2, spec, body)
+      } | functionSpec(withTrusted = true) ~ ("func" ~> idnDef) ~ signature ~ nestedCurlyBracketsConsumer.? ^^ {
+        case spec ~ name ~ sig ~ _ => PFunctionDecl(name, sig._1, sig._2, spec, None)
```
jcp19, line 448: I think that there is some repetition here that we can likely avoid. In particular, both branches start by parsing the functionSpec and then obtain the identifier and signature. They only differ in the way that the body is parsed.

src/main/scala/viper/gobra/frontend/Parser.scala
jcp19: In particular, I do not think that we need the `withTrusted` parameter. Instead, we need to pick the parser for the body according to the result of parsing the spec (look at my suggestion above).

src/main/scala/viper/gobra/frontend/Parser.scala
jcp19: Something like the following could work:

```scala
    lazy val functionDecl: Parser[PFunctionDecl] =
      for {
        spec <- functionSpec // notice that I removed the param in functionSpec
        name <- ("func" ~> idnDef)
        sig  <- signature
        body <- if (spec.isTrusted) nestedCurlyBracketsConsumer else specOnlyParser(blockWithBodyParameterInfo)
      } yield PFunctionDecl(name, sig._1, sig._2, spec, body)
```

This can likely be improved to look better by using operators from the parsing library instead of the `for` notation

jcp19 (changes requested):

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -480,8 +488,10 @@ object Parser {
         repsep(expression, ",") ~ ("if" ~> expression).? ^^ PTupleTerminationMeasure
 
     lazy val methodDecl: Parser[PMethodDecl] =
-      functionSpec ~ ("func" ~> receiver) ~ idnDef ~ signature ~ specOnlyParser(blockWithBodyParameterInfo) ^^ {
+      functionSpec() ~ ("func" ~> receiver) ~ idnDef ~ signature ~ specOnlyParser(blockWithBodyParameterInfo) ^^ {
         case spec ~ rcv ~ name ~ sig ~ body => PMethodDecl(name, rcv, sig._1, sig._2, spec, body)
+      } | functionSpec(withTrusted = true) ~ ("func" ~> receiver) ~ idnDef ~ signature ~ nestedCurlyBracketsConsumer.? ^^ {
+        case spec ~ rcv ~ name ~ sig ~ _ => PMethodDecl(name, rcv, sig._1, sig._2, spec, None)
```
jcp19, line 494: same comments as for `functionDecl`

src/main/scala/viper/gobra/frontend/Parser.scala
jogasser: This was actually my first attempt of solving this task. I used the `into` operator of the parsing library. The problem with an approach like this is, that if you parse the function spec first, the position of the `PFunctionDecl` node moves down to the `func` key word, implying that the spec doesn't belong to a function declaration. I discussed this with Linard and we came to the conclusion, that this isn't desired behaviour. This is the reason why I used this non-elegant approach

jogasser: `regressions/features/interfaces/proofGenFail1.gobra` would be a test case, where the above described behaviour would cause problems

jcp19: I am not sure if I understand what you mean. Do you mean that the position associated with this AST node is not the intended one? If so, isn't this something that we can overcome by using the `at` method to specify exactly the position of a node? (there are some examples of this method in the Parser class)

jcp19: PS: verifying the file `regressions/features/interfaces/proofGenFail1.gobra` with my proposed code seems to lead to the same exact output as the latest commit of this branch. Is it possible that the erroneous behaviour that you found is caused by some special behaviour of the `into` operator?

jogasser: I didn't know about the `at` operator, this fixes the issue, thanks!

About the problem, the output for the test case is almost the same, the test output should be: 
```
...
[info] 	<proofGenFail1.gobra:24:1> Generated implementation proof (*counter implements interface{ hasNext, next }) failed. Precondition of call to implementation method might not hold. 
[info] Permission to x might not suffice.
[info] 	<proofGenFail1.gobra:30:1> Generated implementation proof (*counter implements interface{ hasNext, next }) failed. Precondition of call to implementation method might not hold. 
[info] Permission to x.f might not suffice.
...
```

without the `at` operator, even with your code, the output looks something like this (there is a difference in the error positions): 
```
...
[info] 	<proofGenFail1.gobra:25:39> Generated implementation proof (*counter implements interface{ hasNext, next }) failed. Precondition of call to implementation method might not hold. 
[info] Permission to x might not suffice.
[info] 	<proofGenFail1.gobra:33:42> Generated implementation proof (*counter implements interface{ hasNext, next }) failed. Precondition of call to implementation method might not hold. 
[info] Permission to x.f might not suffice.
...
```

jcp19 (approved):
It looks good to me now. @ArquintL do you have other comments? Otherwise, I suggest that we merge this when the CI has gone through.

docs/tutorial.md
```diff
@@ -44,6 +44,7 @@ import (
 ```go
 requires ... // preconditions 
 ensures  ... // postconditions 
+trusted      // trusted flag (optional)
```
ArquintL, line 47: thanks for keeping the tutorial up-to-date!

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -442,36 +442,39 @@ object Parser {
...
-      functionSpec ~ ("func" ~> idnDef) ~ signature ~ specOnlyParser(blockWithBodyParameterInfo) ^^ {
-        case spec ~ name ~ sig ~ body =>
-          PFunctionDecl(name, sig._1, sig._2, spec, body)
-      }
+      for {
+        spec <- functionSpec
+        name <- "func" ~> idnDef
+        sig  <- signature
+        body <- if (spec.isTrusted) nestedCurlyBracketsConsumer else specOnlyParser(blockWithBodyParameterInfo)
+      } yield PFunctionDecl(name, sig._1, sig._2, spec, body).at(spec)
```
ArquintL, line 450: Is it necessary to  call `.at(spec)` here? I would expect that Kiama sets the positional information itself but I'm not sure in the context of `for ... yield`. Letting Kiama do it would have the advantage that the end position is presumably after the body. As you use `spec` right now, the function's position only spans pre- & postconditions

ArquintL (approved):
One small question regarding positional information, otherwise looks good to me

src/main/scala/viper/gobra/frontend/Parser.scala
jogasser: You are correct, I replaced `.at(...)` with `.range(...)`. 

I don't really know if it makes more sense to let Kiama handle the position, since, with a solution where the spec gets parsed before the body, Kiama wouldn't include the spec in the function position. 

The previous solution, where I let Kiama set the position correctly(see `src/main/scala/viper/gobra/frontend/Parser.scala` in [62cdff1](https://github.com/viperproject/gobra/pull/378/commits/62cdff16b205bda2713f456a2d313215f5d25845)), is, as joao said a little bit repetitive and unelegant.

ArquintL:
Thanks for the improvement. However, this new positions do not cover the body if the trusted flag is set. I've created the attached patch that addresses this. You should simply be able to run `git apply parser.txt` (it's a patch file but GitHub does not allow me to upload files with that extension ^^) and then commit the changes (I can't do it as I don't have access to your fork).

[parser.txt](https://github.com/viperproject/gobra/files/7813625/parser.txt)

If it should ever be handy, I've used the following snippet to print the positional information of functions in the Success case in `parseSources`:
```
val fnDecls = ast.declarations.collect {
  case n: PFunctionDecl => n
}

println(s"position for function decls:\n${fnDecls.map(fn => s"${fn.id.name}: ${pom.positions.getStart(fn).map(pos => s"${pos.line}:${pos.column}")} - ${pom.positions.getFinish(fn).map(pos => s"${pos.line}:${pos.column}")}").mkString("\n")}")
```

ArquintL (changes requested):
Please apply my patch first

ArquintL (approved):

jcp19 (approved):
LGTM

PR #382, opened 2021-12-29
added bytes package stub and added support for time.sleep method

LinoTelschow:
As Is:
- no support for the bytes package
- no support for time.Sleep method

To Be:
- added bytes package
- added time.Sleep method

### DIFF
--- a/src/main/resources/stubs/bytes/bytes.gobra
+++ b/src/main/resources/stubs/bytes/bytes.gobra
@@ -20,7 +20,7 @@ import (
 // A nil argument is equivalent to an empty slice.
 requires forall i int :: 0 <= i && i < len(a) ==> acc(&a[i], _)
 requires forall i int :: 0 <= i && i < len(b) ==> acc(&b[i], _)
-decreases _
+decreases
 pure func Equal(a, b []byte) bool {
 	// Neither cmd/compile nor gccgo allocates for these string conversions.
 	return string(a) == string(b)
--- a/src/main/resources/stubs/bytes/bytes.gobra
+++ b/src/main/resources/stubs/bytes/bytes.gobra
@@ -1,0 +1,27 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in https://golang.org/LICENSE
+
+// Signatures for the public declarations in file
+// https://github.com/golang/go/blob/master/src/bytes/bytes.go
+
+// Package bytes implements functions for the manipulation of byte slices.
+// It is analogous to the facilities of the strings package.
+package bytes
+
+import (
+	// "internal/bytealg"
+	// "unicode"
+	// "unicode/utf8"
+)
+
+// Equal reports whether a and b
+// are the same length and contain the same bytes.
+// A nil argument is equivalent to an empty slice.
+requires forall i int :: 0 <= i && i < len(a) ==> acc(&a[i], _)
+requires forall i int :: 0 <= i && i < len(b) ==> acc(&b[i], _)
+decreases _
+pure func Equal(a, b []byte) bool {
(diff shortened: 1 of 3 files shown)


src/main/resources/stubs/bytes/bytes.gobra
```diff
@@ -0,0 +1,27 @@
...
+	// "unicode"
+	// "unicode/utf8"
+)
+
+// Equal reports whether a and b
+// are the same length and contain the same bytes.
+// A nil argument is equivalent to an empty slice.
+requires forall i int :: 0 <= i && i < len(a) ==> acc(&a[i], _)
+requires forall i int :: 0 <= i && i < len(b) ==> acc(&b[i], _)
+decreases _
```
jcp19, line 23: Why do we assume termination here instead of proving it?

src/main/resources/stubs/time/sleep.gobra
```diff
@@ -0,0 +1,13 @@
...
+
+// Signatures for the public declarations in file
+// https://github.com/golang/go/blob/master/src/time/sleep.go
+
+package time
+
+// Sleep pauses the current goroutine for at least the duration d.
+// A negative or zero duration causes Sleep to return immediately.
+decreases
+func Sleep(d Duration)
```
jcp19, line 13: The documentation does not clearly state that `Sleep` terminates always. In particular, it says that it "returns immediately" when the argument is less than or equal to 0 and it waits at least `d` when it is positive - it does not specify that it terminates in this case. I would change the termination measure to `decreases _ if d <= 0` (or something morally equivalent).

src/main/resources/stubs/bytes/bytes.gobra
LinoTelschow: I tried to prove it first, but it cannot be deduced

LinoTelschow: I can show you the problem in 20 minutes, (I need to wait for gobra to terminate)

src/main/resources/stubs/time/sleep.gobra
LinoTelschow: Ok, I assumed that it will always terminate in finite time. But I change it, such that we are on the safe side

src/main/resources/stubs/bytes/bytes.gobra
LinoTelschow: image
image

LinoTelschow: that's the issue I meant

jcp19: I see. The cause for this is that the conversions to string in `string(a)` and `string(b)` are translated to a function call at the Viper level. That function is defined in the function `byteSliceToStrFuncGenerator` in the class `StringEncoding`. Currently, the encoding does not have a termination measure, but we can assume that it always terminates. As such, adding a wildcard termination measure to that function should fix this issue.

LinoTelschow: ok, in that case we can leave it as it is!

LinoTelschow: or do you mean adding the termination measure to the StringEncoding function?

jcp19: I would add the termination measure to the StringEncoding to avoid similar problems in the future

LinoTelschow: Done

jcp19:
The CI is failing even though the tests are passing. This seems to be because Lino does not have enough permissions to execute the `build-push-action` action.

ArquintL:
> The CI is failing even though the tests are passing. This seems to be because Lino does not have enough permissions to execute the `build-push-action` action.

This problem should have been fixed by #371. Is your and/or Lino's branch up to date with master?

PR #384, opened 2022-01-06
Chopper

Felalolf:
As it has been foretold, it is now here: the chopper.

A student reported a bug in the merging step of the chopper for a high number of chopped programs, but I would like to investigate this after merging with master to minimize the changeset.

After the merge, all problems should be local to the ViperChopper file, meaning that there are no conflicts with everything else. I will also improve performance in a later PR. For instance, if a single method is selected, a depth-first-search instead of a strongly connected component analysis is sufficient.


src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -0,0 +1,363 @@
...
+    }}
+  }
+}
+
+object ViperChopper {
+
+  /** chops 'choppee' into independent Viper programs */
+  def chop(choppee: vpr.Program)(
+    isolate: Option[vpr.Method => Boolean] = None,
+    bound: Option[Int] = Some(1),
```
jcp19, line 43: Why does `bound` have type `Option[Int]`? What would it mean if I passed `None` as the value here?

src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -0,0 +1,363 @@
...
+            case _: Vertex.PredicateBody => 3
+            case _: Vertex.Function => 3
+            case _: Vertex.DomainFunction => 1
+            case _: Vertex.DomainType => 1
+            case _: Vertex.DomainAxiom => 4
+            case Vertex.Always => 0
+          }.sum
+
+          val lDiff = l.diff(r).flatMap(_.nodes)
+          val rDiff = r.diff(r).flatMap(_.nodes)
```
jcp19, line 216: Shouldn't `rDiff` be defined as `val rDiff = r.diff(l).flatMap(_.nodes)`?

src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -0,0 +1,363 @@
...
+    case class Function(functionName: String) extends Vertex
+    case class PredicateSig(predicateName: String) extends Vertex
+    case class PredicateBody(predicateName: String) extends Vertex
+    case class Field(fieldName: String) extends Vertex
+    case class DomainFunction(funcName: String) extends Vertex
+    case class DomainAxiom(v: vpr.DomainAxiom, d: vpr.Domain) extends Vertex
+    case class DomainType(v: vpr.DomainType) extends Vertex
+    case object Always extends Vertex // if something always has to be included
+
+    /** Returns the subprogram induces by the set of vertices. */
```
jcp19, line 246: typo? "induces" -> "induced"

jcp19 (approved):
In general, it looks OK but I have some questions. Also, I did not look closely at the implementation of Tarjan's algorithm

src/main/scala/viper/gobra/util/ViperChopper.scala
Felalolf: None is no bound, i.e. it will use the maximum number of generated Viper files.

src/main/scala/viper/gobra/backend/BackendVerifier.scala
```diff
@@ -47,16 +47,44 @@ object BackendVerifier {
...
-      })
+    verificationResults.map{ results =>
+      val result = results.foldLeft(silver.verifier.Success: VerificationResult){
+        case (acc, silver.verifier.Success) => acc
+        case (silver.verifier.Success, res) => res
+        case (l: silver.verifier.Failure, r: silver.verifier.Failure) => silver.verifier.Failure(l.errors ++ r.errors)
+      }
+      convertVerificationResult(result, task.backtrack)
+    }
   }
```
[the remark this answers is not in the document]
Felalolf: I will do it slightly different, but this is a good idea.

src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -0,0 +1,363 @@
...
+      }) && x.file.getFileName == target.file.getFileName
+    }
+
+    config.isolate.map { names => {
+      case Source(Source.Verifier.Info(_: PFunctionDecl, _, origin, _)) => names.exists(hit(_, origin.pos))
+      case Source(Source.Verifier.Info(_: PMethodDecl, _, origin, _)) => names.exists(hit(_, origin.pos))
+      case _ => false
+    }}
+  }
+}
```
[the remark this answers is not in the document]
Felalolf: good catch. I forgot to remove it. I moved this into another file because I wanted to keep the ViperChopper Gobra independent.

jcp19:
We can merge it after the CI is done if @ArquintL is ok with that

src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -182,21 +181,26 @@ object ViperChopper {
...
+      def inverse(node: Node, m: Mapping): T
+      def toProgram(set: Set[Node]): Program
+      def finish(y: Vector[Program], m: Mapping): Vector[Set[T]]
+      def penaltyAndMerge(l: Program, r: Program): (Int, Program)
     }
 
     object Penalty {
 
-      object Default extends MergePenalty[SCC.Component[Vertex]] {
+      trait DefaultImpl extends MergePenalty[SCC.Component[Vertex]] {
```
jcp19, line 197: Nitpick: even though this trait contains the entire default implementation of `MergePenalty`, it might be a bit misleading to have its name ending in `Impl`. I believe this suffix is most commonly used in names of classes and objects, not so much in traits.

jcp19 (approved):

src/main/scala/viper/gobra/util/ViperChopper.scala
Felalolf: I am open to name suggestions.

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -380,14 +417,23 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
...
       files.partition(_.isLeft) match {
         case (pkgs,  files) if pkgs.length == 1 && files.isEmpty => pkgs.head.swap.map(Left(_)).toOption
         case (pkgs, files) if pkgs.isEmpty && files.nonEmpty => Some(Right(for(Right(s) <- files.toVector) yield s.toPath))
         case _ => None
       }
     }
+
+    def isolatedPosition(cutInputWithIdxs: Option[List[(String, List[Int])]]): List[SourcePosition] = {
+      cutInputWithIdxs.map(_.flatMap { case (input, idxs) =>
+        isGoFilePath(input) match { // only go files can have a position
```
ArquintL, line 431: why is this a go-file-only feature?

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -302,7 +321,19 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
     validateFilesExist(include)
     validateFilesIsDirectory(include)
   }
-  validateInput(input, module, include)
+
+  val cutInputWithIdxs: ScallopOption[List[(String, List[Int])]] = input.map(_.map{ arg =>
```
ArquintL, line 325: `idx` means line numbers here, right? This could be made more clear with a short description

ArquintL (commented):
I've skipped `ViperChopper.scala` for now until Felix has pushed his additional improvements

src/main/scala/viper/gobra/util/ViperChopper.scala
jcp19: I would name the trait `Default` and the object that extends it `DefaultImpl`

src/main/scala/viper/gobra/frontend/Config.scala
Felalolf: it is not. I changed the comment.

src/main/scala/viper/gobra/util/ViperChopper.scala
Felalolf: I now made the trait a class

src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -0,0 +1,644 @@
...
+        )(program.pos, program.info, program.errT)
+      }
+    }
+  }
+
+  type Edge[T] = (T, T)
+  object Edges {
+
+    /** Returns all dependencies induced by a member. */
+    def dependencies(member: vpr.Member): Seq[Edge[Vertex]] = {
```
ArquintL, line 443: Why do you use `Seq` instead of `Set`? Does ordering have some semantics or offer better performance?

src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -0,0 +1,644 @@
...
+            val from = Vertex.DomainFunction(f.name)
+            usages(f).map(from -> _)
+          }
+
+        case _ => Vector.empty
+      }
+    }
+
+    /** Returns all entities referenced in the subtree of node `n`. */
+    def usages(n: vpr.Node): Seq[Vertex] = {
```
ArquintL, line 489: same here

src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -0,0 +1,644 @@
...
+        case f: vpr.Field =>
+          val from = Vertex.Field(f.name)
+          usages(f).map(from -> _)
+
+        case d: vpr.Domain =>
+          d.axioms.flatMap { ax =>
+            val mid = Vertex.DomainAxiom(ax, d)
+            val tos = usages(ax.exp)
+            val froms = tos // this is just an approximation for now. This could be improved.
+            val finalFroms = if (froms.nonEmpty) froms else Vector(Vertex.Always)
```
ArquintL, line 477: I do not understand these two lines. Do I understand it correctly that (for now) you're assuming that all usages in an axiom's body need this axiom and thus have a dependency on this axiom (seems somewhat reasonable). Line 477 then says that in case we have been unable to find any dependencies then we are conservative and include the axiom always.
Some comments would definitely have help me not to figure this out while typing ^^

src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -0,0 +1,644 @@
...
+  }
+
+  object Cut {
+
+    /**
+      * Returns a set of chopped programs where the number of programs is bounded by `bound`.
+      * @param N Number of nodes.
+      * @param nodes Nodes that must be included.
+      * @param edges Edges of the dependency graph.
+      * @param idToVertex Mapping from node ids to vertices used for price computations.
```
ArquintL, line 46: why do you operate on `Int` and use a mapping function to retrieve the corresponding `Vertex`? Is this better for performance?

src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -0,0 +1,644 @@
...
+      * @param bound Specifies upper bound on the number of returned programs.
+      *              If none, then maximum number of programs is returned.
+      * @param penalty Specifies penalty of merging programs.
+      * @return Set of programs.
+      */
+    def boundedCut[T](N: Int, nodes: Vector[Int], edges: Array[mutable.SortedSet[Int]], idToVertex: Int => Vertex)(
+      bound: Option[Int],
+      penalty: Penalty[Vertex]
+    ): Vector[List[Int]] = {
```
ArquintL, line 55: again a use of an ordered collection where I would expect `Set` (see comment below)

src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -0,0 +1,644 @@
...
+    }
+
+    /**
+      * Returns the set of smallest possible programs.
+      * @param N Number of nodes.
+      * @param nodes Nodes that must be included.
+      * @param edges Edges of the dependency graph.
+      * @return Set of smallest possible programs.
+      */
+    def smallestCut(N: Int, nodes: Vector[Int], edges: Array[mutable.SortedSet[Int]]): Vector[List[Int]] = {
```
ArquintL, line 93: same comment on ordered collection

src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -0,0 +1,644 @@
...
+      def dfs2(startColor: Int): immutable.SortedSet[Int] = {
+        val visitedColor  = Array.ofDim[Boolean](numColors)
+        val stack = mutable.Stack[Int](startColor)
+        var result = immutable.SortedSet.empty[Int]
+
+        while (stack.nonEmpty) {
+          val color = stack.pop()
+          if (!visitedColor(color)) {
+            visitedColor(color) = true
+            result ++= colorMembers(color) // using SSC, number of merges might go down
```
ArquintL, line 161:
```suggestion
            result ++= colorMembers(color) // using SCC, number of merges might go down
```

src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -0,0 +1,644 @@
...
+          if (!visited(node)) {
+            visited(node) = true
+            members += node
+            coloring(node) = color
+            stack.pushAll(edges(node))
+          } else {
+            val otherColor = coloring(node)
+            if (otherColor != color && !isAlreadyNeighbor(otherColor)) {
+              isAlreadyNeighbor(otherColor) = true
+              colorIsNotRoot(otherColor) = true
```
ArquintL, line 131: Do I understand it correctly that in the case of a mutually recursive function, the coloring of the first visited function would be marked as non-root. It does however not matter which of the two functions is marked root or non-root because they will be placed in the same component. Mutually recursive methods do not even show up here because there's only a dependency on the other method's spec

src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -0,0 +1,644 @@
...
+        onStack.addOne(v, true)
+        // find all successors
+        val succs = edges.collect{ case (`v`, succ) => succ }
+        for (s <- succs) {
+          if (!indices.contains(s)) {
+            // successor has not been visited yet
+            strongConnect(s)
+            lowLinks.update(v, Math.min(lowLinks.getOrElse(v, -1), lowLinks.getOrElse(s, -1)))
+          } else if (onStack.getOrElse(s, false)) {
+            // s is already on the stack and therefore in the current SSC
```
ArquintL, line 543:
```suggestion
            // s is already on the stack and therefore in the current SCC
```

src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -0,0 +1,644 @@
...
+        onStack(v) = true
+        // find all successors
+        val succs = edges(v)
+        for (s <- succs) {
+          if (!visited(s)) {
+            // successor has not been visited yet
+            strongConnect(s)
+            lowLinks.update(v, Math.min(lowLinks(v), lowLinks(s)))
+          } else if (onStack(s)) {
+            // s is already on the stack and therefore in the current SSC
```
ArquintL, line 602:
```suggestion
            // s is already on the stack and therefore in the current SCC
```

ArquintL (approved):
First of all that's impressive work you've done here. I have a couple of questions now and then and in general think that many different thoughts went into various parts that are not (yet) captured by comments. I'm just thinking of someone trying to improve parts in the future...

src/main/scala/viper/gobra/util/ViperChopper.scala
Felalolf: I am ordering at a later point in time, where it is faster.

src/main/scala/viper/gobra/util/ViperChopper.scala
Felalolf: Using the ints, I can use arrays instead of maps everywhere.

src/main/scala/viper/gobra/util/ViperChopper.scala
Felalolf: Same reasoning, I push back most ordering operations until after the conversion to ints.

src/main/scala/viper/gobra/util/ViperChopper.scala
Felalolf: I will add comments. For axioms, I am not making the distinction of what depends on the axiom and what the axiom depends on. Instead, I make everything a dependency in both directions. For line 477, you are right.

src/main/scala/viper/gobra/util/ViperChopper.scala
Felalolf: I tested different approaches, where I varried the information inside the vertices and the representation of the nodes. The current one seemed to me to have the best performance in terms of transforming to the node representation and then executing the algorithms on it.

src/main/scala/viper/gobra/util/ViperChopper.scala
Felalolf: do you means the result?

src/main/scala/viper/gobra/util/ViperChopper.scala
Felalolf: I do not know which collection you mean. The only ordered collection is the sorted set. Do you mean that I should use  the normal set instead of a sorted set? The reason why I use the sorted set is simply because set relies on hashes.

src/main/scala/viper/gobra/util/ViperChopper.scala
Felalolf: If there were two nodes that are mutually connected, then, yes, for them it does not matter which one is the root, since both will be members of the same color. However, as you noted, the mutual recursive methods do not show up here due to the indirection over the method spec.

src/main/scala/viper/gobra/util/ViperChopper.scala
ArquintL: I see, so you're using a sequence here such that you do not need to convert e.g. a set to a sequence later on

src/main/scala/viper/gobra/util/ViperChopper.scala
ArquintL: I mean the return type. Why is it `Vector[List[Int]]` instead of e.g. `Set[List[Int]]`?

src/main/scala/viper/gobra/util/ViperChopper.scala
ArquintL: yes

src/main/scala/viper/gobra/util/ViperChopper.scala
ArquintL: Thanks

src/main/scala/viper/gobra/util/ViperChopper.scala
Felalolf: Ah. I did not consider that you mean the Vector. Each of them is considered a separate program. With a set, it would check that all of them are different, which can be quite expensive and is unnecessary.

src/main/scala/viper/gobra/util/ViperChopper.scala
Felalolf: Almost. It is about the vertices. Checking that vertices are distinct is more expensive than checking that integers are distinct. And since I am converting to integers either way, I removing the duplicates also at that point.

src/main/scala/viper/gobra/util/ViperChopper.scala
ArquintL: Makes sense, I did not think about that getting the set properties is not free

PR #385, opened 2022-01-11
Add Lino's stubs to master

jcp19:
This PR adds the changes from PR #382 to the main branch. Lino merged it in the branch `verified-scion-preview-without-selective` which will soon be deprecated.

### DIFF
--- a/src/main/resources/stubs/time/sleep.gobra
+++ b/src/main/resources/stubs/time/sleep.gobra
@@ -1,0 +1,13 @@
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in https://golang.org/LICENSE
+
+// Signatures for the public declarations in file
+// https://github.com/golang/go/blob/master/src/time/sleep.go
+
+package time
+
+// Sleep pauses the current goroutine for at least the duration d.
+// A negative or zero duration causes Sleep to return immediately.
+decreases _ if d < 0
+func Sleep(d Duration)
--- a/src/main/resources/stubs/time/sleep.gobra
+++ b/src/main/resources/stubs/time/sleep.gobra
@@ -9,5 +9,5 @@ package time
 
 // Sleep pauses the current goroutine for at least the duration d.
 // A negative or zero duration causes Sleep to return immediately.
-decreases _ if d < 0
+decreases _ if d <= 0
 func Sleep(d Duration)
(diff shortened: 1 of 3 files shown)


src/main/resources/stubs/time/sleep.gobra
```diff
@@ -0,0 +1,13 @@
...
+// license that can be found in https://golang.org/LICENSE
+
+// Signatures for the public declarations in file
+// https://github.com/golang/go/blob/master/src/time/sleep.go
+
+package time
+
+// Sleep pauses the current goroutine for at least the duration d.
+// A negative or zero duration causes Sleep to return immediately.
+decreases _ if d < 0
```
ArquintL, line 12:
```suggestion
decreases _ if d <= 0
```
Furthermore, we could think about termination even for positive arguments because even a 10sec sleep eventually terminates...

src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
```diff
@@ -292,7 +292,7 @@ class StringEncoding extends LeafTypeEncoding {
         results = Vector(res),
         pres = Vector(pre),
         posts = Vector(),
-        terminationMeasures = Vector(),
+        terminationMeasures = Vector(in.WildcardMeasure(None)(info)),
```
ArquintL, line 295: the comment at the top of this `val` should be updated correspondingly.
BTW, I'm wondering whether we distinguish between `decreases` and `decreases _` for abstract functions/methods. I guess the only difference occurs when you eventually add a body as the latter case makes it easy to forget about the obligation to prove termination

ArquintL (approved):
2 minor comments, can be merged after addressing them

src/main/resources/stubs/time/sleep.gobra
jcp19: Unfortunately, the official documentation does not say anything about `Sleep` terminating eventually. In fact, the documentation says exactly what is said in the comment.

I agree that a non-terminating sleep would be mostly useless but I think we should not provide stronger specs than what is provided in the official docs.

ArquintL: Sorry, you're right. I did not watch out for `at least` and relied too much on my intuition

src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
jcp19: For abstract methods, I think that there is no practical difference. It is true that using `decreases _` makes it easier to forget adding the correct termination measure. At the same time, I prefer using this notation because, unlike the first, it does not assume that it has a constant termination measure - it just assumes that there is some termination measure for this function.

PR #387, opened 2022-01-14
Chopper with scc

Felalolf:
Extension of chopper where (for at least 10 isolated vertices) an additional SCC transformation is used. 
The advantage of the SCC is that it decreases the size of the graph later steps work on, in particular the merging step.
To check correctness, I did ran all tests with the chopper enabled, which succeeded. If you are currious, this took 47 minutes, which is 4 minutes slower than the current time on master.

### DIFF
--- a/src/main/scala/viper/gobra/util/ChopperUtil.scala
+++ b/src/main/scala/viper/gobra/util/ChopperUtil.scala
@@ -6,17 +6,26 @@
 
 package viper.gobra.util
 
+import java.io.File
+import java.util.Properties
+
 import viper.silver.{ast => vpr}
 import viper.silver.ast.SourcePosition
 import viper.gobra.frontend.Config
-import viper.gobra.reporting.{ChoppedViperMessage, Source}
+import viper.gobra.reporting.ChoppedViperMessage
 import viper.gobra.backend.BackendVerifier.Task
 
 object ChopperUtil {
 
   /** Splits task program into multiple Viper programs depending on config. */
   def computeChoppedPrograms(task: Task)(config: Config): Vector[vpr.Program] = {
-    val programs = ViperChopper.chop(task.program)(isolate = computeIsolateMap(config), bound = Some(config.choppingUpperBound))
+
+
+    val programs = ViperChopper.chop(task.program)(
+      isolate = computeIsolateMap(config),
+      bound = Some(config.choppingUpperBound),
+      penalty = getPenalty
+    )
 
     // Report Chopped Programs
     programs.zipWithIndex.foreach{ case (chopped, idx) =>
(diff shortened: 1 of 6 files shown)


src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -663,8 +649,79 @@ object ViperChopper {
...
+      // val checkedCycles = Array.ofDim[Boolean](N)
+      // def noCycle(x: Int): Unit = {
+      //   assert(!onStack(x), "found cycle, but expected none.")
+      //   if (!checkedCycles(x)) {
+      //     checkedCycles(x) = true; onStack(x) = true
+      //     for (y <- csEdges(x)) { noCycle(y.proxy) }
+      //     onStack(x) = false
+      //   }
+      // }
+      // for (idx <- 0 until N) noCycle(idToComponent(idx).proxy)
```
jcp19, line 663: We can enable this check in debug mode

src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -62,27 +62,56 @@ object ViperChopper {
...
         val containedInResult = Array.ofDim[Boolean](N)
         for (program <- result; node <- program) { containedInResult(node) = true }
         // check all nodes of the smallest programs are present (no node should be lost)
-        val smallestProgramContainedInResult = smallestPrograms.forall(_.forall(containedInResult(_)))
+        val smallestProgramContainedInResult = forallSmallNode(containedInResult(_))
         assert(smallestProgramContainedInResult, "Chopper Error: Lost nodes during merging step.")
 
         val containedInSmallest = Array.ofDim[Boolean](N)
-        for (program <- smallestPrograms; node <- program) { containedInSmallest(node) = true }
+        forallSmallNode{node => containedInSmallest(node) = true; true }
```
jcp19, line 114: This looks somewhat hacky, but it is fine, given that this code is supposed to be deleted eventually

jcp19 (approved):

src/main/scala/viper/gobra/util/ChopperUtil.scala
```diff
@@ -6,17 +6,28 @@
...
 import viper.silver.{ast => vpr}
 import viper.silver.ast.SourcePosition
 import viper.gobra.frontend.Config
-import viper.gobra.reporting.{ChoppedViperMessage, Source}
+import viper.gobra.reporting.ChoppedViperMessage
 import viper.gobra.backend.BackendVerifier.Task
 
 object ChopperUtil {
 
+  val GobraChopperFileLocation = "GobraChopper.conf"
```
ArquintL, line 20: could you please add an example conf file somewhere? E.g. at Gobra's root directory?

ArquintL (approved):

PR #388, opened 2022-01-17
ANTLR Parser

nicoberling:
This PR replaces the kiama parser with a new ANTLR4 parser.

It addresses performance issues and fixes multiple issues with the old parser.


jcp19:
For future reference, when reviewing this PR we should go through all issues identified with the tag `#parsing` and check which ones still hold

ArquintL:
@nicoberling Could you please address the conflicts? Hopefully, the CI will then pick this PR afterwards

src/test/resources/unimplemented_go/controlflow/controlflow.go
```diff
@@ -0,0 +1,6 @@
+package controlflow;
+
+ensures res > 0
+func a() (res int) {
+    for {continue}
+}
```
jcp19, line 6: this file should parse but not verify. I think we need an annotation to express that this file is parseOnly

src/test/resources/unimplemented_go/floats/float.go
```diff
@@ -0,0 +1,5 @@
+package main
+
+func main() {
+
+}
```
jcp19, line 5: This file is not testing anything. Besides, floats have been added in Gobra in the meanwhile.

src/test/resources/better_errors/weirdfunct.go
```diff
@@ -0,0 +1,23 @@
+package main
```
jcp19, line 1: what is this file testing?

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
```diff
@@ -136,6 +136,9 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
     case _: PBlock => noMessages
     case _: PSeq => noMessages
     case _: PEmptyStmt => noMessages
+    case _: PGoto => noMessages
+    case _: PContinue => noMessages
+    case _: PBreak => noMessages
```
jcp19, line 141: For `continue` and `break`, we should add checks here. In particular, `continue` statements can only appear in loops and `break` can only appear in loops, switch statements and select statements.

jcp19 (commented):
I made a very quick pass through the PR.
I will provide more detailed feedback when `master` is merged in this branch.

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
Felalolf: I think the idea was that only parser support is added for now and nothing else

jcp19: Sure, but then I would replace `noMessages` with either `???` or a violation such that we do not forget in the future to add this.

Felalolf:
Joao reported that the parser accepts files, which should not be accepted. For instance:
```
package p

dsfasfsafasdf

func {
    const c = 2
}
```

build.sbt
```diff
@@ -42,13 +42,14 @@ lazy val gobra = (project in file("."))
...
     libraryDependencies += "org.apache.commons" % "commons-text" % "1.9", // for escaping strings in parser preprocessor
     libraryDependencies += "commons-codec" % "commons-codec" % "1.15", // for obtaining the hex encoding of a string
+    libraryDependencies += "org.antlr" % "antlr4-runtime" % "4.9.2",
 
     scalacOptions ++= Seq(
       "-encoding", "UTF-8", // Enforce UTF-8, instead of relying on properly set locales
       "-Ypatmat-exhaust-depth", "40"
     ),
 
-    javacOptions ++= Seq("-encoding", "UTF-8", "-charset", "UTF-8", "-docencoding", "UTF-8"),
```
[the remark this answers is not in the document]
nicoberling: Yes, `javac` does not have `-charset` or `-docencoding` options, leaving them in causes an error when compiling .java files.

src/main/java/viper/gobra/frontend/GobraLexer.java
```diff
@@ -0,0 +1,854 @@
+// Generated from /home/nico/Documents/repositories/projects/eth/BA/gobraHome/gobra/src/main/antlr4/GobraLexer.g4 by ANTLR 4.9.2
```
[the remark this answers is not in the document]
nicoberling: I've tried this plugin before, and it doesn't offer enough options to be useful on its own. It also creates a dependency mess, so it's way easier to just generate the parser manually when changing the grammar and use the generated files otherwise. I agree that we should find a way to make this more automatic, maybe by moving the parser into its own sub-project folder in gobra?

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
nicoberling: I've changed it to `???` for now.

src/test/resources/better_errors/weirdfunct.go
nicoberling: I think this was leftover from some experiment, I've removed it for now.

src/test/resources/regressions/features/multisets/multiset-range-fail3.gobra
```diff
@@ -5,5 +5,5 @@ package pkg
 
 func foo() {
   //:: ExpectedOutput(assert_error:assertion_error)
-  assert 42 in mset[1..10]
+  assert 42 in mset[1 .. 10]
```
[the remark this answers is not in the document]
nicoberling: I've considered this quite early, and I've revisited the problem and was finally able to fix the lexer so that it avoids this problem. In case it turns that this has introduced some hidden errors again, we can consider the other options again. Otherwise I would also implement a tailored error message for this.

src/test/resources/unimplemented_go/floats/float.go
nicoberling: It now tests the parser for (non-hex) floats.

nicoberling:
> Joao reported that the parser accepts files, which should not be accepted. For instance:
> 
> ```
> package p
> 
> dsfasfsafasdf
> 
> func {
>     const c = 2
> }
> ```

I can't reproduce this, even on a fresh setup. What is Gobra's overall output in this case?

jcp19:
> > Joao reported that the parser accepts files, which should not be accepted. For instance:
> > ```
> > package p
> > 
> > dsfasfsafasdf
> > 
> > func {
> >     const c = 2
> > }
> > ```
> 
> I can't reproduce this, even on a fresh setup. What is Gobra's overall output in this case?

Running gobra with the `-i` flag on this file, I get the following output:
```
[info] Gobra 1.1-SNAPSHOT (22551b40@(detached))
[info] (c) Copyright ETH Zurich 2012 - 2021
[info] +#RESULT={"filename": "/Users/joao/Desktop/tmp.gobra", "parser": "ANTLR_FULL", "time":0.1501507659559138, "nodes":null}
[info] Gobra found no errors
[info]
```
For some reason, the tests seem to pass in our CI but when I run Gobra locally, it seems to accept every input.

nicoberling:
> > > Joao reported that the parser accepts files, which should not be accepted. For instance:
> > > ```
> > > package p
> > > 
> > > dsfasfsafasdf
> > > 
> > > func {
> > >     const c = 2
> > > }
> > > ```
> > 
> > 
> > I can't reproduce this, even on a fresh setup. What is Gobra's overall output in this case?
> 
> Running gobra with the `-i` flag on this file, I get the following output:
> 
> ```
> [info] Gobra 1.1-SNAPSHOT (22551b40@(detached))
> [info] (c) Copyright ETH Zurich 2012 - 2021
> [info] +#RESULT={"filename": "/Users/joao/Desktop/tmp.gobra", "parser": "ANTLR_FULL", "time":0.1501507659559138, "nodes":null}
> [info] Gobra found no errors
> [info]
> ```
> 
> For some reason, the tests seem to pass in our CI but when I run Gobra locally, it seems to accept every input.

I still got the expected errors when running your input, but I found some other input that behaved as you described. I fixed this and some possibly related issues, I hope this resolves it. One of the issues was empty lines at the end of the input getting treated incorrectly by the error listener, so the differences between our outputs might be because of different line/file terminator conventions in our editors/OSes.

src/test/resources/unimplemented_go/controlflow/controlflow.go
ArquintL: This could be done by adding `// ##(--parseOnly)` e.g. after the package clause. Note however that the `Config.merge` functions need to be adapted to use the disjunction when merging the `should...` flags, e.g. `shouldParse = shouldParse || other.shouldParse`

ArquintL: Alternatively, we could add a `GobraParserTests` unit test class in the `viper.gobra.parsing` package that correctly configures Gobra to only do parsing and not more with a dedicated set of input files. It should be possible that this unit test class extends `AbstractGobraTests` to reuse most code used in e.g. `GobraTests`

jcp19:
I'm testing this branch of Gobra in VerifiedSCION, a large project that uses Gobra for code level verification. I noticed that some files that could be parsed before cannot be parsed now. For example, verifying the files [pkg/router/dataplane.gobra](https://github.com/jcp19/VerifiedSCION/blob/master/gobra/pkg/router/dataplane.gobra) and [pkg/router/metrics.gobra](https://github.com/jcp19/VerifiedSCION/blob/master/gobra/pkg/router/metrics.gobra) together would parse in the previous version, but not in the new one.

To reproduce this error, I run the following command in sbt
```
run -i PATH_TO_PROJECT/gobra/pkg/router/metrics.gobra PATH_TO_PROJECT/gobra/pkg/router/dataplane.gobra -I PATH_TO_PROJECT --parseOnly
```
where `PATH_TO_PROJECT` stands for the path to the local copy of the verifiedSCION repository.

With the new parser, I get the following errors:
```
[info] +#RESULT={"filename": "/Users/joao/Code/verifiedScionGo/gobra/pkg/router/metrics.gobra", "parser": "ANTLR_PARSE_SLL", "time":0.2456224259804003, "nodes":null}
[info] +#RESULT={"filename": "/Users/joao/Code/verifiedScionGo/gobra/pkg/router/metrics.gobra", "parser": "ANTLR_TRANSLATE", "time":0.12604055501287803, "nodes":null}
[info] +#RESULT={"filename": "/Users/joao/Code/verifiedScionGo/gobra/pkg/router/dataplane.gobra", "parser": "ANTLR_PARSE_LL", "time":4.213293635984883, "nodes":null}
[info] +#RESULT={"filename": "/Users/joao/Code/verifiedScionGo/gobra/pkg/router/metrics.gobra, /Users/joao/Code/verifiedScionGo/gobra/pkg/router/dataplane.gobra", "parser": "ANTLR_FULL", "time":5.560689482954331, "nodes":null}
[info] Gobra has found 12 error(s):
[info] 	<dataplane.gobra:787:52> unexpected {, expecting boundVariableDecl
[info] 	if _, exists := (d.internalNextHops)[ifID]; exists {
[info]                                                     ^
[info] 	<dataplane.gobra:787:43> unexpected ;, expecting assign_op
[info] 	if _, exists := (d.internalNextHops)[ifID]; exists {
[info]                                            ^
[info] 	<dataplane.gobra:787:14> Unexpected ':=', did you mean '='?
[info] 	if _, exists := (d.internalNextHops)[ifID]; exists {
[info]               ^^
[info] 	<dataplane.gobra:581:45> unexpected {, expecting boundVariableDecl
[info] 	if _, exists := (d.linkTypes)[ifID]; exists {
[info]                                              ^
[info] 	<dataplane.gobra:581:36> unexpected ;, expecting assign_op
[info] 	if _, exists := (d.linkTypes)[ifID]; exists {
[info]                                     ^
[info] 	<dataplane.gobra:581:14> Unexpected ':=', did you mean '='?
[info] 	if _, exists := (d.linkTypes)[ifID]; exists {
[info]               ^^
[info] 	<dataplane.gobra:552:47> unexpected {, expecting boundVariableDecl
[info] 	if _, exists := (d.neighborIAs)[ifID]; exists {
[info]                                                ^
[info] 	<dataplane.gobra:552:38> unexpected ;, expecting assign_op
[info] 	if _, exists := (d.neighborIAs)[ifID]; exists {
[info]                                       ^
[info] 	<dataplane.gobra:552:14> Unexpected ':=', did you mean '='?
[info] 	if _, exists := (d.neighborIAs)[ifID]; exists {
[info]               ^^
[info] 	<dataplane.gobra:501:44> unexpected {, expecting boundVariableDecl
[info] 	if _, exists := (d.external)[ifID]; exists {
[info]                                             ^
[info] 	<dataplane.gobra:501:35> unexpected ;, expecting assign_op
[info] 	if _, exists := (d.external)[ifID]; exists {
[info]                                    ^
[info] 	<dataplane.gobra:501:14> Unexpected ':=', did you mean '='?
[info] 	if _, exists := (d.external)[ifID]; exists {
[info]               ^^
```

nicoberling:
@jcp19 Since ANTLR is a lexing parser, the keywords used in Gobra language constructs like `forall`, `exists` are now all reserved. This also applies to some other Gobra keywords that didn't use to be reserved, like the pre/postcondition keywords, `apply`, `domain` and some others. I'm not sure if there is a reasonable way to work around this in the parser, but I'll at least make sure that you get errors that explain the problem.

nicoberling:
@jcp19 The parser doesn't seem to induce semicolons correctly all of the time, so I enabled the semicolon preprocessor for now, this should also fix your issue.

jcp19:
> @jcp19 The parser doesn't seem to induce semicolons correctly all of the time, so I enabled the semicolon preprocessor for now, this should also fix your issue.

Seems to work now, thanks!

src/test/resources/regressions/features/sequences/seq-contains-simple1.gobra
```diff
@@ -38,11 +38,11 @@ func example6(x int, ghost xs seq[int], ghost ys seq[bool]) {
 }
 
 requires 0 <= n
-ensures !(n in seq[0..n])
```
ArquintL, line 41: seems to occur in many files. I'm fine with having no spaces in some files such that both ways are tested. But in the interest of keeping the diff as small as possible, I would revert it in most files

src/test/resources/regressions/features/trusted/trusted-functions.gobra
```diff
@@ -25,8 +25,18 @@ func positiveSumWithoutTrusted(x int, y int) (ret int) {
 
 trusted
 func functionWithNonSupportedFeature() int{
-	some non supported go feature
-	return 0
+	// fallthrough statements would normally cause a translation error
+	// since they aren't supported
```
ArquintL, line 29: I would rely here on some feature that Gobra does not (yet) support. Couldn't you write a body that would cause a type error if it would be type checked? E.g. assigning something to an incompatible type

src/test/scala/viper/gobra/GobraParserTests.scala
```diff
@@ -0,0 +1,89 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2011-2020 ETH Zurich.
+
+package viper.gobra
```
ArquintL, line 7: I would place this in the `parsing` sub-package

src/test/scala/viper/gobra/GobraParserTests.scala
```diff
@@ -0,0 +1,89 @@
...
+import java.nio.file.Path
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+
+class GobraParserTests extends AbstractGobraTests with BeforeAndAfterAll {
+
+  val regressionsPropertyName = "GOBRATESTS_REGRESSIONS_DIR"
+
+  val regressionsDir: String = System.getProperty(regressionsPropertyName, "better_errors")
+  val testDirectories: Seq[String] = Vector(regressionsDir)
```
ArquintL, line 28: You can directly use the specified directory without the option to allow users to customize it

```suggestion
 val testDirectories: Seq[String] = Vector("better_errors")
```

ArquintL: What's about the `unimplemented_go` folder? What kind of test cases are stored in there?

src/test/scala/viper/gobra/parsing/GoParserUnitTests.scala
```diff
@@ -0,0 +1,197 @@
...
+    test(source) {
+      frontend.parseProgram(source) should be
+    }
+  }
+
+
+
+  /* ** Stubs, mocks and other test setup */
+
+  class TestFrontend {
```
ArquintL, line 176: this seems to be copy pasted from the existing parser unit tests. Please move them to a separate file such that both tests can make use of them without code duplication

src/test/scala/viper/gobra/parsing/GoParserUnitTests.scala
```diff
@@ -0,0 +1,197 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2011-2020 ETH Zurich.
```
ArquintL, line 5: I would use here the license of the Go tests as that's the valuable part of this file

src/test/scala/viper/gobra/parsing/ParserUnitTests.scala
```diff
@@ -364,7 +364,7 @@ class ParserUnitTests extends AnyFunSuite with Matchers with Inside {
   }
 
   test("Parser: expressions of the form '++e' should be parsed to '0 + 0 + e'") {
```
ArquintL, line 366: description does not match the test case

src/test/scala/viper/gobra/parsing/ParserUnitTests.scala
```diff
@@ -804,12 +810,27 @@ class ParserUnitTests extends AnyFunSuite with Matchers with Inside {
     }
   }
 
+  test("Parser: should parse indexed expression with sequence range expressions with spaces") {
```
ArquintL, line 813: same here

src/test/scala/viper/gobra/parsing/ParserUnitTests.scala
```diff
@@ -804,12 +810,27 @@ class ParserUnitTests extends AnyFunSuite with Matchers with Inside {
...
+    }
+  }
+
   test("Parser: shouldn't parse a chain of sequence range operations") {
     frontend.parseExp("seq[1..10][11..20]") should matchPattern {
       case Left(_) =>
     }
   }
 
+  test("Parser: shouldn't parse a chain of sequence range operations with spaces") {
```
ArquintL, line 828: same here

ArquintL (changes requested):
I've mainly many small comments. Some bigger ones involve the grammar as well as the visitor

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -65,7 +89,109 @@ object Parser {
...
+      pkg <- makePackage(programs)
+    } yield pkg
+    // report potential errors:
+    res.left.map(errors => {
+      val groupedErrors = errors.groupBy{ _.position.get.file }
+      groupedErrors.foreach { case (p, pErrors) =>
+        config.reporter report ParserErrorMessage(p, pErrors)
+      }
+      errors
+    })
```
Felalolf, line 191: I think this should be something along the lines of:
```suggestion
    res.left.foreach(errors => {
      val groupedErrors = errors.groupBy{ _.position.get.file }
      groupedErrors.foreach { case (p, pErrors) =>
        config.reporter report ParserErrorMessage(p, pErrors)
      }
    })
    
    res
```

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -185,6 +185,8 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
 
     case n: PIntLit => numExprWithinTypeBounds(n)
 
+    case _: PFloatLit => ???
```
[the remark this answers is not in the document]
jcp19: No, I said in the PR that float literals would be left to another PR. The reason is that there was no reason to duplicate work by extending the old parser to parse float literals.

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -208,6 +209,13 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
     noshort = true
   )
 
+  val legacyParser: ScallopOption[Boolean] = toggle(
+    name = "legacyParser",
+    descrYes = "Use the old Parser Combinator",
+    default = Some(false),
+    noshort = true
+  )
```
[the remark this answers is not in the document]
jcp19: Agree. I will test the new parser in the entirety of SCION's codebase in the next few days.

src/main/java/viper/gobra/frontend/GobraLexer.java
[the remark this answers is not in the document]
jcp19: What options is this plugin missing to make it usable here? 

Also, what do you mean by dependency mess? Does this plugin assume a different version of sbt or is it something else?

Are there any other alternatives? could we for example configure sbt such that a script is run when we execute `sbt compile`?

If we do not manage to automate it, can we at least configure ANTLR to generate files with valid license headers, and without any reference to a specific user's filesystem?

build.sbt
```diff
@@ -57,6 +58,7 @@ lazy val gobra = (project in file("."))
     ),
 
     fork := true,
+    cancelable in Global := true,
```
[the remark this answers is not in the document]
nicoberling: It just allows you to cancel running tasks in sbt without shutting down sbt completely. Not neccessary at all, but I found it useful.

src/main/antlr4/GobraLexer.g4
```diff
@@ -0,0 +1,101 @@
+lexer grammar GobraLexer;
+import GoLexer;
+
+FLOAT_LIT : DECIMAL_FLOAT_LIT | HEX_FLOAT_LIT;
+
+DECIMAL_FLOAT_LIT      : DECIMALS ('.'{_input.LA(1) != '.'}? DECIMALS? EXPONENT? | EXPONENT)
+                       | '.'{_input.index() <2 || _input.LA(-2) != '.'}? DECIMALS EXPONENT?
+                       ;
```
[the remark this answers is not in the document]
nicoberling: This is the same as the GoParser.g4 version, except for the lookahead/behind to check for the range case (..).

src/main/antlr4/GobraParser.g4
```diff
@@ -0,0 +1,428 @@
...
+    | expression
+    ;
+
+range: kind=(SEQ | SET | MSET) L_BRACKET expression DOT_DOT expression R_BRACKET;
+
+// Changed Rules
+
+// Ghost members
+
+sourceFile:
```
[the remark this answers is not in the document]
nicoberling: I'll group them more reasonably.

src/main/antlr4/GobraParser.g4
```diff
@@ -0,0 +1,428 @@
...
+typeLit:
+	arrayType
+	| structType
+	| pointerType
+	| functionType
+	| interfaceType
+	| sliceType
+	| mapType
+	| channelType
+	| predType;
```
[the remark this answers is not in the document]
nicoberling: I don't quite understand what you mean, in Gobra type literals can also be predTypes, so the typeLit rule must always include these, right? For your second question, ANTLR inheritance works like OO inheritance, rules from the parent grammar get copied over unless they are overridden.

src/main/antlr4/GobraParser.g4
```diff
@@ -0,0 +1,428 @@
...
+	L_BRACKET (
+		low? COLON high?
+		| low? COLON high COLON cap
+	) R_BRACKET;
+
+low : expression;
+high: expression;
+cap: expression;
+
+// allow "if ; true {}"
```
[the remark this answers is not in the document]
nicoberling: This was actually an error in the GoParser.g4, but I've since fixed it in the original repo, so this should move back to the GoParser.g4, thanks for noticing.

src/main/antlr4/GobraParser.g4
```diff
@@ -0,0 +1,428 @@
...
+		| STAR
+		| DIV
+		| MOD
+		| LSHIFT
+		| RSHIFT
+		| AMPERSAND
+		| BIT_CLEAR
+	)? ASSIGN;
+
+// allow "import ("import1";"import2") without semicolon at the end
```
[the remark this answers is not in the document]
nicoberling: Same as before, this is an error in the original grammar that I've since fixed upstream.

src/main/scala/viper/gobra/frontend/InformativeErrorListener.scala
```diff
@@ -0,0 +1,225 @@
...
+import viper.gobra.frontend.Source.FromFileSource
+import viper.gobra.reporting.ParserError
+import viper.silver.ast.SourcePosition
+
+import java.nio.file.Path
+import scala.collection.mutable.ListBuffer
+
+class InformativeErrorListener(val messages: ListBuffer[ParserError], val source: Source) extends BaseErrorListener{
+
+  var ignoredImplementationProofError : Option[ParserError] = None
```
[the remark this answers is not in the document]
nicoberling: I think this is a remnant from a previous version, I'll remove it.

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
```diff
@@ -0,0 +1,2348 @@
...
+
+
+  /**
+    * Visit the rule
+    * typeOnly: type_ EOF;
+
+    */
+  override def visitTypeOnly(ctx: TypeOnlyContext): PType = visitType_(ctx.type_())
+
+  def translate[Rule <: ParserRuleContext, Node](tree: Rule):  Node = {
```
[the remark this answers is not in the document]
nicoberling: This is the type of the result, we could also change this to PNode and then have a match statement or .asInstanceOf[<whatever node you need>] wherever the function is called. I'll look into making this more idiomatic.

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
```diff
@@ -0,0 +1,2348 @@
...
+    *
+    * <p>The default implementation returns the result of calling
+    * {@link #visitChildren} on {@code ctx}.</p>
+    */
+  override def visitFpredicateDecl(ctx: FpredicateDeclContext): PFPredicateDecl = {
+    //val id_ = withWildcards{
+    //  idnDef(ctx.IDENTIFIER()).at(ctx.IDENTIFIER())
+    //}
+    // Go allows _ here
+    val id = visitIdentifier(ctx.IDENTIFIER(), Disallow, PIdnDef)
```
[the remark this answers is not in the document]
nicoberling: I don't know why the comment is here in this case, but often, Go allows the blank identifier but Gobra does not. Often it makes sense to just discard the entire function for example, and here wildcards don't really make sense, I'll look into making the comments more consistent.

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
```diff
@@ -0,0 +1,2348 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
nicoberling: I agree that we should group them more reasonably, I don't think function overrides can be split into multiple files. When it comes to the visit... functions, there are three main patterns:

1.  Check with an if statement in which rule alternative we are, and then just call visit...() or visitNode[P...]() on each of the constituents to assemble the resulting PNode.
2. Call visitChildren, which returns a Vector of the rule constituents and then match this to find out which alternative is chosen and construct the resulting PNode. (equivalent to 1, just more practical in some cases)
3. call visitNodeChild(), which is just a convenience function for rules that are just "collections" of single rules.

These should be easy to maintain, since that just involves adding a new match- or if-case. ANTLR also allows you to name main alternatives, this should probably be used for some of the rules with huge if-statement chains. The main problem is that there are indeed a lot of edge cases that don't fall into these three, usually when the Go grammar and the Gobra AST aren't quite compatible, and I agree that these should be commented thoroughly to make them maintainable.

src/main/antlr4/GobraParser.g4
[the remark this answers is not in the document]
Felalolf: We could maybe mark the added cases with a comment, like `// new`

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -40,15 +47,32 @@ object Parser {
...
+      val preprocessedSources = input
+        .map{ Gobrafier.gobrafy }
+        .map{ source => SemicolonPreprocessor.preprocess(source)(config) }
+      for {
+        parseAst <- time("GOBRA", input(0).name) {parseSources(preprocessedSources, specOnly)(config)}
+        postprocessedAst <- new ImportPostprocessor(parseAst.positions.positions).postprocess(parseAst)(config)
+      } yield postprocessedAst
+    } else {
+      for {
+        parseAst <- time("ANTLR_FULL", sources.map(_.name).mkString(", ")) {antlrParseSources(sources, specOnly)(config)}
```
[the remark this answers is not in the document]
nicoberling: Yes, this is a change I just made today to address an issue with the induced semicolons.

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -40,15 +47,32 @@ object Parser {
...
 
   def parse(input: Vector[Source], specOnly: Boolean = false)(config: Config): Either[Vector[VerifierError], PPackage] = {
-    val preprocessedSources = input
-      .map{ Gobrafier.gobrafy }
-      .map{ source => SemicolonPreprocessor.preprocess(source)(config) }
-    for {
-      parseAst <- parseSources(preprocessedSources, specOnly)(config)
-      postprocessedAst <- new ImportPostprocessor(parseAst.positions.positions).postprocess(parseAst)(config)
-    } yield postprocessedAst
+    val sources = input.map(Gobrafier.gobrafy).map(i => SemicolonPreprocessor.preprocess(i)(config))
```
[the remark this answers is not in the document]
nicoberling: This entire section will probably change before actually merging to completely remove the old parser, it's just very useful to have access to the old parser during the development/merge process.

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -65,7 +89,109 @@ object Parser {
     sourceCache = Map.empty
   }
 
-  private def parseSources(sources: Vector[Source], specOnly: Boolean)(config: Config): Either[Vector[ParserError], PPackage] = {
+  private def antlrParseSources(sources: Vector[Source], specOnly: Boolean)(config: Config): Either[Vector[VerifierError], PPackage] = {
```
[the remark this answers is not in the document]
nicoberling: Yeah, I'll try to reduce the amount of old code.

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -307,6 +433,115 @@ object Parser {
...
+
+    def parse(rule : => Rule): Either[Vector[ParserError], Node] = {
+      val name = source.name
+      val tree = try time ("ANTLR_PARSE_SLL", name) { rule }
+      catch {
+        case _: AmbiguityException  => parse_LL(rule) // Resolve `<IDENTIFIER> { }` ambiguities in switch/if-statements
+        case _: ParseCancellationException => parse_LL(rule) // For even faster parsing, replace with `new ParserRuleContext()`.
+        case e => errors.append(ParserError(e.getMessage, Some(SourcePosition(source.toPath, 0, 0)))); new ParserRuleContext()
+      }
+      //println(tokens.getTokens.asScala.map(tok => VOCABULARY.getSymbolicName(tok.getType)))
```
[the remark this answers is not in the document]
nicoberling: Sure, sorry.

src/test/resources/regressions/features/multisets/multiset-range-fail1.gobra
```diff
@@ -5,5 +5,5 @@ package pkg
 
 func foo() {
   //:: ExpectedOutput(type_error)
-  assert mset[1 .. 4] == set[1 .. 4]
```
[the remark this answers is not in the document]
nicoberling: I think this is a case where the original test file had the spaces already(git blame says since 2 years). I accidentally removed them because I just ran a RegEx over the whole folder to remove the spaces I added, but I can try to restore test files that already had the spaces even before I added them to all cases.

src/test/resources/regressions/features/termination/termination-fail-02.gobra
```diff
@@ -3,6 +3,7 @@
 
 package termination
 
+
```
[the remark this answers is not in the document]
nicoberling: This is the weird bug I mentioned a couple weeks ago, I'll open an issue. The file with the added newline doesn't verify with the old parser, while the file with just one newline doesn't verify with the new parser, so this is a bug during verification, not during parsing.

src/test/scala/viper/gobra/GobraParserTests.scala
nicoberling: They include things that the parser can parse, but either don't have a representation in the Gobra AST or fail at some later stage.

src/test/scala/viper/gobra/parsing/ParserUnitTests.scala
```diff
@@ -804,12 +810,27 @@ class ParserUnitTests extends AnyFunSuite with Matchers with Inside {
...
+    frontend.parseExp("seq[1..10][11..20]") should matchPattern {
+      case Left(_) =>
+    }
+  }
```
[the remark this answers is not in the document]
nicoberling: This is from when I required spaces for the range cases, now it's just a duplicate of the previous testcase that was already there from the beginning. `seq[1..10][11..20]` isn't valid Gobra as far as I know.

src/main/antlr4/GobraParser.g4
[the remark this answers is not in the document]
nicoberling: The rules are already split into new rules that are unique to gobra (after `// New rules`), and rules that changed (after `// Changed rules`) and have comments describing what's changed, it's possible that one or two slipped through the cracks though, I'll check.

src/main/java/viper/gobra/frontend/GobraParserBase.java
```diff
@@ -0,0 +1,130 @@
+package viper.gobra.frontend;
+import java.util.List;
+import org.antlr.v4.runtime.*;
+import viper.gobra.frontend.GobraLexer;
+
+/**
+ * All parser methods that used in grammar (p, prev, notLineTerminator, etc.)
+ * should start with lower case char similar to parser rules.
+ */
+public abstract class GobraParserBase extends org.antlr.v4.runtime.Parser
```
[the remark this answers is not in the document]
nicoberling: This is code used to handle line separators in the parser and is unique to the Go grammar. I'll add a comment that links to the original repo.
It's not correct though, I'm also working on fixing it, contributing that to upstream and then update it here after it's merged, but I think this has quite low priority since the old SemicolonPreprocessor fixes this too.

src/main/antlr4/GobraLexer.g4
```diff
@@ -0,0 +1,101 @@
...
+PRESERVES   : 'preserves';
+POST        : 'ensures';
+INV         : 'invariant';
+DEC         : 'decreases';
+PURE        : 'pure';
+IMPL        : 'implements';
+OLD         : 'old';
+LHS         : '#lhs';
+FORALL      : 'forall';
+EXISTS      : 'exists';
```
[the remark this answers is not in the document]
nicoberling: It could be circumvented by replacing all occurences of `IDENTIFIER` by a new rule 
```
identifier: OLD | EXISTS | ... | IDENTIFIER;
```
This could cause a performance hit because it might cause most files to have to be parsed with the slower, more powerful LL parser instead of the faster SLL, but I'm not sure.

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -111,6 +111,8 @@ sealed trait PGhostifiable extends PNode
 
 sealed trait PMember extends PNode
 
+sealed trait PDeclaration extends PNode with PMember
```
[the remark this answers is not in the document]
nicoberling: It's supposed to more closely resemble what's called a declaration in the Go specification. It doesn't seem to be neccessary though.

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -463,6 +471,7 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
     checkOverflows = checkOverflows(),
     int32bit = int32Bit(),
     shouldVerify = shouldVerify,
-    shouldChop = shouldChop
+    shouldChop = shouldChop,
+    legacyParser = legacyParser.getOrElse(false)
```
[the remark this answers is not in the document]
nicoberling: That makes sense, thanks.

src/main/antlr4/GobraLexer.g4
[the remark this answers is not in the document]
Felalolf: Please add this as a comment

src/main/scala/viper/gobra/frontend/Config.scala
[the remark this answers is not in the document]
jcp19: After a round of tests, the new parser seems to work fine with the verified scion codebase

Felalolf:
can you add to termination-fail-02.gobra the right number of empty lines so that it does not fail.

nicoberling:
> can you add to termination-fail-02.gobra the right number of empty lines so that it does not fail.

Because I do not know what factors the right number depends on (the version that failed CI in this pull request consistently works on my machine, using the same version of Z3 etc), I've decided to split the test case into two parts and ignoring the problematic parts for now.

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
```diff
@@ -0,0 +1,2271 @@
...
+  //endregion
+
+  //region Assignments
+  /**
+    * Visit a parse tree produced by `GobraParser`.
+    *
+    * @param ctx the parse tree
+    * @return the visitor result
+    */
+  // TODO : Refactor easy
```
ArquintL, line 1554: there are still a couple of refactoring TODOs in this file. What do you plan to do with them? Should they still be addressed as part of this PR or do you want to do them after merging?

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
```diff
@@ -0,0 +1,2271 @@
...
+  }
+
+  implicit class PositionableToken[T <: Token](tok: T) extends NamedPositionable {
+    val startPos : Position = Position(tok.getLine, tok.getCharPositionInLine + 1, source)
+    val endPos : Position = Position(tok.getLine, tok.getCharPositionInLine + 1 +(tok.getStopIndex -tok.getStartIndex), source)
+    val name : String = GobraParser.VOCABULARY.getDisplayName(tok.getType)
+    val text : String = tok.getText
+  }
+
+  implicit class PositionedPNodeOption[O <: Some[PNode]](some: O) {
```
ArquintL, line 2188: I'm a bit skeptical seeing three highly similar implicit classes. Couldn't we make them more generic such that we only need one of them?

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
```diff
@@ -0,0 +1,2271 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2011-2020 ETH Zurich.
```
ArquintL, line 5: same here

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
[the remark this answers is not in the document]
ArquintL: Overall, this file became much more readable. I had the impression that I could quickly figure out what each function is supposed to do and does. Thanks a lot!

src/main/scala/viper/gobra/frontend/TranslationHelpers.scala
```diff
@@ -0,0 +1,56 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2011-2020 ETH Zurich.
```
ArquintL, line 5: same here

src/main/scala/viper/gobra/frontend/info/base/SymbolTable.scala
```diff
@@ -253,6 +253,9 @@ object SymbolTable extends Environments[Entity] {
   case class BuiltInFPredicate(tag: BuiltInFPredicateTag, rep: PNode, context: ExternalTypeInfo) extends BuiltInPredicate
   case class BuiltInMPredicate(tag: BuiltInMPredicateTag, rep: PNode, context: ExternalTypeInfo) extends BuiltInPredicate with BuiltInMethodLike with GhostTypeMember
 
+  case class BuiltInType(tag: BuiltInTypeTag, rep: PNode, context: ExternalTypeInfo) extends BuiltInActualEntity with TypeEntity {
+    def isPure: Boolean = true
```
ArquintL, line 257: `pure` in the context of a type is weird. Do we need that?

src/main/scala/viper/gobra/frontend/info/implementation/property/UnderlyingType.scala
```diff
@@ -35,6 +36,11 @@ trait UnderlyingType { this: TypeInfoImpl =>
       case PNamedOperand(t) => entity(t) match {
         case st.NamedType(decl, _, ctx) => inCtx(ctx, decl.right)
         case st.TypeAlias(decl, _, ctx) => inCtx(ctx, decl.right)
+        // TODO: Get the type directly from BuiltInType or BuiltInTypeTag
```
ArquintL, line 39: address this TODO. `BuiltInTypeTag` already has a `typ` field

src/main/scala/viper/gobra/frontend/info/implementation/resolution/AmbiguityResolution.scala
```diff
@@ -49,7 +49,9 @@ trait AmbiguityResolution { this: TypeInfoImpl =>
   def resolve(n: PExpressionOrType): Option[ap.Pattern] = n match {
 
     case n: PNamedOperand =>
-      entity(n.id) match {
+      { val e = entity(n.id)
```
ArquintL, line 52: what's the change here? indentation looks off and code style looks weird

src/main/scala/viper/gobra/frontend/info/implementation/resolution/AmbiguityResolution.scala
```diff
@@ -59,12 +61,13 @@ trait AmbiguityResolution { this: TypeInfoImpl =>
...
         case s: st.BuiltInFunction => Some(ap.BuiltInFunction(n.id, s))
         case s: st.BuiltInFPredicate => Some(ap.BuiltInPredicate(n.id, s))
+        case s : st.BuiltInType => Some(ap.BuiltInType(n.id, s))
         // interface method and predicate when referenced inside of the interface definition
         // (otherwise a receiver would be present)
         case s: st.MethodSpec => Some(ap.ImplicitlyReceivedInterfaceMethod(n.id, s))
         case s: st.MPredicateSpec => Some(ap.ImplicitlyReceivedInterfacePredicate(n.id, s))
-        case _ => None
-      }
+        case s => None
```
ArquintL, line 69: isn't `s` unused?

src/main/scala/viper/gobra/frontend/info/implementation/resolution/AmbiguityResolution.scala
```diff
@@ -103,6 +106,9 @@ trait AmbiguityResolution { this: TypeInfoImpl =>
         case Right(t) if n.args.length == 1 => Some(ap.Conversion(t, n.args.head))
         case Left(e) =>
           resolve(e) match {
+            case Some(p: ap.BuiltInType) if n.args.length == 1 =>
+              // TODO: More elegenat version to get the Types
```
ArquintL, line 110: same here

src/main/scala/viper/gobra/frontend/info/implementation/resolution/AmbiguityResolution.scala
```diff
@@ -122,5 +128,25 @@ trait AmbiguityResolution { this: TypeInfoImpl =>
     case _ => None
   }
 
+  val predefinedTypesMap = Map(
```
ArquintL, line 131: can be removed, I guess

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -764,11 +782,13 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
           // then p can either occur as the base or as an argument of `inv`. However, a numeric expression is not a valid base of a
           // PInvoke and thus, `expr` can onlu appear in `n` as an argument
           lazy val errorMessage = s"violation of assumption: a numeric expression $expr does not occur as an argument of its parent $n"
-          resolve(n) match {
+          val r = resolve(n)
```
ArquintL, line 785: this and the next 3 changes seem unnecessary

src/main/scala/viper/gobra/frontend/info/implementation/typing/IdTyping.scala
```diff
@@ -128,10 +128,12 @@ trait IdTyping extends BaseTyping { this: TypeInfoImpl =>
   }
 
   lazy val idSymType: Typing[PIdnNode] = createTyping { id =>
-    entity(id) match {
+    val e = entity(id)
```
ArquintL, line 131: same here

src/test/resources/regressions/examples/evaluation/impl_errors/parallel_search_replace.gobra
```diff
@@ -28,13 +28,13 @@ pred messagePerm(ghost wg *sync.WaitGroup, s []int, ghost x, y int) {
 }
 
 requires acc(c.RecvChannel(),_)
-requires c.RecvGivenPerm() == PredTrue!<!>;
-requires c.RecvGotPerm() == messagePerm!<wg,_,x,y!>;
+requires c.RecvGivenPerm() == PredTrue!<!>
```
ArquintL, line 31: awesome, thanks for adapting the tests!

src/test/resources/regressions/features/runes/rune_values.gobra
```diff
@@ -0,0 +1,18 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package runes
+
+func main() {
+	assert 'a' == 97
+	assert 'ä' == 228
+	assert '本' == 26412
```
ArquintL, line 9: indentation

src/test/resources/regressions/features/termination/termination-fail-03.gobra
```diff
@@ -0,0 +1,66 @@
...
+	infinite(n int) int
+}
+
+type BadFiniteImpl struct {}
+
+func (y BadFiniteImpl) infinite(n int) int {
+	return y.infinite(n)
+}
+
+type Func interface {
```
ArquintL, line 62: I think this is a copy-paste error: this interface is not used in this file

src/test/resources/regressions/features/trusted/trusted-functions.gobra
```diff
@@ -25,8 +25,8 @@ func positiveSumWithoutTrusted(x int, y int) (ret int) {
 
 trusted
 func functionWithNonSupportedFeature() int{
-	some non supported go feature
-	return 0
+	var i int = "not an integer"
```
ArquintL, line 28: please add a comment that the body is parsed but not type-checked due to the `trusted` annotation (as opposed to not even parsed with the old parser)

src/test/resources/regressions/features/trusted/trusted-methods.gobra
```diff
@@ -28,8 +28,7 @@ func test1() (ret IntType){
 
 trusted
 func (x IntType) methodWithNonSupportedFeature() IntType{
-	some non supported go feature
-	return 0
+	    var j int = "not an int"
```
ArquintL, line 31: indentation looks off

ArquintL (commented):
Looks already way better! Thanks a lot. I still have a few minor comments but I think that they can be addressed quite quickly.

One slightly more general comment to the license headers: please use `2022` in the header of all files that have been newly added by you. In the other files you've only edited, you can leader the header as is.
I know that the conversations get messy. But try to look at all comments, address them, and resolve them.

src/main/java/viper/gobra/frontend/LICENSE
```diff
@@ -0,0 +1,5 @@
+This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
nicoberling: This is here because the generated parser won't have the license header. The Mozilla license allows placing license file in the same directory for cases like this.

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -740,13 +752,13 @@ case class PUnnamedReceiver(typ: PMethodRecvType) extends PReceiver
 case class PResult(outs: Vector[PParameter]) extends PNode with PActualMisc
 
 sealed trait PEmbeddedType extends PNode with PActualMisc {
-  def typ: PNamedType
+  def typ: PUnqualifiedTypeName
   def name: String = typ.name
 }
 
-case class PEmbeddedName(typ: PNamedType) extends PEmbeddedType
+case class PEmbeddedName(typ: PUnqualifiedTypeName) extends PEmbeddedType
```
Felalolf, line 759: I know we discussed this already. But could you please remind me what the difference between PTypeName and PUnqualfiedTypeName is?

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -548,11 +551,20 @@ sealed trait PActualType extends PType
...
-sealed abstract class PPredeclaredType(override val name: String) extends PNamedType
+sealed trait PUnqualifiedTypeName extends PTypeName
+
+object PUnqualifiedTypeName {
+  def unapply(arg: PUnqualifiedTypeName): Option[String] = Some(arg.name)
+}
+
+sealed abstract class PPredeclaredType(override val name: String) extends PUnqualifiedTypeName with PUseLikeId {
+  override def id: PUseLikeId = this
+}
```
Felalolf, line 567: could you please add a short description about when to use which and what they represent.

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -40,15 +43,22 @@ object Parser {
...
   }
 
+  private def time[R](parser : String, filename : String)(block: => R): R = {
+    val t0 = System.nanoTime()/1000000000.0
+    val result = block    // call-by-name
+    val t1 = System.nanoTime()/1000000000.0
+    //println("+#RESULT={\"filename\": \""+ filename + "\", \"parser\": \"" + parser + "\", \"time\":" + (t1 - t0)
+    //  + ", \"nodes\":null}")
+    result
+}
```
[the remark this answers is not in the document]
Felalolf: we cannot merge it until this is removed.

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -65,7 +89,109 @@ object Parser {
...
+
+    def parseSource(source: Source): Either[Vector[ParserError], PProgram] = {
+      val errors = ListBuffer.empty[ParserError]
+      val parser = new antlrSyntaxAnalyzer[SourceFileContext, PProgram](source, errors, pom, specOnly)
+      parser.parse(parser.sourceFile) match {
+        case Right(ast) =>
+          config.reporter report ParsedInputMessage(source.name, () => ast)
+          Right(ast)
+        case Left(errors) =>
+          // On parse failure, ANTLR sometimes throws out of bounds exceptions, ignore these for now.
```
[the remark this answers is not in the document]
Felalolf: what is the status on this?

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -307,1232 +262,96 @@ object Parser {
...
+      // full now with full LL(*)
+      getInterpreter.setPredictionMode(PredictionMode.LL)
+
+      val res = try time ("ANTLR_PARSE_LL", source.name) { rule }
+      catch {
+        case e : Exception => errors.append(ParserError(e.getMessage, Some(SourcePosition(source.toPath, 0, 0))))
+          new ParserRuleContext()
+      }
+      if (ll_errors.isEmpty) errors.clear()
+      res
```
Felalolf, line 308: is this standart or is anything special happening here? If it is just standard, then please add this as a comment.

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -307,1232 +262,96 @@ object Parser {
...
+              return Left(Vector(ParserError(e.getMessage + " " + e.getStackTrace.take(4).mkString("Array(", ", ", ")"), pos)))
           }
         }
-
-      val unnamedParam =
-        "ghost" ~> "...".? ~ typ ^^ {
-          case variadicOpt ~ t =>
-            val typ = if (variadicOpt.isDefined) PVariadicType(t) else t
-            Vector(PExplicitGhostParameter(PUnnamedParameter(typ).at(t)).at(t))
+        // TODO : Properly handle parser warnings
```
Felalolf, line 344: what kind of warnings are this? If the ANTLR grammar has a mistake?

Felalolf: If that is the case, then this is ok. Otherwise, please be clearer regarding what has to be changed.

src/main/scala/viper/gobra/frontend/ReportFirstErrorStrategy.scala
```diff
@@ -0,0 +1,59 @@
...
+
+  override def reportFailedPredicate(recognizer: Parser, e: FailedPredicateException): Unit = {
+    val msg = e.getMessage
+    recognizer.notifyErrorListeners(e.getOffendingToken, msg, e)
+  }
+
+  override def recover(recognizer: Parser, e: RecognitionException): Unit = {
+    var context = recognizer.getContext
+    // For blocks that could be interpreted as struct literals (like `if a == b { }` or `switch tag := 0; tag { }`)
+    // Cast a wide net to catch every case, but still allow faster parsing if the error can't be an ambiguity.
```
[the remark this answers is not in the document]
Felalolf: could you add a comment about this. Something like "Default ANTLR code except the match statement which ..."

src/main/scala/viper/gobra/frontend/ReportFirstErrorStrategy.scala
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  override def recoverInline(recognizer: Parser): Token = {
+    val e = new InputMismatchException(recognizer)
+    var context = recognizer.getContext
+    context match {
+      case _ : BlockContext => throw new AmbiguityException
+      case _ : ExprSwitchStmtContext => throw new AmbiguityException
+      case _ =>
+    }
```
Felalolf, line 52: Same here. A comment regarding what is standard ANTLR code and what is introduced by you.

src/test/resources/regressions/features/trusted/trusted-functions.gobra
Felalolf: yes

Felalolf (commented):
Intermediate messages. I still have to look at InformativeErrorListener and ParseTreeTranslator. I will do this later today.

src/main/scala/viper/gobra/frontend/InformativeErrorListener.scala
```diff
@@ -0,0 +1,246 @@
...
+import org.bitbucket.inkytonik.kiama.util.{FileSource, Source}
+import viper.gobra.frontend.GobraParser.{CapContext, EosContext, ExpressionContext, ImplementationProofContext, RULE_blockWithBodyParameterInfo, RULE_eos, RULE_shortVarDecl, RULE_type_, RULE_varDecl, Slice_Context, TypeSpecContext, Type_Context, VarSpecContext, ruleNames}
+import viper.gobra.frontend.Source.FromFileSource
+import viper.gobra.reporting.ParserError
+import viper.silver.ast.SourcePosition
+
+import java.nio.file.Path
+import scala.collection.mutable.ListBuffer
+
+class InformativeErrorListener(val messages: ListBuffer[ParserError], val source: Source) extends BaseErrorListener {
```
Felalolf, line 20: This file needs more documentation. Please add:

1. For your analyze... methods, please add a description of which case they are handling. E.g. that a superfluous token was found which was not expected, or that a token was expected, etc. 
2. Whenever a non-default error type is returned, then please add a sentence describing what is going on. This should be as unambiguous as possible. E.g. "Found ':=', but expected end of line. Return suggestion to use '=' ". 
3. Also add to all other method 1-2 sentences describing what they are doing.

src/main/scala/viper/gobra/frontend/InformativeErrorListener.scala
```diff
@@ -0,0 +1,246 @@
...
+        }
+      }
+      case e : ExpressionContext if e.parent.isInstanceOf[CapContext] => SliceMissingIndex(3)
+      case _ if new_reserved.contains(context.offendingSymbol.getType) => ReservedWord()
+      case _ => DefaultNoViable(exception)
+    }
+  }
+
+
+  protected def underlineError(context : ErrorContext, restOfTheLine : Boolean = false): String = {
```
Felalolf, line 116: I do not understand what this is doing. If something mostly corresponds to a default ANTLR implementation and you only changed a specific part, then please clarify this. This holds for the entire PR.

src/main/scala/viper/gobra/frontend/InformativeErrorListener.scala
```diff
@@ -0,0 +1,246 @@
...
+      case (GobraParser.R_BRACKET, expr : ExpressionContext) if expr.parent.isInstanceOf[CapContext] => SliceMissingIndex(3)
+      case _ => DefaultExtraneous()
+    }
+  }
+
+  def analyzeInputMismatch(implicit context: ParserErrorContext, exception: InputMismatchException): ErrorType = {
+    (context.offendingSymbol.getType, context.recognizer.getContext) match {
+      case (Token.EOF, _) => DefaultMismatch()
+      case (s, i : ImplementationProofContext) if context.recognizer.getExpectedTokens == IntervalSet.of(GobraParser.IMPL)=> {
+        IgnoreError()
```
Felalolf, line 86: I assume this "IgnoreError" does not actually mean to ingore the error? Otherwise, why do we ignore an error?

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
Felalolf: push

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
```diff
@@ -0,0 +1,2240 @@
...
+    *
+    * @param clause
+    * @return
+    */
+
+  def visitIfClause(clause: IfStmtContext) : PIfClause = {
+    val pre = visitNodeOrNone[PSimpleStmt](clause.simpleStmt())
+    val expr = visitNode[PExpression](clause.expression())
+    // Emit a warning about syntax allowed by Gobra, but not by Go
+    if (clause.expression().stop.getType == GobraParser.R_CURLY) warn(clause.expression(), "struct literals at the end of if clauses must be surrounded by parens!")
```
Felalolf, line 1624:
```suggestion
    if (clause.expression().stop.getType == GobraParser.R_CURLY) warn(clause.expression(), "struct literals at the end of if clauses must be surrounded by parentheses!")
```

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
```diff
@@ -0,0 +1,2240 @@
...
+  //endregion
+
+  //region For statements
+  /**
+    * Visits a for statement with specifications
+    * @param specCtx
+    * @return a positioned PStatement
+    */
+  // TODO : Refactor
+  override def visitSpecForStmt(specCtx: SpecForStmtContext): PStatement = {
```
Felalolf, line 1722: Here it seems that you wanted to refactor something. Maybe use visitChildren instead of your current if-else cascade?

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
```diff
@@ -0,0 +1,2240 @@
...
+    * {@link #   visitChildren} on {@code ctx}.</p>
+    */
+  override def visitSelectStmt(ctx: SelectStmtContext): PSelectStmt = {
+    val clauses = ctx.commClause().asScala.toVector.map(visitCommClause)
+    // TODO: Do this in one pass
+    val send = clauses collect { case v: PSelectSend => v }
+    val rec = clauses collect { case v: PSelectRecv => v }
+    val arec = clauses collect { case v: PSelectAssRecv => v }
+    val srec = clauses collect { case v: PSelectShortRecv => v }
+    val dflt = clauses collect { case v: PSelectDflt => v }
```
Felalolf, line 1821: This is not very important, but keep the TODO.

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
```diff
@@ -0,0 +1,2240 @@
...
+  }
+
+  /**
+    * {@inheritDoc  }
+    *
+    * <p>The default implementation returns the result of calling
+    * {@link #   visitChildren} on {@code ctx}.</p>
+    */
+  // TODO : Refactor
+  override def visitCommClause(ctx: CommClauseContext): PSelectClause = {
```
Felalolf, line 1833: Again, it seems here you wanted to refactor something.

Felalolf (commented):
I am now done with the entire PR. It looks very good. Mostly comments about documentation and two refactorings in the translator.

src/main/antlr4/GoLexer.g4
```diff
@@ -0,0 +1,507 @@
...
+ */
+
+// Imported to Gobra from https://github.com/antlr/grammars-v4/blob/4c06ad8cc8130931c75ca0b17cbc1453f3830cd2/golang
+
+
+lexer grammar GoLexer;
+
+// Keywords
+
+BREAK                  : 'break' -> mode(NLSEMI);
```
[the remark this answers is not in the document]
nicoberling: We don't want the parser to have to deal with differentiating between tokens that allow us to omit semicolons at all, so this is always done in the lexer, e.g. "break\n" should emit the tokens `BREAK EOS`, where `EOS` denotes an implicit semicolon and "break;" should emit `BREAK SEMI`. This way, the parser just has to match `SEMI | EOS` whenever it expects a syntactic semicolon, with no regard for the token that came before.

src/main/antlr4/GobraLexer.g4
```diff
@@ -0,0 +1,79 @@
...
+SOME        : 'some'-> mode(NLSEMI);
+GET         : 'get'-> mode(NLSEMI);
+DOM         : 'domain'-> mode(NLSEMI);
+AXIOM       : 'axiom'-> mode(NLSEMI);
+NONE        : 'none' -> mode(NLSEMI);
+PRED        : 'pred';
+TYPE_OF      : 'typeOf'-> mode(NLSEMI);
+IS_COMPARABLE: 'isComparable'-> mode(NLSEMI);
+SHARE       : 'share';
+ADDR_MOD    : '@'-> mode(NLSEMI);
```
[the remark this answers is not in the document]
nicoberling: Because `@` is essentially a part of the identifier, removing nlsemi here would allow somewhat inconsistent syntax:
```
x@
:= pair{1,2}
```
would be a valid declaration of x, while
```
x // implicit ';' induced here 
:= pair{1,2} // statements can't start with ':='
```
would not. In the end, this depends on what should and should not be allowed as syntax, what do you think?

src/main/scala/viper/gobra/ast/frontend/Ast.scala
nicoberling: PTypeName corresponds to [TypeName](https://go.dev/ref/spec#TypeName) in the go specification, while PUnqualifiedTypeName corresponds to types that are just identifiers without a dot, including predeclared types.

src/main/scala/viper/gobra/frontend/Parser.scala
nicoberling: In some cases when Gobra allows some structure that is not allowed in Go, like struct literals that would have to be parenthesised, I create a warning in the translator. We decided not to emit errors, since it's not always detected, so it would be a bit inconsistent. I think Viper verification results don't support warnings right? Either everything is okay or it is a failure with errors?

Felalolf: Just clarify the "Properly handle parser warning" so that we know what is meant by this in the future.

src/main/antlr4/GobraLexer.g4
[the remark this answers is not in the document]
Felalolf: you are right. I mixed up the order.

src/main/antlr4/GoLexer.g4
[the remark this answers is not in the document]
Felalolf: ok

src/main/antlr4/GoLexer.g4
```diff
@@ -0,0 +1,507 @@
...
+                       | '.' HEX_DIGIT ('_'? HEX_DIGIT)*;
+
+fragment HEX_EXPONENT  : [pP] [+-] DECIMALS;
+
+
+IMAGINARY_LIT          : (DECIMAL_LIT | BINARY_LIT |  OCTAL_LIT | HEX_LIT | FLOAT_LIT) 'i' -> mode(NLSEMI);
+
+// Rune literals
+
+fragment RUNE               : '\'' (UNICODE_VALUE | BYTE_VALUE) '\'';//: '\'' (~[\n\\] | ESCAPED_VALUE) '\'';
```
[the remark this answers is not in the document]
nicoberling: As far as I can tell, they do support all rune literals and floats from the go spec. However, my translation does not support all of those, I will make sure to add clear comments.

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -159,47 +285,47 @@ object Parser {
...
     val positions = new Positions
     val pom = new PositionManager(positions)
-    val parsers = new SyntaxAnalyzer(pom, specOnly)
-    translateParseResult(pom)(parsers.parseAll(parsers.program, preprocessedSource))
+    val parser = new antlrSyntaxAnalyzer[SourceFileContext, PProgram](source, ListBuffer.empty[ParserError],  pom, specOnly)
+    parser.parse(parser.sourceFile())
   }
 
-  def parseMember(source: Source, specOnly: Boolean = false): Either[Messages, Vector[PMember]] = {
+  def parseFunction(source: Source, specOnly: Boolean = false): Either[Vector[ParserError], PMember] = {
```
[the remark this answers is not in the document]
nicoberling: No, they are not the same type, should I declare some type alias like `ParserErrors` to make it more readable?

src/main/scala/viper/gobra/frontend/Parser.scala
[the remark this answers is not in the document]
nicoberling: This should now only happen when there was an acual issue with ANTLR, so I just rethrow the exception, since we don't know what happened

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
nicoberling: Most of these would be nice to have, but are readable as is. It may be best to leave them for now.

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
nicoberling: I've somewhat generalized it now.

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
Felalolf: ok.

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
Felalolf: I assume, as with the previous point, we leave this for now, which is fine. I just wanted to bring it to your attention, in case you forgot it.

src/main/scala/viper/gobra/frontend/InformativeErrorListener.scala
```diff
@@ -0,0 +1,324 @@
...
+      case _ => DefaultErrorType()
+    }
+  }
+
+
+  /**
+    * Failed predicate errors are emitted when the no viable alternatives remain, as all conflict with the input
+    * or contain a semantic predicate that evaluated to false. Because the only predicate present in Gobra's grammar
+    * is responsible for inducing semicolons, we know that we are at a point where a statement could have ended, but
+    * another token was discovered.
```
Felalolf, line 76:
```suggestion
    * Failed predicate errors are emitted when no viable alternatives remain, as all alternatives conflict with the input
    * or contain a semantic predicate that evaluated to false. Because the only predicate present in Gobra's grammar
    * is responsible for inducing semicolons, we know that if a predicate failed, we are at a point where a statement could have ended, but
    * another token was discovered.
```

src/main/scala/viper/gobra/frontend/InformativeErrorListener.scala
```diff
@@ -0,0 +1,324 @@
...
+      case (GobraParser.DECLARE_ASSIGN, _ : TypeSpecContext) => GotAssignErrorType()
+      // We expected more tokens inside a slice expression but got a closing bracket: One of the
+      // limits must be missing.
+      case (GobraParser.R_BRACKET, expr : ExpressionContext) if expr.parent.isInstanceOf[CapContext] => SliceMissingIndex(3)
+      case _ => DefaultExtraneous()
+    }
+  }
+
+  /**
+    * The input did not match the expecte tokens. This one of the most common exceptions.
```
Felalolf, line 119:
```suggestion
    * The input did not match the expecte tokens. This is one of the most common exceptions.
```

src/main/java/viper/gobra/frontend/GobraParserBase.java
```diff
@@ -0,0 +1,28 @@
...
+{
+    protected GobraParserBase(TokenStream input) {
+        super(input);
+    }
+
+    /**
+     * Returns true if the current Token is a closing bracket (")" or "}")
```
Felalolf, line 19:
```suggestion
     * Returns true if the current Token is a closing bracket (")" or "}" or "!>")
```

Felalolf: Currently, "!>" is a closing bracket. If the method returns true if the current token is a closing bracket, then this includes "!>". Alternatively, the method is doing something else. I understand that where this function is used, to use it to check for the semicolon before a closing bracket, but this is orthogonal to the description of this method.

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
Felalolf: ok.

Felalolf (commented):
looks ok to me. There a few grammar and spelling errors in some comments. I annotated, the ones that I care about. After these suggestions are commited and the TODO about ANTLR errors is clarified (specified in one of my replies), this can be merged in my opinion.

src/main/scala/viper/gobra/frontend/InformativeErrorListener.scala
Felalolf: did this vanish? I cannot find it anymore.

ArquintL (approved):

Issue #362, opened 2021-11-17
Cannot compare Slices in Assertions


Felalolf:
Currenty, comparing slices is not allowed inside of assertion.

PR #389, opened 2022-01-17
Fix issue 362

jcp19:
Contains the fix to issue #362 that has been in use for some time in verifiedSCION.


src/main/scala/viper/gobra/frontend/info/implementation/resolution/Enclosing.scala
```diff
@@ -42,6 +42,9 @@ trait Enclosing { this: TypeInfoImpl =>
   lazy val isEnclosingExplicitGhost: PNode => Boolean =
     down(false){ case _: PGhostifier[_] => true }
 
+  lazy val isEnclosingtGhost: PNode => Boolean =
```
Felalolf, line 45: spelling

src/main/scala/viper/gobra/frontend/info/implementation/property/Comparability.scala
```diff
@@ -25,6 +25,13 @@ trait Comparability extends BaseProperty { this: TypeInfoImpl =>
     case _ => false
   }
 
+  lazy val ghostComparableTypes: Property[(Type, Type)] = createFlatProperty[(Type, Type)] {
+    case (left, right) => s"$left is not comparable in ghost with $right"
+  } {
+    case (Single(left), Single(right)) => assignableTo(left, right) || assignableTo(right, left)
```
Felalolf, line 31: I was initially confused that not more checks are neccessary, but I guess the `(NilType, r) if isPointerType(r)` check is not necessary since this usually makes comparisons of uncomparable types with nil possible.

Felalolf (approved):
One spelling mistake.

ArquintL (approved):

PR #390, opened 2022-01-18
Std lib changes from the verifiedSCION branch

jcp19:
This PR introduces small changes to the std lib (mostly termination measures) that are helpful to verify the SCION codebase. After this PR, the verified scion branch and master should be on par, with the exception of the support for the outline keyword (merged in the scion branch, not merged on master) which does not work well anyway.

### DIFF
--- a/src/main/resources/stubs/net/ip.gobra
+++ b/src/main/resources/stubs/net/ip.gobra
@@ -122,12 +122,12 @@ func (ip IP) IsGlobalUnicast() bool /*{
 
 // To4 converts the IPv4 address ip to a 4-byte representation.
 preserves forall i int :: 0 <= i && i < len(ip) ==> acc(&ip[i], 1/10000)
+ensures res != nil ==> len(res) == IPv4len
 ensures len(ip) == IPv4len ==> forall i int :: 0 <= i && i < len(ip) ==> &ip[i] == &res[i]
-ensures len(ip) == IPv6len && isZeros(ip[0:10]) && ip[10] == 255 && ip[11] == 255 ==> len(res) == IPv4len && forall i int :: 0 <= i && i < IPv4len ==> &ip[12+i] == &res[i]
-ensures (len(ip) == IPv6len && res != nil) ==> (len(res) == IPv4len && forall i int :: 0 <= i && i < IPv4len ==> &ip[12+i] == &res[i])
+ensures (len(ip) == IPv6len && isZeros(ip[0:10]) && ip[10] == 255 && ip[11] == 255) ==> res != nil
+ensures (len(ip) == IPv6len && res != nil) ==> (forall i int :: 0 <= i && i < IPv4len ==> &ip[12+i] == &res[i])
 ensures len(ip) != IPv4len && len(ip) != IPv6len ==> res == nil
 ensures len(ip) == IPv4len ==> len(res) == IPv4len
-ensures res != nil ==> len(res) == IPv4len
 decreases
 func (ip IP) To4() (res IP) {
 	if len(ip) == IPv4len {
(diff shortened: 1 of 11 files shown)


src/main/resources/stubs/net/ip.gobra
```diff
@@ -110,9 +124,11 @@ func (ip IP) IsGlobalUnicast() bool /*{
 preserves forall i int :: 0 <= i && i < len(ip) ==> acc(&ip[i], 1/10000)
 ensures len(ip) == IPv4len ==> forall i int :: 0 <= i && i < len(ip) ==> &ip[i] == &res[i]
 ensures len(ip) == IPv6len && isZeros(ip[0:10]) && ip[10] == 255 && ip[11] == 255 ==> len(res) == IPv4len && forall i int :: 0 <= i && i < IPv4len ==> &ip[12+i] == &res[i]
+ensures (len(ip) == IPv6len && res != nil) ==> (len(res) == IPv4len && forall i int :: 0 <= i && i < IPv4len ==> &ip[12+i] == &res[i])
```
ArquintL, line 127: Would this be equivalent to state on the line above that res won't be nil. However it seems weird that this is not already implied by `len(res) == IPv4len`. I bring this up because the RHS seems very similar

src/main/resources/stubs/net/net.gobra
```diff
@@ -294,12 +294,14 @@ type DNSError struct {
 pred (e *DNSError) Mem() { acc(e) }
 
 preserves e != nil ==> acc(e, 1/1000)
+trusted
```
ArquintL, line 297: Does this allow you to verify this file with Gobra as the body is ignored and thus the exception does not occur? If so, please add a comment next to the trusted annotation as I don't think I would figure out what its purpose is when I'm not seeing that you've also added the comment below

ArquintL (approved):
To small comments, otherwise looks good to me

src/main/resources/stubs/net/net.gobra
jcp19: I actually found the root cause of this issue and fixed this ^^ so, it is not trusted anymore, and I even manage to make the code simpler to fix this.

src/main/resources/stubs/net/ip.gobra
jcp19: I'm not sure if I understood this comment. Anyhow, I think I can further simplify this line from `ensures (len(ip) == IPv6len && res != nil) ==> (len(res) == IPv4len && forall i int :: 0 <= i && i < IPv4len ==> &ip[12+i] == &res[i])` to `ensures (len(ip) == IPv6len && res != nil) ==> (forall i int :: 0 <= i && i < IPv4len ==> &ip[12+i] == &res[i])` but I don't think I can simplify it further, because it might be the case that `res` is nil when the input has size `IPv6len`.

ArquintL: I was basically looking at the implementation and it seems to me that the only case in which you get a non-nil result and `len(ip) == IPV6len` is the one covered by the specification in the line above.
The newly added postcondition only provides additional insights to clients that do not know whether `isZeros(ip[0:10]) && ip[10] == 255 && ip[11] == 255` holds but know that the result is non-nil

ArquintL: Following this logic, you could save redundancy by changing line 126 to only imply that the result is non-nil (which would then enable clients to use the new implication on line 127

jcp19: I see, for some reason I didn't notice this. Thanks!

ArquintL (approved):

PR #391, opened 2022-01-19
Parser Improvements

ArquintL:
Still a PR for the old parser but I need the following 3 improvements:
- keep bodies of pure functions & methods in imported packages (saves writing an artificial postcondition with the same assertion)
- 'ghost' can be on its own line in interface declarations (thus, one can write similar looking spec as one is used to write for regular functions/methods)
- ghost types can be parsed as one would expect. Without this change, the newly created test case could not be parsed

@jcp19 does the first improvement cause performance problems in VerifiedSCION? Alternatively, we could introduce an annotation `transparent` or `opaque` (depending on what the default should be) or a CLI option such that the behavior could be controlled

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Parser.scala
+++ b/src/main/scala/viper/gobra/frontend/Parser.scala
@@ -1032,7 +1032,7 @@ object Parser {
         ) <~ not("(" | "{")
 
     lazy val typ : Parser[PType] =
-      "(" ~> typ <~ ")" | typeLit | qualifiedType | namedType | ghostTypeLit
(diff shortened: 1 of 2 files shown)


jcp19 (approved):
Looks good to me. I think this should be fine with verifiedSCION. I think that introducing the `opaque`/`transparent` annotations might be a bit too much. An alternative would be to pass a flag to Gobra that tells if pure functions from other packages should keep their bodies.

Felalolf (approved):
If all tests pass, then this should be fine.

PR #393, opened 2022-01-20
Implicit Conversion for Predicate Arguments

ArquintL:
Using a struct for an interface parameter in function performs an implicit conversion `toInterface` such that the encoded Viper program is well typed. This implicit conversion is however not done for predicates, which is fixed by this PR.
The added unit test illustrates the problem best

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Desugar.scala
+++ b/src/main/scala/viper/gobra/frontend/Desugar.scala
@@ -1427,14 +1427,20 @@ object Desugar {
     }
 
     def arguments(symb: st.WithArguments, args: Vector[in.Expr]): Vector[in.Expr] = {
-      val c = args zip symb.args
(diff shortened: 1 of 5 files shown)


src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -2911,45 +2922,76 @@ object Desugar {
...
+          case base: ap.BuiltInReceivedPredicate =>
+            Vector(info.typ(base.recv))
+          case base: ap.BuiltInPredicateExpr =>
+            Vector(info.symbType(base.typ))
+        }
+        Violation.violation(abstractType.typing.isDefinedAt(argsForTyping), s"cannot type built-in member ${pk.symb.tag} as it is not defined for arguments $argsForTyping")
+        abstractType.typing(argsForTyping)
+      }
+
+      /** args must not include the receiver in the case of received predicates */
```
Felalolf, line 2939: Please add the reason, as well.

Felalolf (approved):
One remark about a comment, but ok otherwise.

PR #394, opened 2022-01-20
Implementation Proof Generation for Imported Subtyping Relations

ArquintL:
As discussed and as described in `proofGenForImport.gobra` and `proofGenForImportPkg/pkg.gobra`

Depends on #392


Felalolf (approved):
If the tests go through, then this should be fine.

PR #395, opened 2022-01-21
Stats collector

jogasser:
Added a reporter, that collects stats from received messages. These stats get written to a .gobra directory, which is per default located in the current directory.

Other changes:

- Changed input file management to support whole directories (added -r option for recursive traversal)
- Removed package lookup per input file
- Added option (-p) to filter for specific package names
- Added Builtin Tag for programs, that get injected by gobra (see src/main/resources/builtin or src/main/resources/stubs)
- Added Path to package ast nodes
- Fixes https://github.com/viperproject/gobra/issues/417


src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -212,44 +217,103 @@ class GobraFrontend {
...
+          logger.info("Writing report to " + statsFile.getPath)
+          statsCollector.writeJsonReportToFile(statsFile)
+        }
+      })
+
+      config.inputPackageMap.foreach({ case (pkg, inputs) =>
+        val pkgString = pkg.path + " - " + pkg.name
+
+        logger.info("Verifying Package " + pkgString)
+        val resultFuture = verifier.verify(config.copy(inputs=inputs, reporter = statsCollector), Some(pkgString))(executor)
```
jcp19, line 264: minor: inconsistent spacing in `inputs=inputs, reporter = statsCollector`

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -16,16 +16,27 @@ import org.rogach.scallop.{ScallopConf, ScallopOption, listArgConverter, singleA
...
   val DefaultLevel: Level = Level.INFO
 }
+
+object Config {
+  val DefaultGobraDirectory: String = ".gobra"
+}
+
+case class PackageEntry(path: String, name: String) {
+
+}
```
jcp19, line 34: We can drop the brackets in the definition of `PackageEntry`

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -116,9 +136,31 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
...
+  val recursive: ScallopOption[Boolean] = toggle(
+    name = "recursive",
+    descrYes = "Perform verification on all subfolders of the specified input directory",
+    default = Some(false),
+    short = 'r'
+  )
+
+  val packages: ScallopOption[List[String]] = opt[List[String]](
+    name = "packages",
+    descr = "Which packages should be verified, all packages are verified per default",
```
jcp19, line 152: What do you mean by "all packages are verified per default"?

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -212,44 +217,103 @@ class GobraFrontend {
...
+        exitCode = 1
+    } finally {
+      executor.terminate()
+
+      logger.info("\n")
+      logger.info("Summary:")
+      logger.info("\n")
 
+      // Write console summary
+      if(warningCount > 1) {
```
jcp19, line 307: you mean "> 0" instead of ">1"?

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -212,44 +217,103 @@ class GobraFrontend {
...
+        logger.warn(s"${verifier.name} has found $warningCount warnings(s)")
+      }
+      if(errorCount > 1) {
```
jcp19, line 310: same as before

jcp19 (commented):
I'm still going through the PR, but I've already have some comments

src/main/scala/viper/gobra/frontend/Config.scala
jogasser: The option should limit, which packages are verified. If it is omitted, all packages found in the specified inputs are verified

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,345 @@
...
+
+  def writeJsonReportToFile(file: File): Unit = {
+    if((file.exists() && file.canWrite) || file.getParentFile.canWrite) {
+      FileUtils.writeStringToFile(file, getJsonReport(false), UTF_8)
+    }
+  }
+
+  def getJsonReport(shorten: Boolean): String = {
+    val json = "[\n" +
+      memberMap.map({ case (key, gobraMember) => s"""  {
```
jcp19, line 114: I think that the code here is hard to read.

It would be better if we defined the "jsonification" of `memberMap`, `viperMembers`, and `dependencies` in separate variables and then introduced the result in the final string instead of doing this all in the declaration of variable `json`.

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,345 @@
...
+        "hasBody": ${viperMember.hasBody},
+        "cached": ${viperMember.cached}
+      }""").mkString(", \n") + "\n    ],\n" + s"""    "dependencies": [""" + "\n" +
+        gobraMember.viperMembers.values
+          .flatMap(viperMember => getDependencies(viperMember.member)
+            .flatMap(dep => findGobraMemberByViperMemberName(viperMember.taskName, dep, gobraMember.pkgDir))
+          )
+          // Filter out dependencies on one self, since this information isn't very useful
+          .filter(_ != gobraMember)
+          .map(gobraMemberEntry => "        \"" + gobraMemberKey(gobraMemberEntry.pkgDir, gobraMemberEntry.pkg, gobraMemberEntry.memberName, gobraMemberEntry.args) + "\"")
```
jcp19, line 140: we can use a number of tabs to keep track of the indentation instead of having string literals with fixed spaces

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,345 @@
...
+          .toSet
+          .mkString(", \n") + "\n    ]\n  }"
+      }).mkString(", \n") + "\n]\n"
+
+    if(shorten) {
+      // Replaces all spaces, except the ones inside of quotes
+      json.replaceAll("\\s+(?=([^\"]*\"[^\"]*\")*[^\"]*$)", "")
+    } else {
+      json
+    }
```
jcp19, line 150: when is it desired/useful to have the shortened version instead of the "normal" one?

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jcp19: Out of curiosity, would the classes `JSONArray` and `JSONObject` (http://lampwww.epfl.ch/~hmiller/scaladoc/library/scala/util/parsing/json/package.html) be useful here to avoid having to create this string "by hand"?

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,345 @@
...
+  private var memberMap: Map[String, GobraMemberEntry] = Map()
+  // Maps a viper member name to a gobra member entry
+  private var viperMemberNameGobraMemberMap: Map[String, GobraMemberEntry] = Map()
+
+  case class GobraMemberEntry(pkgDir: String,
+                              pkg: String,
+                              memberName: String,
+                              nodeType: String,
+                              args: String,
+                              var viperMembers: Map[String, ViperMemberEntry],
```
jcp19, line 36: Is there an easy way to avoid having mutable fields here? Couldn't we move this map to a private field of the `StatsCollector`?

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,345 @@
...
+import viper.silver.ast.{FuncApp, Function, Member, Method, MethodCall, Node, Predicate, PredicateAccess}
+import viper.silver.reporter.Time
+
+import java.nio.charset.StandardCharsets.UTF_8
+import scala.annotation.tailrec
+import java.io.File
+
+case class StatsCollector(reporter: GobraReporter) extends GobraReporter {
+  // Stores type info for a task
+  var typeInfos: Map[String, TypeInfo] = Map()
```
jcp19, line 25: Should be private

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,345 @@
...
+  private var viperMemberNameGobraMemberMap: Map[String, GobraMemberEntry] = Map()
+
+  case class GobraMemberEntry(pkgDir: String,
+                              pkg: String,
+                              memberName: String,
+                              nodeType: String,
+                              args: String,
+                              var viperMembers: Map[String, ViperMemberEntry],
+                              isTrusted: Boolean,
+                              var isAbstract: Boolean)
```
jcp19, line 38: Does this field have to be mutable?

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -261,24 +308,29 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
...
   def validateInput(inputOption: ScallopOption[List[String]],
-                    moduleOption: ScallopOption[String],
-                    includeOption: ScallopOption[List[File]]): Unit = validateOpt(inputOption, includeOption) { (inputOpt, includeOpt) =>
+                    recOption: ScallopOption[Boolean],
+                    includePkgOption: ScallopOption[List[String]],
+                    excludePkgOption: ScallopOption[List[String]]): Unit =
+    validateOpt(inputOption, recOption, includePkgOption, excludePkgOption) { (inputOpt, recOpt, includePkgOpt, excludePkgOpt) =>
 
-    def checkConversion(input: List[String], moduleName: String, includeDirs: Vector[Path]): Either[String, Vector[Path]] = {
+    def checkConversion(input: List[String]): Either[String, Vector[Path]] = {
```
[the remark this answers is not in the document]
jogasser: I am not sure if this is true, `PackageResolver` still depends on the `moduleName` field when looking up packages. Here, this argument was only used when interpreting the given input path as a package, which I removed with this PR.

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jogasser: It saves some space, for example, if we compare the output for the `stats_collector` test directory, the normal version uses 17.1KB and the shortened one 12.5KB. This might be relevant if these will be used as artifacts for a GitHub workflow.

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jogasser: I changed the solution to work with prefixes for indentation. I don't know if this is what you had in mind or if you explicitly prefer tabs over spaces.

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jogasser: Nevermind, replacing spaces for large strings this way seems to be actually very slow and the files don't seem to get unreasonably big

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -212,44 +273,49 @@ class GobraFrontend {
...
-          errors foreach (e => logger.error(s"\t${e.formattedMessage}"))
-          sys.exit(1)
+      // Set up gobra directory
+      val gobraDirectory = config.gobraDirectory.toAbsolutePath.toFile
+      if(!gobraDirectory.isDirectory) {
+        Violation.violation(gobraDirectory.mkdir(), s"Could not create directory $gobraDirectory")
+      }
+
+      // Make sure we have the correct permissions to the gobra directory
+      Violation.violation(gobraDirectory.canRead && gobraDirectory.canWrite, "Couldn't write to gobra directory " + config.gobraDirectory.toString)
```
jcp19, line 290: This error message can be misleading if there is write permission to the directory but not read permission. I would separate this in two checks.

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -212,44 +273,49 @@ class GobraFrontend {
...
-          sys.exit(0)
-        case VerifierResult.Failure(errors) =>
-          logger.error(s"${verifier.name} has found ${errors.length} error(s):")
-          errors foreach (e => logger.error(s"\t${e.formattedMessage}"))
-          sys.exit(1)
+      // Set up gobra directory
+      val gobraDirectory = config.gobraDirectory.toAbsolutePath.toFile
+      if(!gobraDirectory.isDirectory) {
+        Violation.violation(gobraDirectory.mkdir(), s"Could not create directory $gobraDirectory")
+      }
```
jcp19, line 287: what happens here if there is already a file with the intended name?

src/main/scala/viper/gobra/backend/BackendVerifier.scala
```diff
@@ -49,10 +50,9 @@ object BackendVerifier {
 
     val verificationResults =  {
       val verifier = config.backend.create(exePaths, config)
-      val programID = s"_programID_${config.inputs.map(_.name).mkString("_")}"
 
       if (!config.shouldChop) {
-        verifier.verify(programID, BacktranslatingReporter(config.reporter, task.backtrack, config), task.program)(executor)
+        verifier.verify(task.name, BacktranslatingReporter(config.reporter, task.backtrack, config, task.name), task.program)(executor)
```
jcp19, line 55: In my previous comment, I mentioned that you could drop the `programID` because it was derived from the `task.name`. What I meant was actually that you could probably drop one parameter altogether - If for some reason the first and last parameters are always the same, could we drop one of them?

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -116,9 +133,38 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
...
+  val recursive: ScallopOption[Boolean] = toggle(
+    name = "recursive",
+    descrYes = "Verify nested packages recursively",
+    default = Some(false),
+    short = 'r'
+  )
+
+  val includePackages: ScallopOption[List[String]] = opt[List[String]](
+    name = "includePackages",
+    descr = "Which packages should be verified, all packages are verified by default",
```
jcp19, line 149: Would replace with "Packages to verify. All packages found in the specified directories are verified by default."

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -116,9 +133,38 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
...
+    default = Some(List.empty),
+    short = 'p'
+  )
+
+  val excludePackages: ScallopOption[List[String]] = opt[List[String]](
+    name = "excludePackages",
+    descr = "Which packages should not be verified, all packages are verified by default",
```
jcp19, line 156: Would replace with "Packages to ignore. These packages will not be verified, even if they are found in the specified directories."

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -116,9 +133,38 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
...
+    default = Some(List.empty),
+    noshort = true
+  )
+
+  val gobraDirectory: ScallopOption[Path] = opt[Path](
+    name = "gobraDirectory",
+    descr = "Directory in which gobra saves it's data like its cache or statistics",
```
jcp19, line 163: Would just say "Output directory for Gobra".

src/test/resources/stats_collector/main.gobra
```diff
@@ -0,0 +1,30 @@
...
+  return s.Area()
+}
+
+
+requires i > 0 && j > 0
+ensures ret > 0
+trusted
+func test(i int, j int) (ret int) {
+
+}
```
jcp19, line 30: You could probably just remove the "trusted" annotation and the method body

src/test/resources/stats_collector/pkg1/pkg1.gobra
```diff
@@ -0,0 +1,26 @@
...
+func abstractFunction(d int) int
+
+trusted
+ensures ret > 0
+func trustedFunction(d int) (ret int) {
+
+}
+
+
+
```
jcp19, line 26: same here

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,357 @@
...
+            isAbstract)
+        case _ =>
+      }
+    }
+
+    msg match {
+      // Capture typeInfo once it's available
+      case TypeInfoMessage(typeInfo, taskName) => this.synchronized({ typeInfos = typeInfos + (taskName -> typeInfo)})
+      // Free up unneeded space, once a task is finished
+      case VerificationTaskFinishedMessage(taskName) => this.synchronized({ typeInfos = typeInfos - taskName})
```
jcp19, line 138: In what contexts could we have a concurrent access to the map? If we really need it, why not use a ConcurrentHashMap for this map? That seems like a less brittle alternative to having synchronized blocks which we can easily forget about when the code is updated

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,357 @@
...
+            // Merge Viper members that occur multiple times during a task(used for chopper)
+            existing.viperMembers + (viperKey -> ViperMemberEntry(
+              existingViperEntry.memberName,
+              existingViperEntry.taskName,
+              existingViperEntry.time + viperMember.time,
+              existingViperEntry.nodeType,
+              existingViperEntry.dependencies ++ viperMember.dependencies,
+              existingViperEntry.success && viperMember.success,
+              existingViperEntry.cached || viperMember.cached,
+              existingViperEntry.fromImport,
```
jcp19, line 214: Here, we are using a synchronized block again. You should try to document why this is required if this is really required. Also, not sure about what kinds of race conditions we can have here, but this is only asserting mutual exclusion when adding a result, although there might still be concurrent accesses e.g. to `memberMap` from other methods.

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jcp19: The reason why I meant tabs is that you can easily tell the indentation of something by looking at the string literal, whereas when you have string literals with an arbitrary number of spaces you literally need to count them

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jcp19: we can just drop it then

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,357 @@
...
+        .filter(_.key != this.key).toSet
+    }
+
+    def key: String = gobraMemberKey(pkgDir, pkg, memberName, args)
+  }
+
+  case class ViperMemberEntry(memberName: String,
+                              taskName: String,
+                              time: Time,
+                              nodeType: String,
```
jcp19, line 74: If this is supposed to have only one value out of a small and well-defined set of values, I would not make this a string. Instead, I would use an Enum or a new trait

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,357 @@
...
+      .flatMap(gobraMember => {
+        val name = gobraMember.pkg + "." + gobraMember.memberName + gobraMember.args
+        gobraMember.dependencies().flatMap({
+          // Trusted implies abstracted, so we match trusted first
+          case GobraMemberEntry(_, pkg, memberName, _, args, _, true, _) =>
+            Some("Warning: Member " + name + " depends on trusted member " + pkg + "." + memberName + args + "\n")
+          case GobraMemberEntry(_, pkg, memberName, _, args, _, _, true) =>
+            Some("Warning: Member " + name + " depends on abstract member " + pkg + "." + memberName + args + "\n")
+          case GobraMemberEntry(pkgDir, pkg, _, _, _, _, _, _) if !config.inputPackageMap.contains(PackageEntry(pkgDir, pkg)) =>
+            Some("Warning: Depending on imported package, that is not verified: " + pkgDir + " - " + pkg)
```
jcp19, line 183: If the idea is to say that the imported package is not verified, drop the `,`

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jogasser: You are correct, I switched to concurrent maps and removed all synchronized blocks

src/test/resources/stats_collector/main.gobra
jogasser: The goal here is to have a trusted member, to check if the stats collector writes the trusted flag correctly. The body doesn't matter to me

src/main/scala/viper/gobra/backend/BackendVerifier.scala
jogasser: I am still not sure what you mean, but I moved `BacktranslatingReporter(config.reporter, task.backtrack, config, task.name)` out of there since it's the same for the normal and the chopper case

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -42,38 +42,91 @@ object GoVerifier {
...
+    })
+
+    if(warningCount > 0) {
+      logger.warn(s"$name has found $warningCount warning(s)")
+    }
+    if(allErrors.nonEmpty) {
+      logger.error(s"$name has found ${allErrors.size} error(s)")
+    } else {
+      logger.info(s"$name found no errors")
+    }
```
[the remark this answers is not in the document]
jogasser: Should this then be logged to `error` or `info`?

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -42,38 +42,91 @@ object GoVerifier {
...
+    Runtime.getRuntime.addShutdownHook(new Thread() {
+      override def run(): Unit = {
+        val statsFile = config.gobraDirectory.resolve("stats.json").toFile
+        logger.info("Writing report to " + statsFile.getPath)
+        statsCollector.writeJsonReportToFile(statsFile)
+      }
+    })
+
+    config.inputPackageMap.foreach({ case (pkg, inputs) =>
+      val pkgString = pkg.path + " - " + pkg.name
```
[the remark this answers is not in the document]
jogasser: I use this as the `taskName`, which should be unique per package. I renamed the variable to make this clearer.
Simply using the pathname or combining the path with the package name could lead to ambiguous `taskName` if we allow packages to have a different name than the directory they are in.

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jogasser: I moved the indentation to  a variable, this way it is much clearer how much it is indented and it is configurable at a single point, how it should be indented

src/main/scala/viper/gobra/Gobra.scala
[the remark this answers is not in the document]
jcp19: I would log this to `info`. To me, the `error` level is supposed to log errors in the Gobra execution itself (if some unexpected condition is met, for example), whereas the errors that we are talking about here are the errors of the program under verification - the fact that Gobra reports them shouldn't be though as an error of Gobra, it's just its expected output under normal operation.

src/main/scala/viper/gobra/backend/BackendVerifier.scala
jcp19: Oops, my bad, for some reason I did not see that the task.name is a param to the BacktranslatingReporter. Nevermind this comment

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -166,7 +166,7 @@ case class StatsCollector(reporter: GobraReporter) extends GobraReporter {
 
   def getNumberOfCachedViperMembers: Int = memberMap.values.flatMap(_.viperMembers.values).count(_.cached)
 
-  def getNumberOfCacheableViperMembers: Int = memberMap.values.flatMap(_.viperMembers.values).count(member => member.hasBody && member.nodeType == "Method")
+  def getNumberOfCacheableViperMembers: Int = memberMap.values.flatMap(_.viperMembers.values).count(member => member.hasBody && member.nodeType.eq(ViperNodeType.Method))
```
jcp19, line 169: why use referential equality here?

jogasser: Oh I didn't realize there where different equality checks, good catch :)

src/main/scala/viper/gobra/reporting/Message.scala
```diff
@@ -29,6 +30,11 @@ sealed trait GobraVerificationResultMessage extends GobraMessage {
   val verifier: String
 }
 
+case class VerificationTaskFinishedMessage(taskName: String) extends GobraMessage {
```
ArquintL, line 33: What is the difference between `VerificationTaskFinishedMessage` and `GobraOverallSuccessMessage` resp. `GobraOverallFailureMessage`?

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,376 @@
...
+  case class GobraMemberEntry(pkgDir: String,
+                              pkg: String,
+                              memberName: String,
+                              args: String,
+                              nodeType: GobraNodeType,
+                              viperMembers: Map[String, ViperMemberEntry],
+                              isTrusted: Boolean,
+                              isAbstract: Boolean) {
+
+    def asJson(p: String = ""): String = {
```
ArquintL, line 56: what is `p`? is this the current indentation level?

ArquintL: Wouldn't it make sense to use the same JSON library as used in ViperServer instead of writing this ourselves?

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,376 @@
...
+  private[reporting] val memberMap: Map[String, GobraMemberEntry] = TrieMap()
+  // Maps a viper member name to a gobra member entry
+  private val viperMemberNameGobraMemberMap: Map[String, GobraMemberEntry] = TrieMap()
+
+  // indentation prefix for generated json
+  private val i = "  "
+  case class GobraMemberEntry(pkgDir: String,
+                              pkg: String,
+                              memberName: String,
+                              args: String,
```
ArquintL, line 50: what is `args`?

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,376 @@
...
+         |$p$i"nodeType": "${this.nodeType}",
+         |$p$i"success": ${this.success},
+         |$p$i"cached": ${this.cached},
+         |$p$i"fromImport": ${this.fromImport},
+         |$p$i"hasBody": ${this.hasBody}
+         |$p}""".stripMargin
+    }
+  }
+
+  case class GobraMemberInfo(pkgDir: String,
```
ArquintL, line 112: why is `GobraMemberInfo` not used as a parameter of `GobraMemberEntry`? `GobraMemberEntry` seems to take a superset of parameters

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,376 @@
...
+              existingViperEntry.success && viperMember.success,
+              existingViperEntry.cached || viperMember.cached,
+              existingViperEntry.fromImport,
+              existingViperEntry.hasBody || viperMember.hasBody
+            ))
+          case None => existing.viperMembers.put(viperKey, viperMember)
+        }
+
+        // If we encounter an abstract version of a member, we know for sure, its abstract
+        val newAbstract = existing.isAbstract || isAbstract
```
ArquintL, line 238: why is disjunction used here? Wouldn't a member occur once non-abstract and possibly multiple times abstract in all chopped parts?

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,376 @@
...
+      case p: PNode => getMemberInformation(nodeTypeInfo.codeRoot(p), typeInfo)
+    }
+  }
+
+  /**
+   * Checks whether a Go Tree contains a node
+   */
+  def treeContains(tree: Info.GoTree, p: PNode): Boolean = {
+    try {
+      tree.whenContains(p, true)
```
ArquintL, line 351: `whenContains` throws an error if node is not contained? That's weird

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,376 @@
...
+    } catch {
+      case NodeNotInTreeException(_) => false
+    }
+  }
+
+  /**
+   * Returns a set of viper member names (Methods, Predicates & Functions) which are accessed by a given viper member
+   */
+  def getViperDependencies(m: Member): Set[String] = m match {
```
ArquintL, line 360: Couldn't we make use of the dependency analysis that the Chopper has already implemented instead of replicating it here?

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,376 @@
...
+                memberName: String,
+                args: String,
+                nodeType: GobraNodeType,
+                viperMember: ViperMemberEntry,
+                isTrusted: Boolean,
+                isAbstract: Boolean): Unit = {
+    val key = gobraMemberKey(pkgDir, pkg, memberName, args)
+
+    val viperKey = viperMemberKey(viperMember.taskName, viperMember.memberName, pkgDir)
+    val shortViperKey = viperMemberKey(viperMember.taskName, viperMember.memberName)
```
ArquintL, line 216: why is it necessary to always perform two lookups, once with the long and once with the short key? Are there Viper members for which wo do not know to which package they belong?

src/test/scala/viper/gobra/reporting/StatsCollectorTests.scala
```diff
@@ -0,0 +1,147 @@
...
+import viper.gobra.util.{DefaultGobraExecutionContext, GobraExecutionContext}
+import viper.gobra.Gobra
+
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+
+
+class StatsCollectorTests extends AnyFunSuite with BeforeAndAfterAll {
+  val statsCollectorTestPathPropertyName = "GOBRATESTS_REGRESSIONS_DIR"
+  val statsCollectorTestDir: String = System.getProperty(statsCollectorTestPathPropertyName, "src/test/resources/stats_collector")
```
ArquintL, line 23: You can directly set the path. We only do the indirection via properties for tests that are meant to be called from external and thus should be configurable

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -42,38 +42,88 @@ object GoVerifier {
...
+    Runtime.getRuntime.addShutdownHook(new Thread() {
+      override def run(): Unit = {
+        val statsFile = config.gobraDirectory.resolve("stats.json").toFile
+        logger.info("Writing report to " + statsFile.getPath)
+        statsCollector.writeJsonReportToFile(statsFile)
+      }
+    })
+
+    config.inputPackageMap.foreach({ case (pkg, inputs) =>
+      val taskName = pkg.path + " - " + pkg.name
```
ArquintL, line 68: I don't think if I have understood the purpose of `taskName`. It's quite cumbersome to pass this around throughout Gobra but it's still an internal identifier that is meaningless to Gobra users. Can't we derive a unique identifier when we need it (e.g. by having some getter function in an object, on Config or on an AST node)

ArquintL (changes requested):

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,376 @@
...
+            ),
+            isTrusted,
+            isAbstract)
+        case _ =>
+      }
+    }
+
+    msg match {
+      // Capture typeInfo once it's available
+      case TypeInfoMessage(typeInfo, taskName) => this.synchronized({typeInfos.put(taskName, typeInfo)})
```
ArquintL, line 154: here, `synchronized` is still used

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jogasser: Note that this abstract flag is for Gobra members and not viper members. Since we don't check if imported members are abstract, an abstract gobra member will only occur once, when the package it belongs to is verified

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -116,32 +117,52 @@ object Parser {
...
+      val parsedPackage = PPackage(clause, programs.values.toVector, pom, packageDir)
+
       // The package parse tree node gets the position of the package clause:
       pom.positions.dupPos(clause, parsedPackage)
       Right(parsedPackage)
     }
 
     val parsingFn = if (config.cacheParser) { parseSourceCached _ } else { parseSource _ }
-    val parserResults = sources.map(parsingFn)
+    val parserResults = sources.map(src => (Path.of(src.name), parsingFn(src))).toMap
```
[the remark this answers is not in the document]
jogasser: I don't like the idea of storing a `Source`, inside of a node since this would mean that the contents of the source are stored in memory until that Node is not needed anymore

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -370,76 +420,87 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
...
+
+          if(path.getNameCount > 1) {
+            PackageEntry(path.getParent.toString, packageName)
+          } else {
+            PackageEntry("", packageName)
+          }
+        })
+        // Filter files by package name
+        .filter({ case (PackageEntry(_, packageName), _) => (includePackages.isEmpty || includePackages.contains(packageName)) && !excludePackages.contains(packageName)})
+        .map({ case (key, srcs) => (key, srcs.map(src => FileSource(src.name))) })
```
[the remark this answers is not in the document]
jogasser: Was badly named, I changed this whole block

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -152,9 +202,17 @@ class Gobra extends GoVerifier with GoIdeVerifier {
...
 
-  private def performTypeChecking(parsedPackage: PPackage, input: Vector[Source], config: Config): Either[Vector[VerifierError], TypeInfo] = {
+  private def performTypeChecking(taskName: String, parsedPackage: PPackage, input: Vector[Source], config: Config): Either[Vector[VerifierError], TypeInfo] = {
     if (config.shouldTypeCheck) {
-      Info.check(parsedPackage, input, isMainContext = true)(config)
+      val typeInfo = Info.check(parsedPackage, input, isMainContext = true)(config)
+
+      // Report type info for later usage
+      typeInfo match {
+        case Right(typeInfo) => config.reporter.report(TypeInfoMessage(typeInfo, taskName))
```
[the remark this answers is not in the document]
jogasser: You are right, I now use the existing type check success message. As for the taskName, I use it as a unique identifier for a verification task, that is also used by ViperServer as a key to cache verification results. This means, it should be unique and should not change between verifications. I thought using the package name and dir should be things that don't change too much

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -42,38 +42,88 @@ object GoVerifier {
...
+        statsCollector.writeJsonReportToFile(statsFile)
+      }
+    })
+
+    config.inputPackageMap.foreach({ case (pkg, inputs) =>
+      val taskName = pkg.path + " - " + pkg.name
+
+      logger.info("Verifying Package " + taskName)
+      val resultFuture = verify(config.copy(inputs = inputs, reporter = statsCollector), Some(taskName))(executor)
+      val result = Await.result(resultFuture, Duration.Inf)
```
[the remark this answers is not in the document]
jogasser: I think that this isn't possible with futures, since they start once created, which means they would still be executed in parallel or am I wrong?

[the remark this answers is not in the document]
jogasser: But I packed the reporting step into the `Future`

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -42,38 +42,88 @@ object GoVerifier {
...
+    var allErrors: Vector[VerifierError] = Vector()
+
+    // write report to file on shutdown
+    Runtime.getRuntime.addShutdownHook(new Thread() {
+      override def run(): Unit = {
+        val statsFile = config.gobraDirectory.resolve("stats.json").toFile
+        logger.info("Writing report to " + statsFile.getPath)
+        statsCollector.writeJsonReportToFile(statsFile)
+      }
+    })
```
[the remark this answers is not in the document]
jogasser: The shutdown hook makes sure that the stats get written even if an unexpected shutdown happens. This is useful if verification takes too long and has to be terminated. In that case, statistical information, for example about which method failed to terminate are still available

src/main/scala/viper/gobra/Gobra.scala
jogasser: I put the `taskName` inside of the `Config` to avoid passing it around

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jogasser: This is because of the ambiguous naming described in https://github.com/viperproject/gobra/issues/417. because of it, the assumption that a viper member can only belong to a single GobraMember is violated, which is why I resorted to this hack. There also is a comment about this in  `addResult`

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jogasser: It contains a string of the arguments for the gobra member

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jogasser: Yes I unfortunately didn't find a better method to do this

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jogasser: Good Idea!
But I didn't make use of the exact  same analysis chopper made, since it is built to isolate parts of the program.
I reused chopper's dependency method instead of writing my own

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jogasser: yes `p` is the prefix for the indentation
I am not sure if we would benefit from that since we still would have to define the exact format for the serialization and we don't care about deserialization.
This approach seems more straightforward to me

src/main/scala/viper/gobra/Gobra.scala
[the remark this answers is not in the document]
ArquintL: You can use e.g. `foldLeft` to iterate over the collection and extend the change of futures by waiting for the result of the previous future before verifying the current package (use `.map()` to do so)

src/main/scala/viper/gobra/reporting/Message.scala
ArquintL: I see, thanks! Please add a comment to this case class to explain this difference

src/main/scala/viper/gobra/reporting/StatsCollector.scala
ArquintL: I think that would simplify the code and I'm not so keen on having additional code just as an ugly workaround for a bug that should be quite easy to fix

src/main/scala/viper/gobra/frontend/Parser.scala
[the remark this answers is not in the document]
ArquintL: I see. What about having a function that takes a Source and returns a unique identifier (e.g. `SourceIdentifier` that could be a string for now and is the path for file & fromfile sources)?

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jogasser: There actually is such a case in VerifiedSCION.That is why I became aware of this issue.
But I think if it's only for the purpose of evaluation, I could as well just rename the package in VerifiedSCION, then we don't have to do this

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -370,76 +475,88 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
...
-        i <- identifyInput(input).toRight("invalid input")
-        sources <- i match {
-          case Right(filePaths) => Right(filePaths.map(f => FileSource(f.toString)))
-          case Left(_) => PackageResolver.resolveSources(RegularImport(""), moduleName, includeDirs)
-        }
-      } yield sources
-      assert(isInputOptional || res.isRight, s"validate function did not catch this problem: '${res.swap.getOrElse(None)}'")
-      res.getOrElse(Vector())
-    }
+    def convert(input: List[String], includeDirs: Vector[Path], recursive: Boolean, includePackages: List[String], excludePackages: List[String]): Map[String, Vector[Source]] = {
```
ArquintL, line 489: do you retrieve all files recursively contains in the specified folders and then group them by package name?
What happens in the case where files in different directories have the same package clause? Based on my understanding of the Go spec, the package clause does not need to be unique and there can be multiple directories that have source files specifying the same package clause

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jogasser: The viper members have no influence on this, it is produced from the `PNode` stored in the `info` field on a message

We prevent the second case from happening by defining the `isAbstract` field this way: 
` isAbstract = p.body.isEmpty && !isImported`
where `p` is a `PNode` with a body
This way we prevent false positives from imported packages, the downside is that we can't say if a member is abstract if he only appears as a imported member.

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -16,22 +16,37 @@ import org.rogach.scallop.{ScallopConf, ScallopOption, listArgConverter, singleA
...
+  val DefaultTaskName: String = "gobra-task"
+}
+
 case class Config(
                  inputs: Vector[Source],
+                 recursive: Boolean = false,
+                 gobraDirectory: Path = Path.of(ConfigDefaults.DefaultGobraDirectory),
+                 // Used as a unique identifier per verification task, ideally it shouldn't change between verifications
+                 // because it is used as a caching key, additionally
+                 taskName: String = ConfigDefaults.DefaultTaskName,
```
ArquintL, line 41: How do you ensure that different verification tasks use different task names? You pass a single `config` instance to your verify all packages functions. Doesn't this then result in all of them using the same task name?

src/main/scala/viper/gobra/frontend/PackageResolver.scala
```diff
@@ -230,22 +230,51 @@ object PackageResolver {
     } yield pkgName
   }
 
-  private lazy val pkgClauseRegex = """(?:\/\/.*|\/\*(?:.|\n)*\*\/|package(?:\s|\n)+([a-zA-Z_][a-zA-Z0-9_]*))""".r
+  private lazy val pkgClauseRegex = """(?:\/\/.*|\/\*(?:.|\n)*?\*\/|package(?:\s|\n)+([a-zA-Z_][a-zA-Z0-9_]*))""".r
```
ArquintL, line 233: please add a comment that multiline comments are lazily matches, i.e. stop at the earliest `*/`

src/main/scala/viper/gobra/frontend/PackageResolver.scala
```diff
@@ -230,22 +230,51 @@ object PackageResolver {
...
       .collectFirst { case m if m.group(1) != null => m.group(1) }
   }
 
+  def getPackageId(packageName: String, src: Source, includeDirs: Vector[Path]): String = {
+    /**
+     * Relativizes the given path with the specified include directories or the go path. The resulting path is relative
+     * to the first include directory, that is a parent of it.
+     */
+    def relativizePath(path: Path): Path = {
+      val allDirs = Properties.envOrNone("GOPATH").map(path => Path.of(path)) match {
```
ArquintL, line 249: What do you do if `GOPATH` is a semicolon separated list of paths?

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jcp19: Couldn't we instead already fix issue 417 by using the package path instead of the package name?

jogasser: Good point, maybe after this PR I could prepare a PR that removes this hack and add a note that it is to be merged after the issue is resolved?

jogasser: > Couldn't we instead already fix issue 417 by using the package path instead of the package name?

Would also be an alternative

src/main/scala/viper/gobra/frontend/PackageResolver.scala
```diff
@@ -230,22 +230,51 @@ object PackageResolver {
...
+  def getPackageClause(src: Source): Option[String] = {
 
     // TODO is there a way to perform the regex lazily on the file's content?
     pkgClauseRegex
-      .findAllMatchIn(content)
+      .findAllMatchIn(src.content)
       .collectFirst { case m if m.group(1) != null => m.group(1) }
   }
 
+  def getPackageId(packageName: String, src: Source, includeDirs: Vector[Path]): String = {
```
ArquintL, line 243: this is hard to understand without any comments what it does / its purpose is. For example, I cannot justify for myself that the resulting string is indeed unique / an identifier. For example, there could be two different includeDirs that then have the same subdirectories. In this case, your algorithm would, I think, return non-unique identifiers.
My suggestion: I would keep the includeDir in the identifier (even though the id becomes thus longer). This obviously results in cache misses if you use non-relative includeDirs but usually they are relative to the project. I'm not sure what to do in the case of `GOPATH` as we would need a solution that works across machines

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -144,23 +207,23 @@ class Gobra extends GoVerifier with GoIdeVerifier {
     inFileConfigs.foldLeft(config){ case (oldConfig, fileConfig) => oldConfig.merge(fileConfig) }
   }
 
-  private def performParsing(input: Vector[Source], config: Config): Either[Vector[VerifierError], PPackage] = {
```
ArquintL, line 147: why are these functions here no longer `private`?

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -116,32 +116,47 @@ object Parser {
...
     }
 
-    def samePackage(programs: Vector[PProgram]): Either[Vector[ParserError], Vector[PProgram]] = {
-      require(programs.nonEmpty)
-      val pkgName = programs.head.packageClause.id.name
-      val differingPkgNameMsgs = programs.flatMap(p =>
+    def samePackage(pkgIdProgramTuples: Vector[(String, PProgram)]): Either[Vector[ParserError], Vector[(String, PProgram)]] = {
+      require(pkgIdProgramTuples.nonEmpty)
+      val expectedPkgId = pkgIdProgramTuples.head._1
+      val differingPkgIdMsgs = pkgIdProgramTuples.flatMap({ case (pkgId, p) =>
```
ArquintL, line 130: `p` seems to be unused

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -116,32 +116,47 @@ object Parser {
...
       Right(parsedPackage)
     }
 
     val parsingFn = if (config.cacheParser) { parseSourceCached _ } else { parseSource _ }
-    val parserResults = sources.map(parsingFn)
+    val parserResults = sources.map(src =>
+      parsingFn(src).map(prog => (PackageResolver.getPackageId(prog.packageClause.id.name, src, config.includeDirs), prog))
+    )
+
+    assert(sources.length == parserResults.size)
```
ArquintL, line 158: Isn't this assertion trivially satisfied by the definition of `map`?

src/main/scala/viper/gobra/frontend/PackageResolver.scala
ArquintL: I'm somehow tempted to just use the directory in which a package is located as identifier. If this directory is within the project then we use a relative path and in the other case we are conservative and use the absolute path

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -348,23 +370,22 @@ object Parser {
       * Optionally consumes nested curly brackets with arbitrary content if `specOnly` is turned on, otherwise optionally applies the parser `p`
       */
     def specOnlyParser[T](isPure: Boolean, p: Parser[T]): Parser[Option[T]] =
-      if (specOnly && !isPure) nestedCurlyBracketsConsumer.? ^^ (_.flatten)
+      if (specOnly && !isPure) nestedCurlyBracketsConsumer.? ^^(_.flatten)
```
ArquintL, line 373: unnecessary change

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -348,23 +370,22 @@ object Parser {
...
       else p.?
 
     /**
       * Consumes nested curly brackets with arbitrary content and returns None
       */
     lazy val nestedCurlyBracketsConsumer: Parser[Option[Nothing]] =
       "{" ~> ("""[^{}]""".r | nestedCurlyBracketsConsumer).* <~ "}" ^^ (_ => None)
-
```
ArquintL, line 359: same here

src/main/scala/viper/gobra/reporting/Message.scala
```diff
@@ -109,7 +115,7 @@ sealed trait TypeCheckMessage extends GobraMessage {
     s"files=$inputs)"
 }
 
-case class TypeCheckSuccessMessage(inputs: Vector[String], ast: () => PPackage, erasedGhostCode: () => String, goifiedGhostCode: () => String) extends TypeCheckMessage {
+case class TypeCheckSuccessMessage(inputs: Vector[String], taskName: String, typeInfo: () => TypeInfo, ast: () => PPackage, erasedGhostCode: () => String, goifiedGhostCode: () => String) extends TypeCheckMessage {
   override val name: String = s"type_check_success_message"
 
   override def toString: String = s"type_check_success_message(" +
```
ArquintL, line 121: please add `taskName`

src/main/scala/viper/gobra/reporting/Message.scala
```diff
@@ -51,24 +57,24 @@ sealed trait GobraEntityResultMessage extends GobraVerificationResultMessage {
...
     s"concerning=${concerning.toString}, " +
-    s"cached=${cached.toString})"
+    s"cached=$cached)"
 }
 
-case class GobraEntityFailureMessage(verifier: String, entity: vpr.Member, concerning: Source.Verifier.Info, result: VerifierResult, cached: Boolean) extends GobraEntityResultMessage {
+case class GobraEntityFailureMessage(taskName: String, verifier: String, entity: vpr.Member, concerning: Source.Verifier.Info, result: VerifierResult, time: Time, cached: Boolean) extends GobraEntityResultMessage {
   override val name: String = s"entity_failure_message"
 
   override def toString: String = s"entity_failure_message(" +
```
ArquintL, line 73: please add `taskName`

src/main/scala/viper/gobra/reporting/Message.scala
```diff
@@ -51,24 +57,24 @@ sealed trait GobraEntityResultMessage extends GobraVerificationResultMessage {
   val cached: Boolean
 }
 
-case class GobraEntitySuccessMessage(verifier: String, entity: vpr.Member, concerning: Source.Verifier.Info, cached: Boolean) extends GobraEntityResultMessage {
+case class GobraEntitySuccessMessage(taskName: String, verifier: String, entity: vpr.Member, concerning: Source.Verifier.Info, time: Time, cached: Boolean) extends GobraEntityResultMessage {
   override val name: String = s"entity_success_message"
   val result: VerifierResult = Success
 
   override def toString: String = s"entity_success_message(" +
```
ArquintL, line 64: please add `taskName`

src/main/scala/viper/gobra/reporting/StatsCollector.scala
ArquintL: > > Couldn't we instead already fix issue 417 by using the package path instead of the package name?
> 
> Would also be an alternative

That should be the easiest solution. It's only about adapting the `name` and `nameWithoutScope` functions in the NameManager class (in Desugarer.scala). Replace the package name by a unique package identifier

src/test/scala/viper/gobra/GobraPackageTests.scala
```diff
@@ -99,7 +100,10 @@ class GobraPackageTests extends GobraTests {
       // set throwError to true: Scallop will throw an exception instead of terminating the program in case an
       // exception occurs (e.g. a validation failure)
       throwError.value = true
-      Some(new ScallopGobraConfig(args.toSeq).config)
+
+      // Simulate pick of package, Gobra normally does
+      val config = new ScallopGobraConfig(args.toSeq).config
+      Some(config.copy(inputs = config.inputPackageMap.values.head))
```
ArquintL, line 106: why do you need to modify the inputs?

src/main/scala/viper/gobra/Gobra.scala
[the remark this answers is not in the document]
ArquintL: exactly. That's how it's done in the chopper as well

src/main/scala/viper/gobra/reporting/StatsCollector.scala
ArquintL: I see. For documentation purposes and reusability, I think it would be better to keep the data as structures as possible, e.g. to use `Vector[PParameter]` instead of a plain string

src/main/scala/viper/gobra/reporting/StatsCollector.scala
ArquintL: Thanks. I missed this part. Maybe renaming `isAbstract` to `isAbstractAndNotImported` would help to make the code more readable

src/main/scala/viper/gobra/Gobra.scala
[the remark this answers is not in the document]
jogasser: Okay, I see what you mean now, but I won't be able to timeout these futures on a package level(see updated code) because I don't think there is a different solution for timeouts than a blocking call like `Await.result`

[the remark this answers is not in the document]
ArquintL: This is true. However, you're currently using `Duration.Inf` anyway

src/main/scala/viper/gobra/frontend/PackageResolver.scala
jogasser: Good catch, I didn't have in mind that it probably isn't just a single path

src/main/scala/viper/gobra/frontend/Parser.scala
jogasser: should be, it is used to get the package clause one line later

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,405 @@
...
+         |$p$i"success": ${this.success},
+         |$p$i"cached": ${this.cached},
+         |$p$i"fromImport": ${this.fromImport},
+         |$p$i"hasBody": ${this.hasBody},
+         |$p$i"verified": ${this.verified}
+         |$p}""".stripMargin
+    }
+  }
+
+  case class GobraMemberInfo(pkgId: String,
```
ArquintL, line 113: why do you replicate parts of the information that we store in the AST? Couldn't you simply store the internal or parse AST node?

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,405 @@
...
+      this.viperMembers.values
+        .flatMap(viperMember => viperMember.dependencies
+          .flatMap(dep => findGobraMemberByViperMemberName(viperMember.taskName, dep, info.pkgId)))
+        .filter(_.key != this.key).toSet
+    }
+
+    def key: String = gobraMemberKey(info.pkgId, info.memberName, info.args)
+  }
+
+  case class ViperMemberEntry(memberName: String,
```
ArquintL, line 87: same here for the Viper AST node

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jogasser: I worry about memory issues when storing the ASTs, particularly viper ASTs because of all the subnodes that then don't get garbage collected

src/main/scala/viper/gobra/Gobra.scala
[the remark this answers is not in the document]
jogasser: No, GitHub unfortunatelly doesn't update this view, the current version looks like this:
```
      try {
        Await.result(future, config.packageTimeout)
      } catch {
        case _: TimeoutException => logger.error("The verification of package " + pkgId + " got terminated after " + config.packageTimeout.toString)
      }
```

src/main/scala/viper/gobra/frontend/Config.scala
jogasser: `verifyAllPackages` loops over the `inputPackageMap` of the config, where we grouped the input files by `packageId`. It then copies config and overwrites `inputs`, `taskName` and `reporter` fields. It uses the `packageId` as a taskName which should be unique. Otherwise, I don't provide any guarantees

src/test/scala/viper/gobra/GobraPackageTests.scala
jogasser: Since the feature to pass a package as input doesn't exist anymore, it is handled via an input directory combined with the -p option. Here we have to mimic the behaviour of the `verifyAllPackages` method to make this test work.

src/main/scala/viper/gobra/reporting/StatsCollector.scala
jogasser: Added a comment for this

src/main/scala/viper/gobra/frontend/PackageResolver.scala
jogasser: I made it relative and gave the user a option to choose a project root directory, see `projectRoot` in `Config.scala`

jcp19:
@jogasser the Parser PR was already merged, so we can merge this PR after a new review and fixing the merge conflicts

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -42,8 +43,20 @@ sealed trait PUnorderedScope extends PScope
 case class PPackage(
                      packageClause: PPackageClause,
                      programs: Vector[PProgram],
-                     positions: PositionManager
+                     positions: PositionManager,
+                     id: String,
+                     isBuiltIn: Boolean = false
```
ArquintL, line 48: I think it would be more readable to use `PackageInfo` here. This would also allow you to make `viperId` a `lazy val` of PackageInfo. This would imho be more natural instead of having this here

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -3152,13 +3152,17 @@ object Desugar {
 
     private def name(postfix: String)(n: String, s: PScope, context: ExternalTypeInfo): String = {
       maybeRegister(s, context)
+      val pkgId = context.getTypeInfo.tree.originalRoot.viperId
```
ArquintL, line 3155: please extend `ExternalTypeInfo` to have a `packageInfo` getter (similar to `pkgName`)

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -3152,13 +3152,17 @@ object Desugar {
...
+
       // n has occur first in order that function inverse properly works
-      s"${n}_${context.pkgName}_$postfix${scopeMap(s)}" // deterministic
+      s"${n}_${pkgId}_$postfix${scopeMap(s)}" // deterministic
     }
 
     private def nameWithoutScope(postfix: String)(n: String, context: ExternalTypeInfo): String = {
       // n has occur first in order that function inverse properly works
-      s"${n}_${context.pkgName}_$postfix" // deterministic
+      val pkgId = context.getTypeInfo.tree.originalRoot.viperId
```
ArquintL, line 3163: same here

src/main/scala/viper/gobra/reporting/StatsCollector.scala
```diff
@@ -0,0 +1,450 @@
...
+    msg match {
+      // Capture typeInfo once it's available
+      case TypeCheckSuccessMessage(_, taskName, typeInfo, _,_,_) => typeInfos.put(taskName, typeInfo.apply())
+      // Free up unneeded space, once a task is finished
+      case VerificationTaskFinishedMessage(taskName) => typeInfos.remove(taskName)
+      // Store every viperMember with an info attached to see, which ones didn't verify on an unexpected shutdown
+      case GeneratedViperMessage(taskName, _, vprAst, _) =>
+        vprAst.apply().members
+          .filter(m => m.info.isInstanceOf[Source.Verifier.Info] && isRelevantInfo(m.info.asInstanceOf[Source.Verifier.Info]))
+          .foreach(m => addViperMember(taskName, m, m.info.asInstanceOf[Source.Verifier.Info], 0, cached = false, verified = false, success = true))
```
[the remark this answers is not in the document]
jogasser: Not sure how exactly you mean that, I think that would not work since I also use `m` as am argument to `addViperMember`

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -43,12 +42,83 @@ object GoVerifier {
...
+        statsCollector.writeJsonReportToFile(statsFile)
+
+        val timedOutMembers = statsCollector.getTimedOutMembers
+        if(timedOutMembers.nonEmpty) {
+          timedOutMembers.foreach(member => logger.error(s"The verification of member $member did not terminate"))
+        }
+      }
+    })
+
+    config.packageInfoInputMap.foreach({ case (pkgInfo, inputs) =>
```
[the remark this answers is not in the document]
jogasser: That's a good idea, it involved quite some changes but I also thinks it makes it less confusing

src/main/scala/viper/gobra/reporting/StatsCollector.scala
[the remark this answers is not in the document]
ArquintL: That's true. You can leave it as is. Otherwise, we would need to pass the tuple `(m, m.info)` which seems a bit redundant either

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -43,56 +41,123 @@ object GoVerifier {
...
+    logger.info(statsCollector.getNumberOfVerifiedMembers + " have specification")
+    logger.info(statsCollector.getNumberOfVerifiedMembersWithAssumptions + " are assumed to be satisfied")
+
+    allErrors
   }
 
-  protected[this] def verify(input: Vector[Source], config: Config)(executor: GobraExecutionContext): Future[VerifierResult]
+  /*def verify(config: Config)(executor: GobraExecutionContext): Future[VerifierResult] = {
+    verify(config.inputs, config)(executor)
+  }*/
```
ArquintL, line 123: please delete

ArquintL (changes requested):
Thanks a lot for addressing my earlier comments! I only have 3 more that are just about deleting / moving / renaming stuff. Afterwards, this PR can imho be merged and I'm already looking forward to having it in the master branch. In any case, please focus on your thesis first

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -39,10 +40,31 @@ object PNode {
...
 sealed trait PUnorderedScope extends PScope
 
+/**
+ * Contains information about a package
+ *
+ * @param id a unique identifier for the package
+ * @param name the name of the package, does not have to be unique
+ * @param isBuiltIn a flag indicating, if the package comes from within Gobra
+ */
+case class PPackageInfo(id: String, name: String, isBuiltIn: Boolean) extends PNode {
```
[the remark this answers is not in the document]
jogasser: Me too, but when we want to attach it to the `PPackage` directly, it needs to extend `PNode` because apparently every object in a `PNode` is  assumed to be part of the AST and tried to be cast to a `PNode` which leads to errors if it does not extend it.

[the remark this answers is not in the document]
ArquintL: Weird. Why does it work for `PositionManager`?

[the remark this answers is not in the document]
jogasser: I did some digging and found the reason why documented in Kiama's `Tree` class: 

```
 * The `originalRoot` value should be the root of a finite, acyclic structure
 * that contains only `Product` instances of type `T` (unless they are skipped,
 * see below).
 *
```

`case classes` are always instances of the type  `Product`. Therefore, `PPackage` may not contain `PackageInfo` without it extending `PNode`. An alternative would be to change `PackageInfo` to a class and implement custom `equals` and `hashCode` methods as we depend on them.

jogasser:
The Gobra IDE plugin probably needs to be adjusted after this is merged, since Gobra's `verify` method signatures changed

ArquintL (commented):
Looks good to me. @jcp19 any comments from your side?

ArquintL:
Wait what? I can push to Johannes' fork when I do so via GitHub UI?!? At least the CI is now (partially) running, i.e. executes PR workflow but for whatever reason not the push workflow. This was for some reason not the case for Johannes' latest commit

ArquintL (approved):

PR #399, opened 2022-01-25
Parser Cache Fix

ArquintL:
Currently, the parser cache caches parse ASTs. However, it does not take the `specOnly` flag into account, meaning that there could be the situation of parsing a package first with `specOnly` set to true (because it's imported in a different package) and later parsing the same package with `specOnly` turned off. In this case, the cache would incorrectly return an AST with abstract functions & methods.
This PR fixes the cache to keep the ASTs for a given file and different  `specOnly` flags separate by including the `specOnly` flag in the key used for cache lookups.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Parser.scala
+++ b/src/main/scala/viper/gobra/frontend/Parser.scala
@@ -53,9 +53,9 @@ object Parser {
   // cache maps a key (obtained by hasing file path and file content) to the parse result
   private var sourceCache: Map[SourceCacheKey, (Either[Vector[ParserError], PProgram], Positions)] = Map.empty
 
-  /** computes the key for caching a particular source. This takes the name as well as content into account */
-  private def getCacheKey(source: Source): SourceCacheKey = {
-    val key = source.name ++ source.content
+  /** computes the key for caching a particular source. This takes the name, the specOnly flag, and the file's content into account */
+  private def getCacheKey(source: Source, specOnly: Boolean): SourceCacheKey = {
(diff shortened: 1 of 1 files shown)


src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -53,9 +53,9 @@ object Parser {
   // cache maps a key (obtained by hasing file path and file content) to the parse result
   private var sourceCache: Map[SourceCacheKey, (Either[Vector[ParserError], PProgram], Positions)] = Map.empty
 
-  /** computes the key for caching a particular source. This takes the name as well as content into account */
-  private def getCacheKey(source: Source): SourceCacheKey = {
-    val key = source.name ++ source.content
+  /** computes the key for caching a particular source. This takes the name, the specOnly flag, and the file's content into account */
```
jcp19, line 56:
```suggestion
  /** computes the key for caching a particular source. This takes the name, the specOnly flag, and the file's contents into account */
```

ArquintL: I would keep it singular as a file only has one content

jcp19: Content and contents have slightly different meanings. https://getproofed.com/writing-tips/word-choice-content-vs-contents/

ArquintL: Thanks, I was not aware of this. Sorry

PR #404, opened 2022-01-31
Fix selective import

jcp19:
Addresses #166 completely. Currently going through test phases with verified scion and some code cleanup is needed.


src/main/scala/viper/gobra/frontend/info/implementation/TypeInfoImpl.scala
```diff
@@ -112,6 +127,24 @@ class TypeInfoImpl(final val tree: Info.GoTree, final val context: Info.Context,
     res
   }
 
+  // TODO: computes every node that might contain a visible invocation to the importing package
+  private def relevantSubnodes(n: PNode): Vector[PNode] = n match {
+    // predicates, methods, pure methods
+    case decl@ PFunctionDecl(id, args, result, spec, _) =>
+      if (decl.spec.isPure) tree.child(decl) else id +: result +: spec +: args
```
ArquintL, line 134: Isn't this equivalent to `tree.child(decl)` because the parser will throw away the body in case of an imported non-pure function & method?

ArquintL (commented):
Thanks a lot for your work! I've left a comment. I also had another look at the corresponding issue and one thing that I still don't understand is why the original approach does not work. The idea was that we do typechecking lazily and simply keep track of the declarations that had to be typechecked. Our assumption was that these correspond to the declarations that are also needed for verification. Looking at the issue, it seems that the function called in a precondition is not correctly marked as being used. But why is this? Aren't we typechecking all children (incl. the precondition) when typechecking the function declaration because we have somewhere a call to this function?

src/main/scala/viper/gobra/frontend/info/implementation/TypeInfoImpl.scala
jcp19: yes, you are right!

Issue #406, opened 2022-02-04
Naming of Domains & Interfaces


ArquintL:
The internal names of domains and interfaces should not depend on their positional information such that shifting them in a file does not result in completely different Viper ASTs. Furthermore, the current naming only considers the filename, which is not unique when importing packages

Issue #17, opened 2019-10-08
Making Gobra deterministic


viper-admin:
> Created by **@Felalolf** on 2019-10-08 15:36
> Last updated on 2020-07-07 09:09

As discussed in some Viper meeting, it seems to be a good idea to make Gobra deterministic.

For Gobra this means that when Gobra is run sequentially on the same input file with the same flags set Gobra should generate identical Viper files.

Felalolf:
In theory this issue was fixed in Silas' work. We should add some tests and then the issue can be closed.

PR #407, opened 2022-02-04
More Consistent Viper ASTs

ArquintL:
This PR further improves on #17 by providing some guarantees even when slightly changing the input program. The overall idea is that whitespace changes do not affect the encoding at all or changes within a particular Gobra member only affect the encoding of that particular member.
In practice, this amounts to avoiding global counters to create fresh names. This PR localizes counter for fresh names in the Desugarer and Encoding to the current code root / member.
Furthermore, struct names are dropped from the internal representation. However, the naming of interfaces and domains is not addressed by this PR (see #406).

I have not added any tests for this feature because we would need to compare ASTs before and after performing some modifications. However, I have diffed the generated Viper files to check that shifting around a Gobra function does not change the encoding. The ordering of domain functions seems to not be stable but I don't think this affects cacheing by ViperServer in any way.

There is one design decision I would like to mention here: I've changed the translator to create a context per member. Each context shares the same generators but resets the fresh name counter to 0. There are 2 different ways how all generated members could be collected at the end: (1) assume that all contexts share the same generators and that all possibly generated members are stored in them. Thus, `finalize` could/should only be called on the `mainCtx`. (2) Alternatively and currently implemented, each member returns the context with which it was encoded and `finalize` is called on each context. This approach makes less assumptions on where it is necessary/sufficient to collect Viper members but requires that each generator is visited only once. To do so, the collector has been adapted to keep track of the generators it visits. This however requires that `finalize` is not directly called on each context but only indirectly via the collector.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/ExternalTypeInfo.scala
+++ b/src/main/scala/viper/gobra/frontend/info/ExternalTypeInfo.scala
@@ -6,7 +6,7 @@
 
 package viper.gobra.frontend.info
 
-import viper.gobra.ast.frontend.{PEmbeddedDecl, PExpression, PFieldDecl, PIdnNode, PIdnUse, PKeyedElement, PMPredicateDecl, PMPredicateSig, PMember, PMethodDecl, PMethodSig, PMisc, PNode, PParameter, PPkgDef, PScope, PType}
+import viper.gobra.ast.frontend.{PCodeRoot, PEmbeddedDecl, PExpression, PFieldDecl, PIdnNode, PIdnUse, PKeyedElement, PMPredicateDecl, PMPredicateSig, PMember, PMethodDecl, PMethodSig, PMisc, PNode, PParameter, PPkgDef, PScope, PType}
 import viper.gobra.frontend.info.base.BuiltInMemberTag.BuiltInMemberTag
 import viper.gobra.frontend.info.base.Type.{AbstractType, InterfaceT, StructT, Type}
 import viper.gobra.frontend.info.base.SymbolTable.{Embbed, Field, MPredicateImpl, MPredicateSpec, MethodImpl, MethodSpec, Regular, TypeMember}
@@ -87,4 +87,7 @@ trait ExternalTypeInfo {
 
   /** returns all implementation proofs found in the current package */
   def localImplementationProofs: Vector[(Type, InterfaceT, Vector[String], Vector[String])]
+
+  /** returns the code root for a given node; can only be called on nodes that are enclosed in a code root */
+  def codeRoot(n: PNode): PCodeRoot with PScope
 }
--- a/src/main/scala/viper/gobra/translator/Names.scala
+++ b/src/main/scala/viper/gobra/translator/Names.scala
@@ -44,7 +44,7 @@ object Names {
     // we use a dollar sign to mark the beginning and end of the type list to avoid that `Tuple(Tuple(X), Y)` and `Tuple(Tuple(X, Y))` map to the same name:
     case in.TupleT(ts, _) => s"Tuple$$${ts.map(serializeType).mkString("")}$$"
     case in.PredT(ts, _) => s"Pred$$${ts.map(serializeType).mkString("")}$$"
-    case in.StructT(name, fields, _) => s"Struct$name${serializeFields(fields)}"
+    case in.StructT(fields, _) => s"Struct${serializeFields(fields)}"
     case in.InterfaceT(name, _) => s"Interface$name"
     case in.ChannelT(elemT, _) => s"Channel${serializeType(elemT)}"
     case t => Violation.violation(s"cannot stringify type $t")
--- a/src/main/scala/viper/gobra/translator/Names.scala
+++ b/src/main/scala/viper/gobra/translator/Names.scala
@@ -6,17 +6,15 @@
 
 package viper.gobra.translator
 
+import viper.gobra.ast.{internal => in}
+import viper.gobra.translator.interfaces.Context
+import viper.gobra.util.Violation
 import viper.silver.{ast => vpr}
 
 object Names {
   def returnLabel: String = "returnLabel"
 
-  private var freshCounter = 0
-  def freshName: String = {
-    val str = s"fn$$$$$freshCounter"
-    freshCounter += 1
-    str
-  }
+  def freshName(ctx: Context): String = s"fn$$$$${ctx.getAndIncrementFreshCounter}"
 
   /* sanitizes type name to a valid Viper name */
   def serializeType(t: vpr.Type): String = {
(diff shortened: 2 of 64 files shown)


src/main/scala/viper/gobra/translator/implementations/ContextImpl.scala
```diff
@@ -113,8 +115,18 @@ case class ContextImpl(
...
+    stmtN,
+    initialFreshCounterValue = initialFreshCounterValueN
   )
 
   override def addVars(vars: LocalVarDecl*): Context = this
+
+
+  private var freshCounter: Int = initialFreshCounterValue
+  override def getFreshCounter: Int = freshCounter
+  override def getAndIncrementFreshCounter: Int = {
```
Felalolf, line 127: it returns a counter that is counting fresh values and not a counter that is fresh, right? I find the name a bit confusing.

src/main/scala/viper/gobra/translator/implementations/CollectorImpl.scala
```diff
@@ -7,16 +7,28 @@
...
   protected var _fields: List[vpr.Field] = List.empty
   protected var _predicates: List[vpr.Predicate] = List.empty
   protected var _functions: List[vpr.Function] = List.empty
   protected var _methods: List[vpr.Method]  = List.empty
   protected var _extensions: List[vpr.ExtensionMember] = List.empty
 
+  /** invokes finalize on each generator */
+  override def finalize(generators: Vector[Generator]): Unit = {
+    generators foreach { generator =>
+      if(!_visitedGenerators.contains(generator)) {
```
Felalolf, line 25: is this just for safety, or are there actually cases where some generators are duplicate?

src/main/scala/viper/gobra/translator/encodings/structs/StructEncoding.scala
```diff
@@ -243,22 +243,22 @@ class StructEncoding extends TypeEncoding {
...
+    *   ensures AND (f: T) in F. [&res.f == dflt(T)]
     */
   private val shDfltFunc: FunctionGenerator[Vector[in.Field]] = new FunctionGenerator[Vector[in.Field]] {
     override def genFunction(fs: Vector[in.Field])(ctx: Context): vpr.Function = {
-      val resType = in.StructT("the name does not matter", fs, Shared)
+      val resType = in.StructT(fs, Shared)
       val vResType = typ(ctx)(resType)
       val src = in.DfltVal(resType)(Source.Parser.Internal)
-      val resDummy = in.LocalVar(Names.freshName, resType)(src.info)
+      val resDummy = in.LocalVar("res", resType)(src.info)
```
Felalolf, line 254: Is the justification that using "res" is fine because there are no other variables? Please, write this as a comment.

src/main/scala/viper/gobra/translator/encodings/arrays/SharedArrayComponentImpl.scala
```diff
@@ -29,23 +29,23 @@ class SharedArrayComponentImpl extends SharedArrayComponent {
...
+      val idxEq = vpr.EqCmp(resAccess, pure(ctx.expr.translate(in.DfltVal(t.elemT)(src.info))(ctx))(ctx).res)()
       val forall = vpr.Forall(
         Seq(vIdx),
         Seq(vpr.Trigger(Seq(resAccess))()),
         idxEq
       )()
 
       vpr.Function(
-        name = s"${Names.arrayDefaultFunc}_${Names.freshName}",
+        name = s"${Names.arrayNilFunc}_${t.serialize}",
```
Felalolf, line 48: good catch

src/main/scala/viper/gobra/translator/encodings/arrays/ArrayEncoding.scala
```diff
@@ -291,23 +296,23 @@ class ArrayEncoding extends TypeEncoding with SharedArrayEmbedding {
...
-    def genFunction(t: (BigInt, in.Type))(ctx: Context): vpr.Function = {
-      val argType = in.ArrayT(t._1, t._2, Shared)
+  private val conversionFunc: FunctionGenerator[ComponentParameter] = new FunctionGenerator[ComponentParameter]{
+    def genFunction(t: ComponentParameter)(ctx: Context): vpr.Function = {
+      val argType = t.arrayT(Shared)
       val x = in.LocalVar("x", argType)(Source.Parser.Internal)
       val resultType = argType.withAddressability(Exclusive)
       val vResultType = typ(ctx)(resultType)
-      val resultVar = in.LocalVar(Names.freshName, resultType)(Source.Parser.Internal)
+      val resultVar = in.LocalVar("res", resultType)(Source.Parser.Internal)
```
Felalolf, line 309: Similar to some other comment. How do you know that there is no other variable with the name "res". Is it because there are no other variables? Please write the reason as a comment.

src/main/scala/viper/gobra/translator/encodings/arrays/ArrayEncoding.scala
```diff
@@ -320,32 +325,32 @@ class ArrayEncoding extends TypeEncoding with SharedArrayEmbedding {
...
-  private val exDfltFunc: FunctionGenerator[(BigInt, in.Type)] = new FunctionGenerator[(BigInt, in.Type)]{
-    def genFunction(t: (BigInt, in.Type))(ctx: Context): vpr.Function = {
-      val resType = in.ArrayT(t._1, t._2, Exclusive)
+  private val exDfltFunc: FunctionGenerator[ComponentParameter] = new FunctionGenerator[ComponentParameter]{
+    def genFunction(t: ComponentParameter)(ctx: Context): vpr.Function = {
+      val resType = t.arrayT(Exclusive)
       val vResType = typ(ctx)(resType)
       val src = in.DfltVal(resType)(Source.Parser.Internal)
-      val resDummy = in.LocalVar(Names.freshName, resType)(src.info)
+      val resDummy = in.LocalVar("res", resType)(src.info)
```
Felalolf, line 337: same here

src/main/scala/viper/gobra/translator/implementations/translator/ProgramsImpl.scala
```diff
@@ -45,15 +47,17 @@ class ProgramsImpl extends Programs {
...
       }
+      memberWriter.map(m => (m, ctx))
     }
 
     val progW = for {
-      memberss <- sequence(program.members map goM)
-      members = memberss.flatten
+      membersWithCtxs <- sequence(program.members map goM)
+      ctxs = membersWithCtxs map { case (_, ctx) => ctx }
+      members = membersWithCtxs flatMap { case (m, _) => m }
```
Felalolf, line 56: You could use unzip and then flatten.

Felalolf (changes requested):

src/main/scala/viper/gobra/translator/implementations/CollectorImpl.scala
ArquintL: I like the proposed alternative

src/main/scala/viper/gobra/translator/implementations/ContextImpl.scala
ArquintL: I call it now `getAndIncrementFreshVariableCounter`. I hope this is slightly better

src/main/scala/viper/gobra/translator/implementations/translator/ProgramsImpl.scala
ArquintL: Thanks, that's exactly what I was looking for. Somehow I did not spot `unzip`

ArquintL:
This PR just got a lot bigger, sorry. I had to touch every/almost every encoding file after changing the way we collect generated Viper members

src/main/scala/viper/gobra/frontend/info/ExternalTypeInfo.scala
```diff
@@ -87,4 +87,7 @@ trait ExternalTypeInfo {
 
   /** returns all implementation proofs found in the current package */
   def localImplementationProofs: Vector[(Type, InterfaceT, Vector[String], Vector[String])]
+
+  /** returns the code root for a given node; can only be called on nodes that are enclosed in a code root */
+  def codeRoot(n: PNode): PCodeRoot with PScope
```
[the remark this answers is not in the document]
ArquintL: We definitely could but this would mean yet another trait to add to most AST nodes. Even if we would do so, the type checker cannot give us any guarantees that `codeRoot` indeed does not cause an exception (we would have this assumption implicitly be carefully mixing the trait into AST nodes below code root

src/main/scala/viper/gobra/translator/Names.scala
```diff
@@ -6,17 +6,16 @@
...
 object Names {
   def returnLabel: String = "returnLabel"
 
-  private var freshCounter = 0
-  def freshName: String = {
-    val str = s"fn$$$$$freshCounter"
-    freshCounter += 1
-    str
-  }
+  def freshName(ctx: Context): String = s"fn$$$$${ctx.getAndIncrementFreshVariableCounter}"
```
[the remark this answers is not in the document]
ArquintL: I like the idea of having an infinite iterator. I'm however not sure how exactly you want to separate the state because the iterator's current value depends on the context or rather the iterator's initial value should be settable by the context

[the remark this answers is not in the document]
ArquintL: @jcp19 could you have a look at `Context.scala` and `ContextImpl.scala` and let me know whether you like the solution?

Felalolf (approved):

src/main/scala/viper/gobra/translator/Names.scala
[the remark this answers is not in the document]
jcp19: Fine by me

src/main/scala/viper/gobra/translator/implementations/CollectorImpl.scala
```diff
@@ -21,15 +21,15 @@ class CollectorImpl extends Collector {
 
   /** invokes finalize on each generator */
   override def finalize(generators: Vector[Generator]): Unit = {
-    generators foreach { generator =>
+    for (generator <- generators) {
```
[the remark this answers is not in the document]
ArquintL: That's easy. I can fix it

PR #412, opened 2022-02-10
Fixes issue #411

### LINKED ISSUE #411: Chopper loses isolated node.
Chopper prints that an isolated node is lost in attached file.

[pkg1.txt](https://github.com/viperproject/gobra/files/8041868/pkg1.txt)

Felalolf:
Fixes #411. The problem was that dfs stopped when it found an already visited node.

### DIFF
--- a/src/main/scala/viper/gobra/util/ViperChopper.scala
+++ b/src/main/scala/viper/gobra/util/ViperChopper.scala
@@ -199,7 +199,7 @@ object ViperChopper {
         * and then returns for each dominating node, the set of reachable nodes in a separate sorted list.
         */
 
-      // Stores whether a node was visited or not.
+      // Stores whether a node was visited by any call to dfs.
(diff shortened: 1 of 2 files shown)


src/main/scala/viper/gobra/util/ViperChopper.scala
```diff
@@ -69,7 +69,7 @@ object ViperChopper {
       var forallSmallNode: (Int => Boolean) => Boolean = null // for the safety check
 
       val result = {
-        if (nodes.size < 10) {
+        if (nodes.size <= 2) {
```
jcp19, line 72: Won't this make it very unlikely to ever execute the non-SCC algorithm? If that is the case, I wonder if we should keep this algorithm that only kicks in in small programs

jcp19: Yes, you are right. Thanks!

ArquintL (approved):

PR #413, opened 2022-02-10
Fix issue 238

### LINKED ISSUE #238: Access to a slice pointer should yield access to its cells
In a conversation with @Felalolf, we came to the idea that `acc(s)` should yield access to all `s`'s cells when `s` is of pointer to slice type, i.e. `*[]T`, for arbitrary `T`. This would be consistent with what is done for pointers to structs and its fields.

jcp19:
Adds a short-hand for representing access to all elements of a slice. If the underlying type of `s` is `[]T`, then `acc(s)` is equivalent to `forall i int :: { s[i] } 0 <= i && i < len(s) ==> acc(&s[i])`.

An alternative design would be to translate it to `forall i int :: { s[i] } 0 <= i && i < cap(s) ==> acc(&s[i])`, but my experience with Gobra tells me that this is not often desired.

It might be better to merge this after merging PR #407 to already use the new support for generating names in contexts.

Closes #238

### DIFF
--- a/src/main/scala/viper/gobra/ast/internal/Program.scala
+++ b/src/main/scala/viper/gobra/ast/internal/Program.scala
@@ -1186,7 +1186,7 @@ case class SliceT(elems : Type, addressability: Addressability) extends PrettyTy
   */
 case class MapT(keys: Type, values: Type, addressability: Addressability) extends Type {
   def hasGhostField(k: Type): Boolean = k match {
-    case StructT(_, fields, _) => fields exists (_.ghost)
+    case StructT(fields, _) => fields exists (_.ghost)
     case _ => false
   }
   // this check must be done here instead of at the type system level because the concrete AST does not support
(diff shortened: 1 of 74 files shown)


src/main/scala/viper/gobra/translator/implementations/translator/AssertionsImpl.scala
```diff
@@ -44,8 +46,24 @@ class AssertionsImpl extends Assertions {
...
+                    in.And(
+                      in.AtMostCmp(in.IntLit(0)(acc.info), iterVar)(acc.info),
+                      in.LessCmp(iterVar, in.Length(op)(acc.info))(acc.info))(acc.info),
+                    in.Access(
+                      in.Accessible.Address(in.IndexedExp(op, iterVar, underlyingType)(acc.info)),
+                      acc.p
+                    )(acc.info)
+                  )(acc.info)
+                )(acc.info)
+                ctx.ass.translate(quantifiedAssert)(ctx)
```
Felalolf, line 66: I would move this into the slice encoding (like the case for the map is in the map encoding).

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -376,7 +376,7 @@ sealed trait Accessible extends Node {
 
 object Accessible {
   case class Predicate(op: PredicateAccess) extends Accessible with TriggerExpr
-  case class ExprAccess(op: Expr) extends Accessible
+  case class ExprAccess(op: Expr, underlyingType: Type) extends Accessible
```
Felalolf, line 379: I do not think you need to add the underlying type since we can get this with the underlyingType function of the Context (which uses the LookupTable internally).

Felalolf (changes requested):
Two small things, once they are done, this should be fine.

Felalolf (approved):
See comments.

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -47,6 +47,28 @@ class SliceEncoding(arrayEmb : SharedArrayEmbedding) extends LeafTypeEncoding {
...
     }
   }
+  /**
+    * Encodes assertions.
+    * [acc(m: []T, perm)] -> [forall i int :: 0 <= i && i < len(m) ==> acc(&m[i], perm)]
+    */
+  override def assertion(ctx: Context): in.Assertion ==> CodeWriter[vpr.Exp] = {
+    default(super.assertion(ctx)) {
+      case n@ in.Access(in.Accessible.ExprAccess(exp :: ctx.Slice(elem)), perm) =>
+        val iterVar = in.BoundVar(Names.freshName, in.IntT(Addressability.Exclusive))(n.info)
```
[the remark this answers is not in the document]
jcp19: Yes, that is what I hinted in the PR message.

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -47,6 +47,28 @@ class SliceEncoding(arrayEmb : SharedArrayEmbedding) extends LeafTypeEncoding {
...
+        val underlyingType = in.SliceT(elem, Addressability.exprInAcc)
+        val quantifiedAssert = in.SepForall(
+          vars = Vector(iterVar),
+          triggers = Vector(in.Trigger(Vector(in.IndexedExp(exp, iterVar, underlyingType)(n.info)))(n.info)),
```
Felalolf, line 61: should the trigger be `m[i]` or `&m[i]`. To my understanding, `m[i]` is a heap trigger and in theory is better. However, my trust in heap trigger is not that big. You have experience with this in practice; Which of the two does work better for you?

PR #414, opened 2022-02-14
Further Improve Consistency of Encoding

ArquintL:
This PR addresses the following two sources of inconsistency:
- interface implementations are not ordered and thus postconditions or if statements appear in different orders (fix: use `SortedSet` instead of `Set`)
- Generation of functions in the `Types` domain happens in non-deterministic order. Thus, the tag assigned to each type might change across encodings (fix: introduction of a unique `tag` domain function)

@jogasser This PR should address the cache misses reported by you

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Desugar.scala
+++ b/src/main/scala/viper/gobra/frontend/Desugar.scala
@@ -21,7 +21,7 @@ import viper.gobra.util.Violation.violation
 import viper.gobra.util.{DesugarWriter, TypeBounds, Violation}
 
 import scala.annotation.{tailrec, unused}
-import scala.collection.Iterable
+import scala.collection.{Iterable, SortedSet}
 import scala.reflect.ClassTag
 
 object Desugar {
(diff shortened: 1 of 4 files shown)


src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -2261,11 +2261,11 @@ object Desugar {
       }
     }
 
-    lazy val interfaceImplementations: Map[in.InterfaceT, Set[in.Type]] = {
+    lazy val interfaceImplementations: Map[in.InterfaceT, SortedSet[in.Type]] = {
       info.interfaceImplementations.map{ case (itfT, implTs) =>
         (
           interfaceType(typeD(itfT, Addressability.Exclusive)(Source.Parser.Unsourced)).get,
-          implTs.map(implT => typeD(implT, Addressability.Exclusive)(Source.Parser.Unsourced))
+          SortedSet(implTs.map(implT => typeD(implT, Addressability.Exclusive)(Source.Parser.Unsourced)).toSeq: _*)
```
Felalolf, line 2268: Is this how a data structure is changed into a sorted set. I would have expected some function that can take an Iterable directly.

src/main/scala/viper/gobra/translator/encodings/interfaces/TypeComponentImpl.scala
```diff
@@ -267,12 +269,19 @@ class TypeComponentImpl extends TypeComponent {
...
+        formalArgs = Seq.empty,
+        typ = vpr.Int,
+        unique = true
+      )(domainName = domainName)
+      val tagValueFuncApp = vpr.DomainFuncApp(func = tagValueFunc, Seq.empty, Map.empty)()
+
       val tagAxiom = {
         val tagApp = vpr.DomainFuncApp(func = tagFunc, Seq(funcApp), Map.empty)()
-        val eq = vpr.EqCmp(tagApp, vpr.IntLit(tag)())()
+        val eq = vpr.EqCmp(tagApp, tagValueFuncApp)()
```
Felalolf, line 284: This feels a bit weird. Could you not just collect all functions necessary and then generate them at the finalizing step.

Felalolf: Yes, I am worried about performance.

Felalolf: if there is no performance drop, then you can merge.

PR #421, opened 2022-03-06
Gitsubmodules

Dspil:
This PR tries to simplify the installation process of gobra. The dependencies silver, silicon, carbon and viperserver are now submodules and a script checking out the proper commits and making the symbolic links between them has been added.


link_dependencies.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/bash
+
+git submodule update --init
+cd silver
+git checkout 7228e7144d41c91f02a70a68a93b6b3efae57d14
```
Aurel300, line 5: None of the submodules pushed to the repo in this PR match the commits specified here. Is this intentional?

Dspil: I got the commit checksums from viper-toolchain-versions.sh

ArquintL: You could source `viper-toolchain-versions.sh` and then use the defined variables such that we do not have commit hashes in this file

ArquintL: Or even better, is there a way to read the commit hash from the submodule? Then we could get rid of `viper-toolchain-versions.sh`

Aurel300: You can `cd` into the submodule and use `git rev-parse HEAD` to get the long commit hash and nothing else on stdout.

ArquintL: Isn't that the whole point of the submodules? That the submodules specify the versions with which the current codebase is supposed to be run against. Why else do we want to have submodules?

ArquintL: > But don't we use a commit that is not the latest one? We need to specify this somehow

That's correct but this should imho be reflected by the submodules

Aurel300: Git submodules are always locked to a particular hash. If you look at the "Files changed" summary for this PR, you can see e.g. "Submodule [silver](https://github.com/viperproject/silver) added at `25f60d`". You can choose which commit to push to the repository, e.g.:
```
$ cd silver
$ git checkout 7228e7144d41c91f02a70a68a93b6b3efae57d14
$ cd ..
$ git add silver
$ git commit -m "..."
```
Instead of storing the hashes in a shell script, the hash can be "stored" with the submodule itself.

Dspil: oh I see, I will change that

Dspil:
The checks fail now because viper-toolchain-versions.sh does not exist

PR #423, opened 2022-03-17
Fix some compiler warnings

jcp19:
The last PRs introduced some compiler warnings. This PR addresses those that I could fix quickly, and introduces some simplifications. The remaining issues have to do with the way we pattern match against nested generic types, which is not sound due to erasure. I could just introduce an annotation to ignore those, but I would rather go with a better solution.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/InformativeErrorListener.scala
+++ b/src/main/scala/viper/gobra/frontend/InformativeErrorListener.scala
@@ -15,6 +15,7 @@ import viper.gobra.reporting.ParserError
 import viper.silver.ast.SourcePosition
 
 import java.nio.file.Path
+import scala.annotation.unused
 import scala.collection.mutable.ListBuffer
 
 class InformativeErrorListener(val messages: ListBuffer[ParserError], val source: Source) extends BaseErrorListener {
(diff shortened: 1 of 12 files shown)


src/main/scala/viper/gobra/frontend/InformativeErrorListener.scala
```diff
@@ -77,15 +76,14 @@ class InformativeErrorListener(val messages: ListBuffer[ParserError], val source
     *
     * @see [[FailedPredicateException]]
     * @param context
-    * @param exception
     * @return
     */
-  def analyzeFailedPredicate(implicit context: ParserErrorContext, exception: FailedPredicateException): ErrorType = {
+  def analyzeFailedPredicate(implicit context: ParserErrorContext): ErrorType = {
```
ArquintL, line 81: I'm not sure about this change here, because the exception is the only parameter that somewhat bounds this error handler to the mentioned error. Maybe an unused annotation would be better suited?

src/main/scala/viper/gobra/frontend/InformativeErrorListener.scala
```diff
@@ -120,10 +118,9 @@ class InformativeErrorListener(val messages: ListBuffer[ParserError], val source
     *
     * @see [[InputMismatchException]]
     * @param context The context of the error
-    * @param exception
     * @return
     */
-  def analyzeInputMismatch(implicit context: ParserErrorContext, exception: InputMismatchException): ErrorType = {
+  def analyzeInputMismatch(implicit context: ParserErrorContext): ErrorType = {
```
ArquintL, line 123: same here

ArquintL (approved):

Issue #302, opened 2021-06-14
Generated Viper AST Should Be Consistent


ArquintL:
Currently, consistency checks do not pass for all generated Viper ASTs.
The encoding should be improved and the following line in `GobraTests` should be uncommented such that consistency is checked for all our regression tests (it is already enabled for `GobraPackageTests`):
```
backendConfig = ViperVerifierConfig.EmptyConfig(performConsistencyChecks = true)
```

jcp19:
Looking at the slice encoding might be a good first step. A simple program like the following does not pass the consistency tests.
```go
package p

pure func f(a, b []byte) bool {
	return true
}
```

This is the corresponding error message:
```
[info] An assumption was violated during execution.
[info] Logic error: List(Consistency error: expected the same type, but got Slice[T] and Slice[T] (<no position>))
[info] viper.gobra.util.Violation$LogicException: Logic error: List(Consistency error: expected the same type, but got Slice[T] and Slice[T] (<no position>))
```

Also, the stubs for the std-lib also do not pass these consistency checks (PR #425)

PR #425, opened 2022-03-28
Check std-lib stubs when testing

jcp19:
So far, our test pipeline has never tested the stubs for the std library and breaking changes might have gone unnoticed. This PR addresses that issue.

Currently failing because the stub files do not pass the consistency checks


jcp19:
Should be merged when #302 is closed

src/main/resources/stubs/github.com/scionproto/scion/go/lib/assert/assert.gobra
```diff
@@ -1,13 +0,0 @@
-// This Source Code Form is subject to the terms of the Mozilla Public
```
ArquintL, line 1: why do you want to remove this file? There is a test case that relies on it

ArquintL (approved):
Looks good but why does CI fail (besides the test case using assert)? Consistency checks are disabled by default

src/main/resources/stubs/github.com/scionproto/scion/go/lib/assert/assert.gobra
jcp19: Mostly, because it does not make sense to have this file as a stub, given that we expect to never use it

ArquintL: Because we prove that an assertion always holds and thus we don't need to check it at runtime?

jcp19: What do you mean?
Actually, this file does not seem to be used in any project (not even scion), and it is also not a built-in or part of the std-lib. As such, having builtin support in Gobra to `import  "github.com/scionproto/scion/go/lib/assert"` is unnecessary.

However, I will keep this file but move to the tests such that we can keep the tests that depend on it

ArquintL: I think I (?) / we have included it because SCION was using it. If it no longer thus then we can definitely remove it.

PR #428, opened 2022-04-01
Update Dependencies to Latest Viper Release without fixing Injectivity

ArquintL:
To make use of the latest improvements in e.g. ViperServer, this PR updates Gobra to the latest Viper release. Note however that we still rely on assumed injectivity by the backends, which should be turned off eventually (see #427)


jcp19 (approved):
LGTM

PR #430, opened 2022-04-01
Update Dependencies to Latest Commits Silver submodules

ArquintL:
Updates Silver, Silicon, Carbon, and ViperServer to their latest commits on `master`. Furthermore, this PR adds support for the newly introduced submodules in Silicon and Carbon.

We should discuss in the next Viper / Gobra meeting whether ViperServer and Gobra should also switch to submodules.

Depends on #428


README.md
```diff
@@ -17,26 +17,23 @@ We call annotated Go programs Gobra programs and use the file extension `.gobra`
...
-    - Change directory to `gobraHome/carbon` and create the symbolic links:
-        - silver -> ..\silver
+3. Get submodules and add symbolic links
+    - Change directory to `gobraHome/silicon` and fetch & update submodules:
+        - `git submodule init; git submodule update`
+    - Change directory to `gobraHome/carbon` and fetch & update submodules:
+        - `git submodule init; git submodule update`
+   - To create a symbolic link from A to B, you have to run
+      - `mklink /D A B` (Windows (as admin)) resp.
+      - `ln -s B A` (Linux & macOS) (use forward instead of backward slashes in the following)
```
jcp19, line 32: instead of suggesting these steps, we could pass the `--recurse-submodules` flag to the `git clone` command

README.md
```diff
@@ -52,9 +49,9 @@ We call annotated Go programs Gobra programs and use the file extension `.gobra`
     4. Set the environment variable `BOOGIE_EXE` to the path of your Boogie executable.
 
 ### Compilation
-1. Change directory to `gobraHome/gobra-one`
+1. Change directory to `gobraHome/gobra`
```
jcp19, line 52: Nice catch :D

build.sbt
```diff
@@ -8,7 +8,9 @@ import scala.sys.process.Process
 import scala.util.Try
 
 // Import general settings from Silver, Silicon and Carbon
-lazy val silver = project in file("silver")
+
+// we assume that carbon/silver and silicon/silver point to the same version of the silver repo
+lazy val silver = project in file("silicon/silver")
```
jcp19, line 13: I'm fine with this assumption for now, while we do not come up with a better solution for viperserver. However, to avoid any surprises, I would probably add a check here similar to the one we do on the Dockerfile (we could even eval the result of the same git command using something like [this](https://alvinalexander.com/scala/how-to-execute-external-commands-stdout-scala/))

jcp19 (approved):
I have two minor suggestions, but otherwise looks fine

build.sbt
ArquintL: I chose not to include such a check in SBT (neither in Gobra nor ViperServer) because I think this might interfere with whatever local setup you're using. Maybe you want to still rely on sym links for development purposes or you know what you're doing or ...

jcp19: But would the project even compile if you try to use different versions of the dependencies?

jcp19: At any rate, feel free to merge it

PR #431, opened 2022-04-05
Add "only files with header" mode to Gobra

jcp19:
This PR adds the flag `--onlyFilesWithHeaders` to Gobra. The purpose of this flag is to instruct Gobra to only consider files that contain the annotation `// ##(gobra)`. This is useful for gradually adding Gobra annotations to the SCION repository, where the goal is to gradually introduce annotations to the code with a minimal impact on the non-annotated code. In particular, this PR allows the following goals:
- allow an annotated Go file to import a package whose files use features not supported by Gobra or that relies on unspecified packages.
- allow Gobra to target only a part of the files in a package without having to provide an explicit list of files to target.

There are currently some features of Gobra that pose additional challenges:
- Gobra rejects packages if there are files on the package path that have different package clauses. This is the case in SCION. A temporary fix (which works for my use case but might not be suitable in general) consists of not reading the package clause of files without the header.
- Every package imported in this mode must have at least one file "with header", otherwise Gobra complains that the package is empty.

### DIFF
--- a/src/main/scala/viper/gobra/Gobra.scala
+++ b/src/main/scala/viper/gobra/Gobra.scala
@@ -13,6 +13,7 @@ import viper.gobra.ast.frontend.PPackage
 import viper.gobra.ast.internal.Program
 import viper.gobra.ast.internal.transform.{CGEdgesTerminationTransform, OverflowChecksTransform}
 import viper.gobra.backend.BackendVerifier
+import viper.gobra.frontend.Source.TransformableSource
 import viper.gobra.frontend.info.{Info, TypeInfo}
 import viper.gobra.frontend.{Config, Desugar, PackageInfo, Parser, ScallopGobraConfig}
 import viper.gobra.reporting._
@@ -23,6 +24,12 @@ import viper.silicon.BuildInfo
 import viper.silver.{ast => vpr}
 
 import scala.concurrent.{Await, Future, TimeoutException}
+
+// TODO:
+// 1. fix issues with empty packages when running in --onlyFilesWithHeader mode
+// 2. fix issues with more than one package name in the same folder (?) => maybe not a problem if we only parse files with header
+// 3. test with verifiedSCION
+// 4. clean-up, add tests, commit and open a PR with a motivating text
 
 object GoVerifier {
 
@@ -133,6 +140,7 @@ class Gobra extends GoVerifier with GoIdeVerifier {
 
     val task = Future {
       val finalConfig = getAndMergeInFileConfig(config, pkgInfo)
+      // println(s"Files with header: ${finalConfig.filesWithHeader}")
       for {
         parsedPackage <- performParsing(pkgInfo, finalConfig)
         typeInfo <- performTypeChecking(parsedPackage, finalConfig)
@@ -192,12 +200,15 @@ class Gobra extends GoVerifier with GoIdeVerifier {
       } else {
         // our current "merge" strategy for potentially different, duplicate, or even contradicting configurations is to concatenate them:
         val args = configs.flatMap(configString => configString.split(" ")).toList
+        // TODO
+        val hasHeader = args.contains("gobra")
+        val actualArgs = args.filter(_ != "gobra")
         // skip include dir checks as the include should only be parsed and is not resolved yet based on the current directory
-        val inFileConfig = new ScallopGobraConfig(args, isInputOptional = true, skipIncludeDirChecks = true).config
+        val inFileConfig = new ScallopGobraConfig(actualArgs, isInputOptional = true, skipIncludeDirChecks = true).config
         // modify all relative includeDirs such that they are resolved relatively to the current file:
         val resolvedConfig = inFileConfig.copy(includeDirs = inFileConfig.includeDirs.map(
           // it's important to convert includeDir to a string first as `path` might be a ZipPath and `includeDir` might not
-          includeDir => Paths.get(input.name).getParent.resolve(includeDir.toString)))
+          includeDir => Paths.get(input.name).getParent.resolve(includeDir.toString)), filesWithHeader = if (hasHeader) Set(Paths.get(input.name)) else Set.empty)
         Some(resolvedConfig)
       }
     })
@@ -208,7 +219,8 @@ class Gobra extends GoVerifier with GoIdeVerifier {
 
   private def performParsing(pkgInfo: PackageInfo, config: Config): Either[Vector[VerifierError], PPackage] = {
     if (config.shouldParse) {
-      Parser.parse(config.packageInfoInputMap(pkgInfo), pkgInfo)(config)
+      // TODO move this logic to Config
+      Parser.parse(config.packageInfoInputMap(pkgInfo).filter(p => if(config.onlyFilesWithHeader) config.filesWithHeader.contains(p.toPath) else true), pkgInfo)(config) // TODO: update this, move somewhere else
     } else {
       Left(Vector())
     }
--- a/src/main/scala/viper/gobra/Gobra.scala
+++ b/src/main/scala/viper/gobra/Gobra.scala
@@ -25,9 +25,6 @@ import viper.silver.{ast => vpr}
 import scala.concurrent.{Await, Future, TimeoutException}
 
 // TODO:
-// 1. fix issues with empty packages when running in --onlyFilesWithHeader mode
-// 2. fix issues with more than one package name in the same folder (?) => maybe not a problem if we only parse files with header
-// 3. test with verifiedSCION
 // 4. clean-up, add tests, commit and open a PR with a motivating text
 
 object GoVerifier {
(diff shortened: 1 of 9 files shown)


ArquintL:
> * Gobra rejects packages if there are files on the package path that have different package clauses. This is the case in SCION. A temporary fix (which works for my use case but might not be suitable in general) consists of not reading the package clause of files without the header.

It still is a miracle to me how one can relax this restriction. 
The Go spec says the following:
> A set of files sharing the same PackageName form the implementation of a package. An implementation may require that all source files for a package inhabit the same directory.
That's fine as we can simply say that a directory can contain 1 or more packages.
The spec then continues on import declarations:
> If the PackageName is omitted, it defaults to the identifier specified in the [package clause](https://go.dev/ref/spec#Package_clause) of the imported package.
> The interpretation of the ImportPath is implementation-dependent but it is typically a substring of the full file name of the compiled package and may be relative to a repository of installed packages.
Here, I'm not sure how a (relative) file path can be resolved to the package clause if the package clause is not unique

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -192,8 +192,9 @@ class Gobra extends GoVerifier with GoIdeVerifier {
       } else {
         // our current "merge" strategy for potentially different, duplicate, or even contradicting configurations is to concatenate them:
         val args = configs.flatMap(configString => configString.split(" ")).toList
+        val actualArgs = args filter (_ != Config.headerTag)
```
ArquintL, line 195: I'm not super excited about this solution as I'm a bit afraid this will end up being a long list of some special cases in the future.

What about [something more Go like](https://www.digitalocean.com/community/tutorials/building-go-applications-for-different-operating-systems-and-architectures), e.g. `\\ +verify gobra`, and keep it completely separate from our infile configs?

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -115,6 +121,13 @@ case class Config(
     }
 }
 
+object Config {
+  // the header tag is assumed to be an invalid option to Gobra's cli
+  val headerTag = "gobra"
+  val header = """##\(%s\)""".format(headerTag).r
```
ArquintL, line 127: here we are assuming anyway that one cannot combine it with other infile configs (except when splitting them into separate comments)

src/main/scala/viper/gobra/frontend/PackageResolver.scala
```diff
@@ -194,7 +197,19 @@ object PackageResolver {
       case resource if !res.contains(resource) => resource.close()
       case _ =>
     })
-    res
+    res filter { p => !onlyFilesWithHeaders || isResourceWithHeader(p) }
```
ArquintL, line 200: Do you need the `onlyFilesWithHeaders` parameter in this entire file to avoid the problem of having different package clauses in the same folder? In contrast to filter the files only afterwards (i.e. outside of this object)?

jcp19: Yes. That was initially the reason. However, I now modified the code such that different package clauses can co-exist inside the same directory when **using testing packages for blackbox testing**. Even though the documentation that I came across is not very clear about that, that seems to be the only case where different package clauses are allowed in the same directory.

jcp19: Sure: I found [this](https://tip.golang.org/cmd/go/#hdr-Test_packages) while navigating on [this stack overflow thread](https://stackoverflow.com/questions/20427890/does-it-make-sense-to-have-two-packages-in-the-same-directory)

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -115,6 +121,14 @@ case class Config(
     }
 }
 
+object Config {
+  // the header signals that a file should be considered when running on "header-only" mode
+  val header = """\/\/\s*\+gobra""".r
+  val prettyPrintedHeader = "// +gobra"
+  require(header.matches(prettyPrintedHeader))
```
ArquintL, line 128: nice!

ArquintL (approved):
I have one comment on our treatment of files ending in `_test.go`. Otherwise, looks good to me

src/main/scala/viper/gobra/frontend/PackageResolver.scala
```diff
@@ -31,6 +31,8 @@ object PackageResolver {
   val stubDirectories = Vector("stubs")
   val fileUriScheme = "file"
   val jarUriScheme = "jar"
+  // packages ending with "_test" are considered test packages
+  private val testPackageRegex = """(.+)_test$""".r
```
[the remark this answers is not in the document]
jcp19: The documentation is very weak here. But in practice, the go compiler enforces that you can only compile together files belonging to package X and X_test. I checked this with the following minimal example:

file f.go:
```go
package p
```

file f_test.go:
```go
package p_test
```

Compiling these two files together with the command `go build f.go g_test.go` compiles the package. If instead you rewrite file f_test.go to the following, you get an error:
```go
package p_bla
```

Error:
```
found packages p (f.go) and p_bla (f_test.go) in /Users/joao/Desktop/test/test/test_test
```

This does not rule out necessarily that test files can have arbitrary names, but it does at least indicate that packages ending with `_test` have some special treatment. Unfortunately, I did not find unambiguous official sources that claim this.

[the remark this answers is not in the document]
jcp19: > When we are already at it, we could also add a special case for a directory named testdata that should completely be removed (search for "testdata" [here](https://pkg.go.dev/cmd/go@master#hdr-Test_packages)). In particular, this would help me as I'm currently facing a Gobra error because of not ignoring files in such directories

What kinds of errors are you having? Wouldn't gobra ignore these folders, as they usually do not have any .go and .gobra files?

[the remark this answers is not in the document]
ArquintL: > What kinds of errors are you having? Wouldn't gobra ignore these folders, as they usually do not have any .go and .gobra files?

I get the following error in package `golang.org/x/tools`:
`Missing package clause in ./.modules/golang.org/x/tools/cmd/fiximports/testdata/src/old.com/bad/bad.go`
I think it's a folder of invalid go files (for testing purposes) and thus have a `.go` file ending ^^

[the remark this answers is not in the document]
ArquintL: > This does not rule out necessarily that test files can have arbitrary names, but it does at least indicate that packages ending with `_test` have some special treatment. Unfortunately, I did not find unambiguous official sources that claim this.

Thanks for this additional bit of information. I would thus rephrase my observation that Go treats `testdata` folder and files ending in `_test.go` differently and additionally enforces that packages in files ending in `_test.go` cannot have arbitrary package clauses by must either be the same one as in the other files in this folder or ending in `_test`.

I'm not sure though to what extend we want to handle this in Gobra.  It seems however more desirable to watch out for `_test.go` files and completely ignore them (we would sacrifice the error if you would use an invalid package clause in there but technically that does only matter for compilation but not really at runtime).
Only checking the package clause for a `_test` ending seems less desirable to me as we could end up treating a file as being part of the codebase even though it's a unit test used for whitebox testing (and thus has the same package clause as the actual source files)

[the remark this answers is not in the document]
ArquintL: Haha but unfortunately not because it's throwing a violation before filtering the packages (as it could not identify the package clause as there is none)

src/main/scala/viper/gobra/frontend/PackageResolver.scala
```diff
@@ -180,16 +183,39 @@ object PackageResolver {
...
+    * Determines if a file is to be imported when the package implemented in the same directory is imported.
+    * All test files (ending in "_test.go") are skipped, as well as all files whose path contains the `testdata` directory.
+    * One difference between our implementation in Gobra and what the Go compiler does is that Gobra does not complain
+    * if the package clause in a test file mentions a package other than the one implemented in the same directory or that package
+    * suffixed with "_test".
+    * TODO: add this check
+    * More information about this can be found in https://tip.golang.org/cmd/go/#hdr-Test_packages.
+    */
+  private def shouldIgnoreResource(r: InputResource): Boolean = {
+    // Here, we could chose either a full path
```
ArquintL, line 195:
```suggestion
    // Here, we could choose the full path instead
```

src/main/scala/viper/gobra/frontend/PackageResolver.scala
```diff
@@ -180,16 +183,39 @@ object PackageResolver {
...
+    val path = r.path.toString
+    println(s"path: $path")
```
ArquintL, line 197: seems you've forgotten to remove this

ArquintL (approved):
Only a minor comment is left now

src/main/scala/viper/gobra/frontend/PackageResolver.scala
jcp19: oops, this was actually meant to be deleted, nice catch!

PR #432, opened 2022-04-12
Add break and continue support

Dspil:
Break is implemented as a goto directly outside the loop body that we want to break (depending on the label).

Continue is implemented as a goto directly before the loop body ends. In the case the loop has stuff to do before the next iteration (e.g. incrementing a variable) the goto label is before that.


src/main/scala/viper/gobra/ast/internal/utility/GobraStrategy.scala
```diff
@@ -40,7 +40,7 @@ object GobraStrategy {
       case (_: Seqn, Seq(stmts: Vector[Stmt@unchecked])) => Seqn(stmts)(meta)
       case (_: Label, Seq(label: LabelProxy)) => Label(label)(meta)
       case (_: If, Seq(cond: Expr, thn: Stmt, els: Stmt)) => If(cond, thn, els)(meta)
-      case (_: While, Seq(cond: Expr, invs: Vector[Assertion@unchecked], terminationMeasure: Option[TerminationMeasure@unchecked], body: Stmt)) => While(cond, invs, terminationMeasure, body)(meta)
+      case (_: While, Seq(cond: Expr, invs: Vector[Assertion@unchecked], terminationMeasure: Option[TerminationMeasure@unchecked], body: Stmt, label: Option[String])) => While(cond, invs, terminationMeasure, body, label)(meta)
```
jcp19, line 43: If `label` is kept as a filed, its type should be `Option[String@unchecked]` to avoid compiler warning

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -267,9 +267,13 @@ case class Seqn(stmts: Vector[Stmt])(val info: Source.Parser.Info) extends Stmt
 
 case class Label(id: LabelProxy)(val info: Source.Parser.Info) extends Stmt
 
+case class Continue(label: Option[String], escLabel: String, invs: Vector[Assertion])(val info: Source.Parser.Info) extends Stmt
+
```
jcp19, line 271: Would there be a way for us to not have to include the invariants here? Would it be impractical to get them from the ast somehow?

src/main/scala/viper/gobra/frontend/info/ExternalTypeInfo.scala
```diff
@@ -92,4 +92,8 @@ trait ExternalTypeInfo {
 
   /** returns the code root for a given node; can only be called on nodes that are enclosed in a code root */
   def codeRoot(n: PNode): PCodeRoot with PScope
+
+  def enclosedInLoop(n: PNode) : Option[PForStmt]
+
+  def enclosedInLabeledLoop(label: PLabelUse, n: PNode) : (Option[PForStmt], Vector[PExpression])
```
jcp19, line 98: Maybe a better name would be `enclosingLabeledLoop`. What is the second elemnt of the returned tuple here? I would add comments to both methods introduced in this file

src/main/scala/viper/gobra/translator/util/ViperUtil.scala
```diff
@@ -30,6 +30,10 @@ object ViperUtil {
     it.foldLeft[Exp](TrueLit()(pos, info, errT)){case (l, r) => And(l, r)(pos, info, errT)}
   }
 
+  def bigOr(it: Iterable[Exp])(pos: Position, info: Info, errT: ErrorTrafo): Exp = {
+    it.foldLeft[Exp](FalseLit()(pos, info, errT)){case (l, r) => Or(l, r)(pos, info, errT)}
+  }
+
```
jcp19, line 36: Is this method actually used somewhere?

src/main/scala/viper/gobra/frontend/info/implementation/resolution/Enclosing.scala
```diff
@@ -30,6 +30,38 @@ trait Enclosing { this: TypeInfoImpl =>
...
+    down[Option[PForStmt]](None){ case x: PForStmt => Some(x) }
+  }
+
+  def enclosingLabeledLoop(label: PLabelUse, node: PNode) : (Option[PForStmt], Vector[PExpression]) =
+    enclosingLoop(node) match {
+      case None => (None, Vector.empty)
+      case Some(l) => {
+        val p = tree.parent(l).head
+        if (p.isInstanceOf[PLabeledStmt]) {
+          val loopLabel = p.asInstanceOf[PLabeledStmt].label.name
```
jcp19, line 43: I would avoid doing `isInstanceOf[...]` followed by `asInstanceOf[...]` and just do pattern matching when that is practical

src/main/scala/viper/gobra/frontend/info/implementation/resolution/Enclosing.scala
```diff
@@ -30,6 +30,38 @@ trait Enclosing { this: TypeInfoImpl =>
     case _ => id
   })
 
+  lazy val enclosingLoop: PNode => Option[PForStmt] = {
+    down[Option[PForStmt]](None){ case x: PForStmt => Some(x) }
+  }
+
+  def enclosingLabeledLoop(label: PLabelUse, node: PNode) : (Option[PForStmt], Vector[PExpression]) =
+    enclosingLoop(node) match {
```
jcp19, line 38: I think this code can be simplified to something *similar* to
```scala
  def enclosingLabeledLoop(label: PLabelUse, node: PNode) : (Option[PForStmt], Vector[PExpression]) = {
    enclosingLoop(node) match {
      case None => (None, Vector())
      case Some(encLoop) => encLoop match {
        case tree.parent(l: PLabeledStmt) if l.label.name == label.name => (Some(encLoop), encLoop.spec.invariants)
        case _ => enclosingLabeledLoop(label, encLoop)
      }
    }
  }
```

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -812,13 +812,40 @@ object Desugar {
...
 
-          case l: PLabeledStmt =>
+          case l: PLabeledStmt => {
             val proxy = labelProxy(l.label)
-            for {
-              _ <- declare(proxy)
-              _ <- write(in.Label(proxy)(src))
-              s <- goS(l.stmt)
-            } yield s
+            if (l.stmt.isInstanceOf[PForStmt]) {
```
jcp19, line 817: We can use match instead of using the pattern `isInstanceOf` followed by `asInstanceOf`

jcp19 (changes requested):
So far, I only took a look at the low hanging fruit. I will try to review the encoding when I find the time

src/main/scala/viper/gobra/frontend/info/ExternalTypeInfo.scala
```diff
@@ -92,4 +92,8 @@ trait ExternalTypeInfo {
 
   /** returns the code root for a given node; can only be called on nodes that are enclosed in a code root */
   def codeRoot(n: PNode): PCodeRoot with PScope
+
+  def enclosedInLoop(n: PNode) : Option[PForStmt]
```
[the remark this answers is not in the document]
Dspil: That is the name of the function defined in Enclosing.scala. Can I use exactly the same name?

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -267,9 +267,13 @@ case class Seqn(stmts: Vector[Stmt])(val info: Source.Parser.Info) extends Stmt
...
+case class Break(label: Option[String], escLabel: String)(val info: Source.Parser.Info) extends Stmt
+
 case class If(cond: Expr, thn: Stmt, els: Stmt)(val info: Source.Parser.Info) extends Stmt
 
-case class While(cond: Expr, invs: Vector[Assertion], terminationMeasure: Option[TerminationMeasure], body: Stmt)(val info: Source.Parser.Info) extends Stmt
+case class While(cond: Expr, invs: Vector[Assertion], terminationMeasure: Option[TerminationMeasure], body: Stmt, label: Option[String])(val info: Source.Parser.Info) extends Stmt
```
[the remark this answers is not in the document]
Dspil: Yes, I could not think of a way to find a parent so I just stored it beforehand.

src/main/scala/viper/gobra/frontend/info/ExternalTypeInfo.scala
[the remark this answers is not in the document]
Dspil: I have renamed them to enclosingLoopNode and enclosingLabeledLoopNode respectively to address the alias issue

src/test/resources/regressions/features/loops/loops-continue-label4.gobra
```diff
@@ -0,0 +1,20 @@
+package pkg
```
jcp19, line 1: This file is missing the header, causing the test to fail

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -3145,6 +3209,15 @@ object Desugar {
       */
     private var scopeMap: Map[PScope, Int] = Map.empty
 
+    /**
+      * 'continueCounter' should continue a unique identifier for each for statement,
```
jcp19, line 3213: "contain" instead of "continue"?

jcp19: Also, I would add more documentation explaining this stack-based algorithm.

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -3200,6 +3273,27 @@ object Desugar {
...
 
+    def pushFor(node: PNode, info: TypeInfo): String = {
+      val codeRoot = info.codeRoot(node)
+      val (stack, max) = continueCounter.getOrElse(codeRoot, (Seq(), 0))
+      continueCounter += (codeRoot -> (Seq(max + 1) ++ stack, max + 1))
+      CONTINUE_LABEL_PREFIX + (max + 1)
+    }
+
+    /** returns a unique identifier for a for stmt*/
+    def fetchForId(node: PNode, up: Int, info: TypeInfo): String = {
```
jcp19, line 3284: why do we use the id here? Wouldn't it be enough ti just pass the node and then find it in the stack? Is it some kind of optimization?

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1062,6 +1099,32 @@ object Desugar {
...
+          case n@PContinue(label) =>
+            label match {
+              case None => unit(in.Continue(nm.fetchForId(n, 0, info))(src))
+              case Some(l) => unit(in.Continue(nm.fetchForId(n, info.enclosingLabeledLoopOrder(l, n), info))(src))
+            }
+
+          case n@PBreak(label) =>
+            label match {
+              case None =>
+                info.enclosingLoopNode(n) match{
```
jcp19, line 1111: space missing

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1062,6 +1099,32 @@ object Desugar {
...
+                  case None => violation("break should be enclosed in a loop")
+                  case Some(loop) =>
+                    for {
+                      (dInvPre, dInv) <- prelude(sequence(loop.spec.invariants map assertionD(ctx)))
```
jcp19, line 1115: What is the purpose of this line here? In particular, `dInvPre` and `dInv` don't appear to be used

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -812,13 +812,45 @@ object Desugar {
...
+                    wh = in.Seqn(
+                      Vector(dPre) ++ dCondPre ++ dInvPre ++ dTerPre ++ Vector(
+                        in.While(dCond, dInv, dTer, in.Seqn(
+                          Vector(dBody, loopLabel, dPost) ++ dCondPre ++ dInvPre ++ dTerPre
+                        )(src), Some(l.label.name))(src)
+                      )
+                    )(src)
+                  } yield wh
+                ))
+              }
```
jcp19, line 844: Isn't this pretty much the same code as the code for desugaring a loop. Why do we need to special case this?

src/main/scala/viper/gobra/translator/implementations/translator/StatementsImpl.scala
```diff
@@ -85,30 +102,45 @@ class StatementsImpl extends Statements {
       case in.Label(id) =>
         unit(vpr.Label(id.name, Seq.empty)(pos, info, errT))
 
+      case in.Continue(escLabel) =>
+        unit(vpr.Goto(escLabel)(pos, info, errT))
+
+      case in.Break(_, escLabel) =>
```
jcp19, line 108: what is the purpose of having an `escLabel` and the other label?

jcp19 (commented):
Looks much better. This is more fine-grained feedback. Besides, there are some points I am curious about:
1. I would add a test with unlabeled continue / break in different functions - This makes sure that it is not problematic to repeat labels at the viper code.
2. Verifying one of the tests with the `--checkConsistency` flag leads to consistency errors of the kind "No matching identifier CL_2 found of type Label". Given that this flag might become enabled by default, it might be good to understand what is the problem here. 
3. The current solution does not seem to reject repeated labels in the Go source. Does the Go compiler reject them? If so, would it be hard to replicate this here?

src/main/scala/viper/gobra/ast/internal/PrettyPrinter.scala
```diff
@@ -304,6 +304,13 @@ class DefaultPrettyPrinter extends PrettyPrinter with kiama.output.PrettyPrinter
       showVar(resTarget) <> "," <+> showVar(successTarget) <+> "=" <+> showExpr(mapLookup)
     case PredExprFold(base, args, p) => "fold" <+> "acc" <> parens(showExpr(base) <> parens(showExprList(args)) <> "," <+> showExpr(p))
     case PredExprUnfold(base, args, p) => "unfold" <+> "acc" <> parens(showExpr(base) <> parens(showExprList(args)) <> "," <+> showExpr(p))
+
+    case Continue(_) => "continue"
```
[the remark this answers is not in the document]
Dspil: The break node has both the label of the loop it wants to break and the label that the goto statement will use. Continue node has only the latter, so printing it wouldn't show any information. Would it be better to pass the loop label just for debugging and prettyPrinting reasons?

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -267,9 +267,13 @@ case class Seqn(stmts: Vector[Stmt])(val info: Source.Parser.Info) extends Stmt
...
 case class Label(id: LabelProxy)(val info: Source.Parser.Info) extends Stmt
 
+case class Continue(escLabel: String)(val info: Source.Parser.Info) extends Stmt
+
+case class Break(label: Option[String], escLabel: String)(val info: Source.Parser.Info) extends Stmt
+
 case class If(cond: Expr, thn: Stmt, els: Stmt)(val info: Source.Parser.Info) extends Stmt
 
-case class While(cond: Expr, invs: Vector[Assertion], terminationMeasure: Option[TerminationMeasure], body: Stmt)(val info: Source.Parser.Info) extends Stmt
+case class While(cond: Expr, invs: Vector[Assertion], terminationMeasure: Option[TerminationMeasure], body: Stmt, label: Option[String])(val info: Source.Parser.Info) extends Stmt
```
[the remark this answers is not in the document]
Dspil: Because these are not currently existing labels, they will be created when the viper code is created and will have the name stored in that string.

Dspil:
Go does not reject multiple labels at the same point in the code.

jcp19:
> Go does not reject multiple labels at the same point in the code.

Sorry, I was probably not clear. I was wondering whether you can have multiple times the **same** label in Go. I would guess not

Dspil:
You are right, Go doesn't allow that. But neither does Gobra, I didn't implement this behaviour but it's there.

src/main/scala/viper/gobra/frontend/Desugar.scala
Dspil: You are right, we can use the info object to produce the 'up' argument inside instead of computing it outside.

src/main/scala/viper/gobra/frontend/Desugar.scala
Dspil: Maybe I could make the repetitive code a function and call it

src/main/scala/viper/gobra/translator/implementations/translator/StatementsImpl.scala
Dspil: `esclabel` is the label that the break will `goto`. the other label is the loop label the break statement corresponds to.

src/test/resources/regressions/features/loops/loops-continue-label3.gobra
```diff
@@ -0,0 +1,14 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package pkg
+
+func foo() {
+l:
+	invariant true
+	for i := 0; i < 10; i++ {
+		for j := 0; j < 10; j++ {
```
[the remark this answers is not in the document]
Dspil: I added two of those

src/main/scala/viper/gobra/ast/internal/PrettyPrinter.scala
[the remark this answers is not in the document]
jcp19: Yes, I would include for both debugging and prettyPrinting. Also, I think it is better to be consistent here - if we print the labels for break when they are present, we should do the same for continue

src/main/scala/viper/gobra/ast/internal/Program.scala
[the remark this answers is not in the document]
Dspil: LabelProxy seems to have much more information than I need, I just need the name of that label and use nothing more. Is it a problem if we kept the String?

src/main/scala/viper/gobra/ast/internal/PrettyPrinter.scala
```diff
@@ -650,7 +653,7 @@ class ShortPrettyPrinter extends DefaultPrettyPrinter {
     case Seqn(_) => emptyDoc
     case Label(label) => showProxy(label)
     case If(cond, _, _) => "if" <> parens(showExpr(cond)) <+> "{...}" <+> "else" <+> "{...}"
-    case While(cond, invs, measure, _) => "while" <> parens(showExpr(cond)) <> line <>
+    case While(cond, invs, measure, _, label) => opt(label)(text) <> opt(label.map(_ => ": "))(text) <> "while" <> parens(showExpr(cond)) <> line <>
```
ArquintL, line 656: couldn't this be simplify as follows?
```suggestion
    case While(cond, invs, measure, _, label) => opt(label.map(l => s"${l}: "))(text) <> "while" <> parens(showExpr(cond)) <> line <>
```

src/main/scala/viper/gobra/ast/internal/PrettyPrinter.scala
```diff
@@ -250,7 +250,7 @@ class DefaultPrettyPrinter extends PrettyPrinter with kiama.output.PrettyPrinter
     case Seqn(stmts) => ssep(stmts map showStmt, line)
     case Label(label) => showProxy(label)
     case If(cond, thn, els) => "if" <> parens(showExpr(cond)) <+> block(showStmt(thn)) <+> "else" <+> block(showStmt(els))
-    case While(cond, invs, measure, body) => "while" <> parens(showExpr(cond)) <> line <>
+    case While(cond, invs, measure, body, label) => opt(label)(text) <> opt(label.map(_ => ": "))(text) <> "while" <> parens(showExpr(cond)) <> line <>
```
ArquintL, line 253: couldn't this be simplify as follows?
```suggestion
    case While(cond, invs, measure, body, label) => opt(label.map(l => s"${l}: "))(text) <> "while" <> parens(showExpr(cond)) <> line <>
```

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -267,9 +267,13 @@ case class Seqn(stmts: Vector[Stmt])(val info: Source.Parser.Info) extends Stmt
 
 case class Label(id: LabelProxy)(val info: Source.Parser.Info) extends Stmt
 
+case class Continue(label: Option[String], escLabel: String)(val info: Source.Parser.Info) extends Stmt
```
ArquintL, line 270: what does `escLabel` mean? I would either expand the parameter name to make it self explanatory or add a JavaDoc comment

src/main/scala/viper/gobra/frontend/info/implementation/resolution/Enclosing.scala
```diff
@@ -30,6 +30,37 @@ trait Enclosing { this: TypeInfoImpl =>
...
+  }
+
+  // Returns the enclosing loop that has a specific label
+  // It also returns the invariants of that loop
+  def enclosingLabeledLoop(label: PLabelUse, node: PNode) : Option[PForStmt] = {
+    enclosingLoop(node) match {
+      case None => None
+      case Some(encLoop) => encLoop match {
+        case tree.parent(l: PLabeledStmt) if l.label.name == label.name => Some(encLoop)
+        case _ => enclosingLabeledLoop(label, tree.parent(encLoop).head)
```
ArquintL, line 44: What does `head` do here?
Is this equivalent to this?
```suggestion
        case tree.parent(p) => enclosingLabeledLoop(label, p)
```

src/main/scala/viper/gobra/frontend/info/implementation/resolution/Enclosing.scala
```diff
@@ -30,6 +30,37 @@ trait Enclosing { this: TypeInfoImpl =>
...
+  // Used when 'continue l' statements that correspond to
+  // a for statement with label l are being desugared. This number will be
+  // used to fetch the unique label identifier for that for loop where
+  // the continue statement must goto.
+  def enclosingLoopOrder(label: PLabelUse, node: PNode, order: Int = 0) : Int = {
+    enclosingLoop(node) match {
+      case None => violation("Didn't find enclosing loop to continue statement with the same label.")
+      case Some(encLoop) => encLoop match {
+        case tree.parent(l: PLabeledStmt) if l.label.name == label.name => order
+        case _ => enclosingLoopOrder(label, tree.parent(encLoop).head, order + 1)
```
ArquintL, line 59: same here

src/test/resources/regressions/features/loops/loops-break1.gobra
```diff
@@ -0,0 +1,10 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package pkg
+
+func foo() {
+	for true {
+		break
+	}
```
ArquintL, line 9: I would add an `assert false` after the loop to show that this loop indeed terminates (and thus `break` behaves as expected)

src/test/resources/regressions/features/loops/loops-break2.gobra
```diff
@@ -0,0 +1,23 @@
...
+
+func foo() {
+	var a [15]int
+	invariant 0 <= i && i <= cap(a)
+	for i := 0; i < cap(a); i++ {
+		a[i] = i
+	}
+	var contains5 = false
+	invariant 0 <= i && i <= cap(a)
+	invariant contains5 == (exists j int :: 0 <= j && j < i && a[j] == 5)
```
ArquintL, line 14: I think it would be good habit to have proper triggers for all testcases (for demonstration purposes but also to avoid performance issues)
```suggestion
	invariant contains5 == (exists j int :: { a[j] } 0 <= j && j < i && a[j] == 5)
```

src/test/resources/regressions/features/loops/loops-break2.gobra
```diff
@@ -0,0 +1,23 @@
...
+	decreases cap(a) - i
+	for i := 0; i < cap(a); i++ {
+		if a[i] == 5 {
+			contains5 = true
+			break
+		}
+	}
+	assert contains5 == (exists i int :: 0 <= i && i < cap(a) && a[i] == 5)
```
ArquintL, line 22:
```suggestion
	assert contains5 == (exists i int :: { a[i] } 0 <= i && i < cap(a) && a[i] == 5)
```

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1062,6 +1101,14 @@ object Desugar {
               }
             } yield in.Seqn(Vector(dPre, exprAss, clauseBody))(src)
 
+          case n@PContinue(label) => unit(in.Continue(label.map(x => x.name), nm.fetchForId(n, info))(src))
+
+          case n@PBreak(label) =>
+            label match {
+              case None => unit(in.Break(None, nm.fresh(n, info))(src))
+              case Some(l) => unit(in.Break(Some(l.name), nm.fresh(n, info))(src))
+              }
```
ArquintL, line 1110:
```suggestion
            }
```

ArquintL: i.e. indentation is slightly off

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -3145,6 +3193,28 @@ object Desugar {
...
+      * 'continueCounter' should return a unique identifier for each for statement,
+      * in order to handle continue statements. These will be translated to gotos to
+      * labels with prefix 'CONTINUE_LABEL_PREFIX' and suffix the identifier of the
+      * loop that has to continue. For every code root, there can be the same identifiers
+      * for for loops since there can be identical labels in different functions. For
+      * each codeRoot, a tuple of a stack and a max value is stored. Every time a new
+      * for statement is desugared, it first pushes max + 1 to the stack and saves the new
+      * max. Every time there is a need for the identifier of a for loop for a continue statement
+      * there are two possibilities:
+      * 1) The continue statement is unlabeld so the identifier is at the top of the stack.
```
ArquintL, line 3206:
```suggestion
      * 1) The continue statement is unlabeled so the identifier is at the top of the stack.
```

src/main/scala/viper/gobra/translator/implementations/translator/StatementsImpl.scala
```diff
@@ -68,6 +68,23 @@ class StatementsImpl extends Statements {
       } yield exhale
     }
 
+    def gatherBreakLabels(node: in.Node, label: Option[String], depthControl : Boolean = true): Vector[String] =
```
ArquintL, line 71: I'm lacking some intuition what this function is gathering. It seems to collect all `escLabel` of break statements if they are either unlabeled or are labeled `label`. But what's the purpose of `depthControl`?

ArquintL (commented):
I have some more comments. Thanks a lot already for the many testcases!

src/main/scala/viper/gobra/translator/implementations/translator/StatementsImpl.scala
Dspil: You are right, this is not obvious, I added a javaDoc comment explaining it. Basically, since a break will become a goto to a unique escLabel, we need to gather all these escLabels corresponding to one while loop so we can insert them after the while loop. The escLabels that correspond to a specific loop are the ones that belong to unlabeled breaks directly inside the loop's body and the escLabels of labeled breaks with the same label as the while loop. The argument 'depthControl' is needed to determine if we are directly inside the loop body or we are currently traversing nested while loops so we should disregard unlabeled breaks.

src/main/scala/viper/gobra/frontend/info/implementation/resolution/Enclosing.scala
```diff
@@ -30,6 +30,39 @@ trait Enclosing { this: TypeInfoImpl =>
...
+      }
+    }
+  }
+
+  // Returns the number of loops between a node 'node' and a label 'label'
+  // Used when 'continue l' statements that correspond to
+  // a for statement with label l are being desugared. This number will be
+  // used to fetch the unique label identifier for that for loop where
+  // the continue statement must goto.
+  def enclosingLoopOrder(label: PLabelUse, node: PNode, order: Int = 0) : Int = {
```
jcp19, line 55: Is the purpose of `order` here to be able to increment its value in the recursive call to itself? Is that the only purpose of this param? If so, I usually prefer to introduce an auxiliary method that takes that extra param inside the definifition of the method to avoid cluttering the signature of the method. Signature of public methods should only have parameters with a clear meaning outside of the implementation. Here, this is less of a problem given that the signature that you expose in the interface `ExternalTypeInfo` don't have this parameter.

jcp19 (approved):
The code improved a lot since the original proposal! I noticed some small things that I think should be improved before merging. If @ArquintL and @Felalolf agree, you can merge the PR after addressing my comments.
PS: please squash your PR before merging, given that it already has a considerable number of commits

src/test/scala/viper/gobra/typing/StmtTypingUnitTests.scala
```diff
@@ -0,0 +1,70 @@
...
+      PPackageClause(PPkgDef("pkg")),
+      Vector(),
+      Vector(PMethodDecl(
+        PIdnDef("foo"),
+        PUnnamedReceiver(PMethodReceiveName(PNamedOperand(PIdnUse("self")))),
+        inArgs.map(_._1),
+        PResult(Vector()),
+        PFunctionSpec(Vector(), Vector(), Vector(), Vector(), isPure = false),
+        Some(PBodyParameterInfo(inArgs.collect{ case (n: PNamedParameter, true) => PIdnUse(n.id.name) }), PBlock(Vector(body)))
+      ))
```
Felalolf, line 49: Please add the program as a comment so that it is easier to read, e.g. 
```
package pkg

func (whatever the receiver is) foo(<arguments>) {
  <body>
}
```

It seems to me that the receiver is incorrect. You should have positives tests (where you assert that typechecking is successful) as well, to make sure that the tests do not fail due to something else, e.g. an incorrect receiver.

src/test/resources/regressions/features/loops/loops-continue-label1.gobra
```diff
@@ -0,0 +1,14 @@
...
+package pkg
+
+func foo() {
+l:
+	invariant true
+	for true {
+		for true {
+			continue l
+		}
+	}
```
Felalolf, line 13: please, add some assertions. It can be the easiest assertion that you can think of that is not true.

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -267,9 +267,24 @@ case class Seqn(stmts: Vector[Stmt])(val info: Source.Parser.Info) extends Stmt
...
+  * been placed properly while desugaring for loops.
+  */
+case class Continue(label: Option[String], escLabel: String)(val info: Source.Parser.Info) extends Stmt
+
+/**
+  * 'label' corresponds to the loop label we want to break out of. In case it is a normal
+  * break node it is 'None'.
+  * The break node will be replaced by a goto statement to 'escLabel'.
+  */
+case class Break(label: Option[String], escLabel: String)(val info: Source.Parser.Info) extends Stmt
```
Felalolf, line 283: The current approach looks a bit too complicated to me. I would just transform a loop label into three labels:

```
L:
<loop>
```
to
```
L:
L$Continue:
<loop>
L$Break
```

For loops that do not have a label you just take a fresh identifier `F` and create the `F$Continue` and `F$Break` label. Similarly, for unlabeled break and continues, you just add the corresponding fresh identifier and treat them as if they had a label.

And then tranform 
`goto L` to `goto L`
`break L` to `goto L$Break`
`continue L` to `goto L$Continue`

You can do that either in the desugaring (then you only need a goto node in the internal AST and loops also do not have a label) or in the encoding (then the internal AST has goto, break, continue, but they have exactly one label, and loops have a label). I prefer the solution within the desugarer. Also you should always use `LabelProxy` instead of `String`. Laslty, it would be interesting if too many labels cause performance problems. But we could address this with a Viper to Viper transformation, which eliminates unused labels or maybe even merges labels together if possible.

Felalolf (changes requested):
You have a lot of tests without any assertions. It would be nice if you could add some assertions. They can be the easiest assertions you can come up with, I gave two examples. I wrote my biggest issue in one comment, about the approach.

src/main/scala/viper/gobra/ast/internal/Program.scala
Felalolf: I just noticed that my continue label is wrong, but you get the idea (I guess you put it at the end of the loop body)

Felalolf (approved):

PR #437, opened 2022-04-26
Reduce the size of names in the encoding

jcp19:
Right now, the identifiers obtained in the encoding are so big that reading the generated viper file for debugging is very hard. This PR tackles that problem in two fronts:
1. remove the viper id from the identifiers of local variables - @ArquintL maybe I'm missing something, but I do not see the purpose in including this in every local variable
2. (less impactful) use base64 instead of hexadecimal to print the hashsum of the package identifier when needed, which should lead to smaller names

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Desugar.scala
+++ b/src/main/scala/viper/gobra/frontend/Desugar.scala
@@ -3136,6 +3136,13 @@ object Desugar {
       * are consistently named across verification runs (i.e. independent of modifications to other function)
       */
     private var scopeCounter: Map[PCodeRoot, Int] = Map.empty
+
+    /**
+      * 'scopeMap' should return a unique identifier only within a package,
+      * and thus may not be unique in the resulting Viper encoding.
+      * This should be fine as long as the resulting name is only used in a non-global scope.
+      * Currently, it is used for variables, in and out parameters, and receivers.
+      */
     private var scopeMap: Map[PScope, Int] = Map.empty
 
     private def maybeRegister(s: PScope, ctx: ExternalTypeInfo): Unit = {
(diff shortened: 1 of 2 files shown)


src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -3154,24 +3154,24 @@ object Desugar {
       maybeRegister(s, context)
 
       // n has occur first in order that function inverse properly works
-      s"${n}_${context.pkgInfo.viperId}_$postfix${scopeMap(s)}" // deterministic
+      s"${n}_$postfix${scopeMap(s)}" // deterministic
```
ArquintL, line 3157: As far as I can tell, the `scopeMap` returns a unique identifier only within a package. I.e. it is not unique in the resulting Viper encoding.
This should however be fine as long as the resulting name is only used in a non-global scope. It's currently used for variables, in and out parameters, and receivers, which should be fine. Nevertheless, I would add a short documentation to this function

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -3154,24 +3154,24 @@ object Desugar {
...
     }
 
-    private def nameWithoutScope(postfix: String)(n: String, context: ExternalTypeInfo): String = {
+    private def topLevelName(postfix: String)(n: String, context: ExternalTypeInfo): String = {
```
ArquintL, line 3160: Much better, thanks!

ArquintL (approved):

src/main/scala/viper/gobra/frontend/Desugar.scala
jcp19: done :)

PR #439, opened 2022-04-26
Simplify 'copy' function to avoid matching loops

jcp19:
Our current encoding of `copy` handles all cases supported in Go, including the case when the src and dst are overlapping. The spec is more complicated to deal with this particular (rarely used) case. As it stands, it is also triggering often matching loops. To avoid that problem for now, this PR disables those. This is a temporary solution, until we design a more careful spec for all cases

### DIFF
--- a/src/main/scala/viper/gobra/translator/implementations/translator/BuiltInMembersImpl.scala
+++ b/src/main/scala/viper/gobra/translator/implementations/translator/BuiltInMembersImpl.scala
@@ -571,7 +571,7 @@ class BuiltInMembersImpl extends BuiltInMembers {
             in.ExprAssertion(
               in.EqCmp(
                 in.IndexedExp(dstParam, i, dstUnderlyingType)(src),
-                in.IndexedExp(srcParam, i, srcUnderlyingType)(src)
+                in.Old(in.IndexedExp(srcParam, i, srcUnderlyingType)(src), srcUnderlyingType.elems)(src)
               )(src)
             )(src)
           }
--- a/src/main/scala/viper/gobra/translator/implementations/translator/BuiltInMembersImpl.scala
+++ b/src/main/scala/viper/gobra/translator/implementations/translator/BuiltInMembersImpl.scala
@@ -366,18 +366,6 @@ class BuiltInMembersImpl extends BuiltInMembers {
       in.SepForall(Vector(i), trigger(i), in.Implication(range(i), body(i))(src))(src)
     }
 
-    def quantifyPure(trigger: in.BoundVar => Vector[in.Trigger], range: in.BoundVar => in.Expr, body: in.BoundVar => in.Expr): in.Expr = {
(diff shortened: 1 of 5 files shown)


src/main/scala/viper/gobra/translator/implementations/translator/BuiltInMembersImpl.scala
```diff
@@ -505,20 +493,23 @@ class BuiltInMembersImpl extends BuiltInMembers {
...
           * requires forall i int :: { src[i] } (0 <= i && i < len(src)) ==> acc(&src[i], p)
-          * requires forall i int :: { dst[i] } (0 <= i && i < len(dst) && (forall j int :: 0 <= j && j < len(src) ==> &dst[i] != &src[j])) ==> acc(&dst[i], p)
           * ensures len(dst) <= len(src) ==> res == len(dst)
           * ensures len(src) < len(dst) ==> res == len(src)
-          * ensures forall i int :: { dst[i] } 0 <= i && i < len(dst) ==> acc(&dst[i], 1-p)
+          * ensures forall i int :: { dst[i] } 0 <= i && i < len(dst) ==> acc(&dst[i], write)
           * ensures forall i int :: { src[i] } 0 <= i && i < len(src) ==> acc(&src[i], p)
-          * ensures forall i int :: { dst[i] } (0 <= i && i < len(dst) && (forall j int :: 0 <= j && j < len(src) ==> &dst[i] != &src[j])) ==> acc(&dst[i], p)
-          * ensures forall i int :: { dst[i] } (0 <= i && i < len(src) && i < len(dst)) ==> dst[i] == old(src[i])
+          * ensures forall i int :: { dst[i] } (0 <= i && i < len(src) && i < len(dst)) ==> dst[i] == src[i]
```
ArquintL, line 503: Is there a particular reason you dropped the ‚old‘? You loose all knowledge about the values of ‚src‘ if you pass all permissions to it to ‚copy‘

src/test/resources/regressions/features/builtin/copy-fail2.gobra
```diff
@@ -49,6 +49,8 @@ func test3() {
 	assert s[4] == 4
 	assert s[5] == 5
 
+  /*
```
ArquintL, line 52: You could add our ignore annotation together with the issue‘s number

ArquintL (approved):

src/test/resources/regressions/features/builtin/copy-fail3.gobra
```diff
@@ -0,0 +1,35 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+//:: IgnoreFile(/gobra/issue/439/)
```
ArquintL, line 4: I would create an actual issue for that and then use that number instead of referencing the PR as an issue

PR #442, opened 2022-05-02
Add options to assume Injectivity On Inhale and to disable it

jcp19:
This PR would optimally be combined with PR #409, but we can merge it already. When #409 is merged, we should adapt the default value of the option and its description. (This option should exist though even after we merge #409, as it might be necessary to verify already existing codebases like SCION)

PS: We can merge this after merging #434 to avoid more changes in that PR

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Config.scala
+++ b/src/main/scala/viper/gobra/frontend/Config.scala
@@ -549,9 +549,9 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
 
   val assumeInjectivityOnInhale: ScallopOption[Boolean] = toggle(
     name = "assumeInjectivityOnInhale",
-    descrYes = "Assumes injectivity of predicates on inhale, like in Viper versions previous to 2022.02 (default)",
+    descrYes = "Assumes injectivity of the receiver expression when inhaling quantified permissions, instead of checking it, like in Viper versions previous to 2022.02 (default)",
     descrNo = "Does not assume injectivity on inhales (this will become the default in future versions)",
-    default = Some(true),
+    default = Some(ConfigDefaults.DefaultAssumeInjectivityOnInhale),
     noshort = true
   )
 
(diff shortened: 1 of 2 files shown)


src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -540,6 +547,13 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
     noshort = true
   )
 
+  val assumeInjectivityOnInhale: ScallopOption[Boolean] = toggle(
+    name = "assumeInjectivityOnInhale",
+    descrYes = "Assumes injectivity of predicates on inhale, like in Viper versions previous to 2022.02 (default)",
+    descrNo = "Does not assume injectivity on inhales (this will become the default in future versions)",
+    default = Some(true),
```
ArquintL, line 554:
```suggestion
    default = Some(ConfigDefaults.DefaultAssumeInjectivityOnInhale),
```

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -540,6 +547,13 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
     noshort = true
   )
 
+  val assumeInjectivityOnInhale: ScallopOption[Boolean] = toggle(
+    name = "assumeInjectivityOnInhale",
+    descrYes = "Assumes injectivity of predicates on inhale, like in Viper versions previous to 2022.02 (default)",
```
ArquintL, line 552:
```suggestion
    descrYes = "Assumes injectivity of the receiver expression when inhaling quantified permissions, instead of checking it, like in Viper versions previous to 2022.02 (default)",
```

ArquintL: This is the description of Viper's config option

ArquintL (approved):
only two small comments

PR #443, opened 2022-05-02
fix bug where int8 would be parsed as named operand

Dspil:
This PR fixes a bug where int8 would be parsed as a named operand and not as a type. It was just a missing case in a match construct.


jcp19 (approved):
Awesome, thanks!

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
```diff
@@ -235,6 +235,7 @@ class ParseTreeTranslator(pom: PositionManager, source: Source, specOnly : Boole
     typ.name match {
       case "perm" => PPermissionType().at(typ)
       case "int" => PIntType().at(typ)
+      case "int8" => PIntType().at(typ)
```
jcp19, line 238: shouldn't this be `PInt8Type`?

jcp19 (changes requested):
Oops I overlooked this question

jcp19 (approved):

PR #445, opened 2022-05-06
strengthen specification for ParseUint

Dspil:
Strengthens the specification of ParseUint

### DIFF
--- a/src/main/resources/stubs/strconv/atoi.gobra
+++ b/src/main/resources/stubs/strconv/atoi.gobra
@@ -40,10 +40,20 @@ pure func (e *NumError) Unwrap() error { return e.Err }
 // IntSize is the size in bits of an int or uint value.
 // const IntSize = intSize
 
+// a to the power of b
+ghost
+requires exp >= 0
+ensures res == (exp == 0 ? 1 : (base * Exp(base, exp - 1)))
+pure func Exp(base int, exp int) (res int) {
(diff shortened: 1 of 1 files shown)


src/main/resources/stubs/strconv/atoi.gobra
```diff
@@ -40,10 +40,20 @@ pure func (e *NumError) Unwrap() error { return e.Err }
...
+ghost
+requires exp >= 0
+ensures res == (exp == 0 ? 1 : (base * Exp(base, exp - 1)))
+pure func Exp(base int, exp int) (res int) {
+	return exp == 0 ? 1 : (base * Exp(base, exp - 1))
+}
+
 // ParseUint is like ParseInt but for unsigned numbers.
 requires base == 0 || (2 <= base && base <= 36)
+requires bitSize > 0
```
jcp19, line 53: the doc (https://pkg.go.dev/strconv#ParseInt) also mentions that "The bitSize argument specifies the integer type that the result must fit into. Bit sizes 0, 8, 16, 32, and 64 correspond to int, int8, int16, int32, and int64. If bitSize is below 0 or above 64, an error is returned.". Value '0' is a special case here and I would ignore it, but I would the condition `bitSize <= 64`  to the precondition

jcp19 (approved):
We can merge this after adding the precondition I mentioned and after the tests go through

PR #447, opened 2022-05-10
Fix interface issues 10 05 22

### LINKED ISSUE #406: Naming of Domains & Interfaces

### LINKED ISSUE #408: Interface methods calls do not check that the receiver is the `nil` of the interface type
The following code is accepted by Gobra, but compiling it and running it leads to a runtime execution because the var `x` in `main` is `nil` and its dynamic type does not allow to dispatch this call to one of the implementation methods.
```go
type T interface{
    f() int
}

type TImpl struct{}

func (x* TImpl) f() int {
    return 0
}

func main() {
    var x T
    x.f()
}
```

Felalolf:
Fixes #406 
Fixes #408

### DIFF
--- a/src/test/resources/regressions/features/interfaces/proofGenForImport.gobra
+++ b/src/test/resources/regressions/features/interfaces/proofGenForImport.gobra
@@ -11,6 +11,9 @@ package client
 import pkg "proofGenForImportPkg"
 
 func client() {
-    impl := &pkg.ImplPkgState{}
+    impl := getImpl()
     impl.LibraryFn()
 }
+
+ensures acc(res) && res.Itf != nil
+func getImpl() (res *pkg.ImplPkgState)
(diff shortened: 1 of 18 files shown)


Dspil (approved):

src/test/resources/regressions/features/interfaces/proofGenForImport.gobra
```diff
@@ -11,6 +11,9 @@ package client
 import pkg "proofGenForImportPkg"
 
 func client() {
-    impl := &pkg.ImplPkgState{}
```
ArquintL, line 14: is it not guaranteed that `&pkg.ImplPkgState{}` is non-nil?

ArquintL (approved):

src/test/resources/regressions/features/interfaces/proofGenForImport.gobra
Felalolf: This is probably hard to see in the diff: ImplPkgState is a struct that contains the interface, so `ImplPkgState{}` initializes the interface fields with the default value, meaning that the interface field is initialized with nil.

PR #448, opened 2022-05-12
Fix issue 416

### LINKED ISSUE #416: Interface Equality
The following program results in an unexpected Z3 output because `Size_pkg_PMRect_Size_shape_SY$shape$gobra6$12__15$2_proof` is declared twice.
The problem is that Gobra identifies the provided implementation proof in pkg.gobra but generates an additional one.
The additional one is generated because Gobra does not identify `shape.Shape` in main.gobra as the same interface as `shape.Shape` in pkg.gobra (even though they even share the same declaration site).
Expected behavior: Gobra checks whether an implementation proof exists for (any) interface with the same structure as the currently considered one.

main.gobra
```
package main

// ##(-I ./ -I ./importedImplementationPkg)
import pkg "importedImplementationPkg"
import "importedImplementationPkg/shape"

func main () {
  r@ := pkg.Rect{width: 2, height: 3}
  r.Area()
  Area(&r)
}

requires s.mem() && s.Size()
func Area(s shape.Shape) (ret int) {
  return s.Area()
}
```

importedImplementationPkg/pkg.gobra
```
package pkg

// ##(-I ./)
import shape "shape"

type Rect struct {
  width int
  height int
}

pred (r *Rect) mem() {
  acc(r, 1/2)
}

requires acc(r, 1/2)
pure func (r *Rect) Size() bool {
  return r.width > 0 && r.height > 0
}

requires acc(r, 1/2) && r.Size()
ensures  acc(r, 1/2) && ret > 0
func (r *Rect) Area() (ret int) {
  return r.width * r.height
}

(*Rect) implements shape.Shape {
  pure (r *Rect) Size() bool {
    return unfolding acc(r.mem(), 1/2) in r.Size()
  }

  (r *Rect) Area() (ret int) {
    unfold r.mem()
    ret = r.Area()
    fold r.mem()
  }
}
```

importedImplementationPkg/shape/shape.gobra
```
package shape

type Shape interface {
   pred mem()

   requires acc(mem(), 1/2)
   pure Size() bool

   requires mem() && Size()
   ensures ret > 0
   Area() (ret int)
}
```

Felalolf:
Fixes #416 

The problem was that AbstractImport does not uniquely identify a package. In the example, the package shape was parsed twice, once for the import path "shape" and once for the import path "importedImplementationPkg/shape". To fix this I introduce the indirection AbstractPackage which is supposed to uniquely identify a package. It is not clear to me whether tjhe last folder of the import path is always sufficient, e.g. using "shape" for ".../shape". In this PR, I use the canonical path of the directory the package is stored in, which might be overkill.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/PackageResolver.scala
+++ b/src/main/scala/viper/gobra/frontend/PackageResolver.scala
@@ -32,12 +32,36 @@ object PackageResolver {
   val fileUriScheme = "file"
   val jarUriScheme = "jar"
 
-  trait AbstractImport
+  sealed trait AbstractImport
   /** represents an implicit unqualified import that should resolve to the built-in package */
   case object BuiltInImport extends AbstractImport
   /** relative import path that should be resolved; an empty importPath results in looking for files in the current directory */
-  case class RegularImport(importPath: String) extends AbstractImport {
+  case class RegularImport private (importPath: String) extends AbstractImport {
     override def toString: String = importPath
+  }
+
+  sealed trait AbstractPackage
+  /** represents an error */
+  case object NoPackage extends AbstractPackage
+  /** represents all built-in packages together */
+  case object BuiltInPackage extends AbstractPackage
+  /** represents a regular package */
+  case class RegularPackage (id: String) extends AbstractPackage
+
+  object AbstractPackage {
+    def apply(imp: AbstractImport)(config: Config): AbstractPackage = {
+      imp match {
+        case BuiltInImport => BuiltInPackage
+        case imp: RegularImport =>
+          getLookupPath(imp)(config) match {
+            case Left(_) => NoPackage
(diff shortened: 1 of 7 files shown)


ArquintL (approved):
Looks in general good to me. However, I'm wondering whether you could achieve the same by reusing `Source.getPackageInfo` that returns `PackageInfo` that uniquely identifies a package.

Felalolf:
I was not aware of the `Source.getPackageInfo` method. However, I have the impression that I cannot use it for this purpose, at least not as is. The method requires as argument a `Source` of the package you want an identifier for. Here, we want an id before we have parsed the package, so there are no `Source` instances yet. I could create an alternative version, which takes the package path and then parses the file until it reaches the import class. Can there be more than one package in the exact same folder?

jcp19 (approved):
LGTM

Issue #418, opened 2022-02-22
Whitespace-dependent bug: "Assertion unknown might not hold"


nicoberling:
Depending on the amount of empty lines after `package termination`, the test case `termination-fail-02.gobra` fails with the following message:

`<.:0:0> Postcondition might not hold. Assertion unknown might not hold.`
 
There does not seem to be a simple pattern (odd, even, ...) to predict whether it will fail with a given amount of empty lines. Below is the minimal code to duplicate this: It fails with 0,1 or 2 empty lines, succeeds with 3 and then fails again with 4. The same behaviour can be seen when parsing with ANTLR ( #388 ),  but for different numbers of empty lines.

```
// Any copyright is dedicated to the Public Domain.
// http://creativecommons.org/publicdomain/zero/1.0/

package termination

type Func interface {
	pure
	decreases n
	fun(n int, a Func) int
}

type FuncImpl struct {}

(FuncImpl) implements Func {
	pure (f FuncImpl) fun (n int, a Func) int {
		//:: ExpectedOutput(pure_function_termination_error)
		return f.fun(n, a)
	}
}

pure
decreases n
func (f FuncImpl) fun (n int, a Func) int {
	//:: ExpectedOutput(pure_function_termination_error)
	return n * a.fun(n, f)
}
```

jcp19:
Is this really dependent on the number of empty lines? Did you try multiple times with the same number of empty lines until you get a different result from before? This might be a non-deterministic issue

jogasser:
I encountered the same bug when working on the uniqueness of viper member names: 
Again adding a very specific amount of whitespaces fixes the error.

You may find the two files here: [termination_error.zip](https://github.com/viperproject/gobra/files/8196971/termination_error.zip)

termination-error.gobra.vpr contains the viper code that causes the error
termination-noerror.gobra.vpr contains the viper code where the error doesn't happend (I had to add 6 white spaces after the package clause)



The Gobra output for the correctly working version:
```
[info] Gobra has found 4 error(s) in package src/test/resources/regressions/features/termination - termination
[info]     <src/test/resources/regressions/features/termination/termination-fail-02.gobra:34:11> Function might not terminate.
[info] Required tuple condition might not hold.
[info]     <src/test/resources/regressions/features/termination/termination-fail-02.gobra:34:11> Function might not terminate.
[info] Required tuple condition might not hold.
[...]

PR #450, opened 2022-05-12
Nil Check with Termination Transformation

Felalolf:
Adds a nil check to the fallback function that is generated by the termination transformer. 
This also enables that the termination-fail-02 testcase can be included again. We could consider closing #418

### DIFF
--- a/src/main/scala/viper/gobra/ast/internal/transform/CGEdgesTerminationTransform.scala
+++ b/src/main/scala/viper/gobra/ast/internal/transform/CGEdgesTerminationTransform.scala
@@ -100,7 +100,7 @@ object CGEdgesTerminationTransform extends InternalTransform {
                   *     pure func (r recv) M (x1 T1, ..., xN TN) (res TRes)
                   *
                   *   we generate the following fallback:
-                  *     requires [PRE]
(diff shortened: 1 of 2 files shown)


Felalolf:
It worked on my machine. I will investigate why it is failing online.

ArquintL (approved):
I cannot reproduce the CI problem locally either. It seems very strange since it's complaining about a failed postcondition of the interface even though there isn't any

src/main/scala/viper/gobra/translator/Names.scala
```diff
@@ -25,6 +25,24 @@ object Names {
     scala.util.hashing.MurmurHash3.stringHash(s).toHexString
   }
 
+  object InterfaceMethod {
```
jcp19, line 28: As a tiny tiny nitpick (feel free to ignore), I would have named this object `InterfaceMethodProxy`

jcp19 (approved):

src/main/scala/viper/gobra/translator/Names.scala
Felalolf: It is in the names object. So, I feel that proxy is redundant.

PR #452, opened 2022-05-17
Added postconditions for interfaces

### LINKED ISSUE #446: Postconditions not allowed in specification of pure methods in interface definitions

Felalolf:
Fixes #446 

I still have to double check whether the substitution for the proof is correct, but I think it is. Input on that is welcome.

### DIFF
--- a/src/main/scala/viper/gobra/translator/encodings/interfaces/InterfaceEncoding.scala
+++ b/src/main/scala/viper/gobra/translator/encodings/interfaces/InterfaceEncoding.scala
@@ -686,7 +686,7 @@ class InterfaceEncoding extends LeafTypeEncoding {
     */
   private def function(p: in.PureMethod)(ctx: Context): MemberWriter[Vector[vpr.Function]] = {
     Violation.violation(p.results.size == 1, s"expected a single result, but got ${p.results}")
-    Violation.violation(p.posts.isEmpty, s"expected no postcondition, but got ${p.posts}")
+    // Violation.violation(p.posts.isEmpty, s"expected no postcondition, but got ${p.posts}")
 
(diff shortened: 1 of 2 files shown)


Felalolf:
I double checked the substitution and I think it is correct. The parameter are substituted because proof and interface method might have differently named formal arguments. However, the result is at this point already substituted with Viper's result expression, so no renaming is necessary (or even possible).

src/test/resources/regressions/issues/000446.gobra
```diff
@@ -0,0 +1,42 @@
...
+pred (s *S) m() {
+    acc(s) && s.x > 5
+}
+
+requires acc(s.m(), _)
+pure func (s *S) g() int {
+    return unfolding acc(s.m(), _) in s.x
+}
+
+//:: ExpectedOutput(generated_implementation_proof:postcondition_error:assertion_error)
```
ArquintL, line 41: it seems that this output occurs at g's precondition instead of here. Do you want to fix the annotation or the error location?

src/main/scala/viper/gobra/translator/encodings/interfaces/InterfaceEncoding.scala
```diff
@@ -686,7 +686,6 @@ class InterfaceEncoding extends LeafTypeEncoding {
     */
```
ArquintL, line 686: would be good to extend the encoding's description in the line above here

ArquintL: while you're on it, I would somehow indicate that there is one implication per implementation / type that implements the interface. This is not immediately clear in the current version of the encoding's description

ArquintL (approved):
One minor comment otherwise looks good to me

PR #456, opened 2022-05-19
Iota

jcp19:
Current annoyance: redundant messages are produced. E.g., for the following program
```go
const (
    // Should no be allowed, first assigment missing
    xxxx
    yyyy = 1
)
```
we get two errors, one of which is unnecessary:
```
[info] 	</Users/joao/tmp/iota_reject.gobra:5:5> /Users/joao/tmp/iota_reject.gobra:5:5:error: got unknown identifier xxxx
[info]     xxxx
[info]     ^
[info] 	</Users/joao/tmp/iota_reject.gobra:5:5> /Users/joao/tmp/iota_reject.gobra:5:5:error: missing init expr for Vector(xxxx)
[info]     xxxx
[info]     ^
```


jcp19:
Should be squashed and merged

Dspil (approved):

PR #457, opened 2022-05-20
Add Mem predicate in Error

jcp19:
This PR introduces a predicate `ErrorMem` in the `error` interface. An instance of this predicate is required to be able to call the `Error` method.

This PR also addresses the cyclic import error that is caused when `ErrorMem` is introduced. To solve that, we remove the call to method `tryUnqualifiedBuiltInPackageLookup` in method `tryUnqualifiedPackageLookup` when the importing package is `builtin`.


jcp19:
I cannot reproduce these errors locally

Dspil (approved):

PR #459, opened 2022-05-24
Avoid redundant test executions

jcp19:
As it stands, we run the Gobra tests twice on every push to a PR. This is redundant and wasteful. This PR addresses that by only executing the tests once when there is either a pr or push

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,11 @@
 
 name: test
 
-on: [push, pull_request]
+on:
+  push: # run this workflow on pushes to master
+    branches:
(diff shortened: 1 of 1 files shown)


ArquintL:
Not sure what you exactly want to achieve. Do we have the issue that we are running out of concurrent machines? Imho I like to get feedback on push whether CI goes through or not without having to open a PR first

jcp19:
Although this does not address the core issue, the goal was to reduce the number of times I need to re-trigger the workflows because one of two executions of the tests failed due to a non-deterministic bug in Z3.

PR #462, opened 2022-05-27
Add support for perm constants

jcp19:


src/test/resources/regressions/features/global_consts/global-const-8.gobra
```diff
@@ -0,0 +1,30 @@
...
+
+requires acc(x, p2)
+func test2(x *int) int {
+	return *x
+}
+
+func test3() {
+	assert i1 == perm(1/1)
+	assert i2 == perm(1/2)
+	assert i3 == perm(1/3)
```
ArquintL, line 29: I would maybe add a failing assert stmt. Also, you could relate e.g. `i1` to  `writePerm`

ArquintL (approved):

Dspil (approved):

Felalolf (approved):

PR #463, opened 2022-05-28
Add support for short-hand for implementation proofs

jcp19:
A typical pattern in some Go codebases is to add lines like the following to enforce that `ImplType` is an implementation of `InterfaceType`.
```go
var _ InterfaceType = ImplType(nil)
```
Because this does not have any effect at runtime, we can already add support for it without the more in-depth support for global variables. Having the line above is also compatible with explicitly providing an implementation proof in Gobra.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Desugar.scala
+++ b/src/main/scala/viper/gobra/frontend/Desugar.scala
@@ -375,7 +375,20 @@ object Desugar {
       in.Program(types.toVector, dMembers ++ additionalMembers, table)(meta(p))
     }
 
-    def varDeclGD(decl: PVarDecl): Vector[in.GlobalVarDecl] = ???
+    def varDeclGD(decl: PVarDecl): Vector[in.GlobalVarDecl] = {
+      val leftAreWildcards = decl.left.forall(_.isInstanceOf[PWildcard])
+      val typIsInterface = decl.typ.exists { p =>
+        val typ = info.symbType(p)
+        underlyingType(typ).isInstanceOf[Type.InterfaceT]
+      }
+      val rightArePureExpr = decl.right.forall(info.isPureExpression)
+      if (leftAreWildcards && typIsInterface && rightArePureExpr ) {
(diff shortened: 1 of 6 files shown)


Dspil (approved):

ArquintL (commented):
I agree that you can drop the global variable in this particular case. However, isn't it confusing that this is the way of checking the subtype relation in Go but Gobra doesn't check it?
The way I usually check the same thing is to simply write `ImpT implements ItfT`. This is automatically part of implementation proofs but if no implementation proof is necessary, I think it would make sense to generate this check for all automatically generated implementation proofs (resp. to make sure that an implementation proof is indeed generated)

ArquintL:
I've just seen that you drop the global variable in the desugarer which means that the expected subtype relation is checked and an implementation proof is indeed generated if necessary (as demonstrated by the unit tests)

ArquintL (approved):

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -375,7 +375,20 @@ object Desugar {
...
+        underlyingType(typ).isInstanceOf[Type.InterfaceT]
+      }
+      val rightArePureExpr = decl.right.forall(info.isPureExpression)
+      if (leftAreWildcards && typIsInterface && rightArePureExpr) {
+        // When the lhs is a wildcard of an interface type and the rhs are pure expressions,
+        // the variable declaration can be safely ignored. In some codebases, this idiom is
+        // used to check that types implement interfaces.
+        Vector()
+      } else ???
+    }
```
Felalolf, line 391: Gobra should check that the RHS type is a behavioral subtype of the LHS interface type. Is this currently the case? It would be nice if you could add a test with a failing generated implementation proof.

jcp19: I added a new test case to demonstrate the case where proof generation fails

Felalolf (approved):

PR #464, opened 2022-05-30
add variable to proofName to generate unique names

Dspil:
Trying to implement an interface for both a type and a pointer to the same type generated two functions implementing the interface with the same name. To fix this bug, an extra parameter to proofName was added containing the type of the receiver.


jcp19 (commented):
Can you please also add a test for this?

Dspil:
You mean a test that would fail before this change? If yes that might be hard because we couldn't reproduce it with Felix in a small file. I will try though

jcp19:
> You mean a test that would fail before this change? If yes that might be hard because we couldn't reproduce it with Felix in a small file. I will try though

Yes. In the worst case scenario, we could put the scion code that causes this problem as a test (after cleaning up a bit to remove the dependencies)

jcp19 (approved):
LGTM, we can merge it when the tests finish

PR #469, opened 2022-06-20
Simple defer

Felalolf:
I heard someone wanted a defer?


Dspil (approved):

src/main/scala/viper/gobra/translator/util/Core.scala
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  def core(@unused x: in.Expr): Core[in.LocalVar] = Core { args => (args.head, args.tail) }
+
+  def deepCore(x: in.PredicateConstructor): Core[in.PredicateConstructor] = {
+    for {
+      args <- vector(x.args.map(y => option(y.map(core))))
+    } yield x.copy(args = args)(x.info)
+  }
+
```
jcp19, line 64: Is `deepCore` meant to be accessible from the outside, or is it an internal auxiliar method which we can make private?

src/main/scala/viper/gobra/translator/encodings/typeless/DeferEncoding.scala
```diff
@@ -0,0 +1,78 @@
...
+
+  case class Defer(
+                    activationVar: vpr.LocalVarDecl,
+                    argumentVars: Vector[vpr.LocalVarDecl],
+                    stmt: MemberWriter[vpr.Stmt]
+                  ) extends ViperWriter.CodeCollectible
+
+  override def statement(ctx: Context): in.Stmt ==> CodeWriter[vpr.Stmt] = {
+    case n: in.Defer =>
+      val (pos, info, errT) = n.vprMeta
```
jcp19, line 31: I know that this is not super easy for `defer` because its encoding has an impact in the entire control flow of the method, but would it be possible somehow if we had a sketch of the encoding of this feature? If not, we can merge it as it is

jcp19 (approved):
LGTM: just two minor comments

PR #470, opened 2022-06-22
Make `defer` statement ghostifiable

jcp19:
Just noticed that `defer` cannot be used in ghost contexts, which is useful when deferring a fold/unfold.


Felalolf (approved):
I would also check that you cannot ghostify an actual call, but you can also do that in a later PR.

PR #471, opened 2022-06-23
Improve trigger for acc(s) when s is a slice

### LINKED ISSUE #238: Access to a slice pointer should yield access to its cells

jcp19:
Changes the trigger of `acc(s)` when `s` is a slice to be heap-independent. The generated trigger should be the same as the one inferred for the assertion `forall i int :: 0 <= i && i < len(s) ==> acc(&s[i])`

Fixes #238

### DIFF
--- a/src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
+++ b/src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
@@ -47,26 +47,17 @@ class SliceEncoding(arrayEmb : SharedArrayEmbedding) extends LeafTypeEncoding {
       case Shared => vpr.Ref
     }
   }
+
   /**
     * Encodes assertions.
-    * [acc(m: []T, perm)] -> [forall i int :: 0 <= i && i < len(m) ==> acc(&m[i], perm)]
+    *     [acc(m: []T, perm)] -> getCellPerms(m, perm, SliceBound.Len)
     */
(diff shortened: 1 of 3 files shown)


src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -215,31 +206,42 @@ class SliceEncoding(arrayEmb : SharedArrayEmbedding) extends LeafTypeEncoding {
     }
   }
 
-
   /**
     * Obtains permission to all cells of a slice
-    * getCellPerms[loc: []T] -> forall idx :: {loc(a, idx) 0 <= idx < cap(l) ==> Footprint[ loc[idx] ]
+    *     getCellPerms[loc: []T] ->
+    *         forall idx :: { loc(a, idx) } 0 <= idx < cap(l) ==> Footprint[ loc[idx] ]
```
Felalolf, line 212:
```suggestion
    *         forall idx :: { loc(a, idx) } 0 <= idx < bound ==> Footprint[ loc[idx] ]
    *            where bound is len(l) if sliceBound = SliceBound.Len, and cap(l) otherwise.
```

Felalolf (approved):
I am not sure that this is a good choice, but it is a better choice than what we have right now and I do not have a better idea.

PR #479, opened 2022-07-11
Introduce support for closures

stefanomil:
This PR introduces support for closure verification in Gobra.

The new primitives are:

- **Function literals**, that can capture variables and can be assigned to another variable. They are treated as expressions with function type. Example:
```
     x@ := 0
     c := preserves acc(&x)
            ensures x == old(x) + n && m == x
            func f(n int) (m int) {
                  x += n;
                  return x
            }
```

- **Function and method objects**, operands with function types obtained from functions or methods. They can be assigned to variables and treated as closures. For example, assuming there exist function `f1` and method `SomeStruct.m1`, one can write `c := f1` or `c := SomeStruct{...}.m1`.

- **Closure spec instances**, entities that use existing functions or function literals, in combination with 0 or more parameters, to express properties of closures. For examples, assume that there exists function (or literal) `func fsp(n int, m bool) (r int)`. The possible closure spec instances are `fsp`, `fsp{2}`, `fsp{m: true}`,  `fsp{2, true}`, which have respective types `func(int, bool)int`, `func(bool)int`, `func(int)int`, `func()int`. Each parameter is associated with one argument of the base function.

- **Implementation assertions**, which assert that a closure implements a particular spec. Example: `c implements fsp{m: true}`. This means that it can be called assuming the preconditions of `fsp`, and ensuring the postconditions of `fsp`.

- **Calls with spec**, which allow to call a closure using some spec that we know it implements. Example: `c(3) as fsp{m: true}`. For the purposes of verification, this is equivalent to calling `fsp(3, true)`, with the additional precondition that `c` implements `fsp{m: true}`.

- **Closure spec implementation proofs**, that allow to verify that a closure implements a certain spec. Example:
```
proof c implements fsp{m: true} {
     [fold, unfold]
     r = c(n) as f
     [fold, unfold]
}
```
The proof verifies if, assuming the preconditions of `fsp` and assigning value `true` to the argument `m`, it's possible to establish the preconditions of `f` and, after the call, the postconditions of `fsp` are established. After this statement, the assertion `c implements fsp{m: true}` holds.

Spec implementation proofs can also be used with function or method objects. In this case, the call in the proof can be a normal function/method invocation. For example, assuming `f1` is a function, we can write (if the type of the function matches the type of the spec):
```
proof f1 implements fsp{3} {
   ...
   r = f1(m)
   ...
}
```
From now on `f1 implements fsp{3}` holds and it's possible to call `f1(_) as fsp{3}`. Similarly for methods.


src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -249,13 +257,19 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
             case PredT(args) =>
               if (n.args.isEmpty && args.isEmpty) noMessages
               else multiAssignableTo.errors(n.args map exprType, args)(n) ++ n.args.flatMap(isExpr(_).out)
-            case c => Violation.violation(s"This case should be unreachable, but got $c")
+            case c => Violation.violation(s"This caseA should be unreachable, but got $c")
```
jcp19, line 260: you rewrote "case" multiple times to "caseA", "caseB", etc. Did you do this for debugging? If so, I would revert these changes. (For debugging, we can always use the line number reported in the stack trace)

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -54,6 +61,7 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
         // imported members, we simply assume that they are wellformed (and were checked in the other package's context)
         case Some(_: ap.Constant) => noMessages
         case Some(_: ap.Function) => noMessages
+        case Some(_: ap.Closure) => violation(s"the name of a function literal should not be accessible from a different package")
```
jcp19, line 64: What does it mean exactly? Does it mean that you cannot assign a function to a variable if that function is declared in another package? If so, why do we have that restriction?

jcp19 (commented):
I started looking at this but did not have a lot of time to continue. I will hopefully continue soon

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
stefanomil: No, we cannot access a function literal from another package (simply because they are declared within functions, so they are never visible from another package). This is not an additional restriction, just a sanity check - and to stop the compiler from warning about missing cases.

Functions from a different package can be assigned to variables, and they would match case (_: ap.Function)

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
stefanomil: Fixed

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -416,12 +423,33 @@ case class PExpCompositeVal(exp: PExpression) extends PCompositeVal // exp is ne
...
+
+case class PClosureImplements(closure: PExpression, spec: PClosureSpecInstance) extends PGhostExpression
+
+case class PClosureImplProof(impl: PClosureImplements, block: PBlock) extends PGhostStatement with PScope
 
 case class PInvoke(base: PExpressionOrType, args: Vector[PExpression]) extends PActualExpression
 
 // TODO: Check Arguments in language specification, also allows preceding type
 
+case class PCallWithSpec(base: PExpression, args: Vector[PExpression], spec: PClosureSpecInstance) extends PActualExpression
```
Felalolf, line 451: I would maybe add a spec option to invoke to reduce redundancy. However, I am not very sure about this.

src/main/scala/viper/gobra/frontend/info/implementation/resolution/NameResolution.scala
```diff
@@ -301,9 +307,51 @@ trait NameResolution { this: TypeInfoImpl =>
...
+            tryPackageLookup(RegularImport(pkg.decl.importPath), id, pkg.decl) match {
+              case Some((f: Function, _)) => (f, f.decl.body, f.context)
+              case _ => return UnknownEntity()
+            }
+          case _ => return UnknownEntity()
+        }
+      case _ => return UnknownEntity()
+    }
+
+    def addressable: Boolean = fBody.exists(_._1.shareableParameters.exists(_.name == id.name))
```
Felalolf, line 336: this code looks quite complicates. Could you not call invoke on `proof.impl.spec.func` and from there extract the symbol. Also, the in and out paramters of a proof are always exclusive, i.e. the addressable method should always return false.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -28,8 +28,15 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
   lazy val wellDefExprAndType: WellDefinedness[PExpressionAndType] = createWellDef {
-
-    case _: PNamedOperand => noMessages // no checks to avoid cycles
+    case n: PNamedOperand =>
+      resolve(n) match {
+        /* A closure name can only be used as a spec, if we are not directly within the closure itself
+           (the same limitation applies within closures nested inside the closure itself) */
+        case Some(ap.Closure(id, _)) => error(n, s"here, the closure name ${n} can only be used as a spec",
+          !tree.parent(n).head.isInstanceOf[PClosureSpecInstance] &&
+            tryEnclosingFunctionLit(n).fold(true)(lit => lit.decl.id.fold(true)(encId => encId.name != id.name)))
```
Felalolf, line 37: I do not think that you need such a strict restrictions. You could introduce an additional internal type, e.g. `spec` and then the usual type rules should enforce that a spec id is not used in a context that it should not be used in.

src/main/scala/viper/gobra/frontend/info/implementation/typing/MiscTyping.scala
```diff
@@ -106,7 +110,7 @@ trait MiscTyping extends BaseTyping { this: TypeInfoImpl =>
 
       case tree.parent.pair(_: PCompositeVal, ke: PKeyedElement) => expectedMiscType(ke)
 
-      case c => Violation.violation(s"This case should be unreachable, but got $c")
+      case c => Violation.violation(s"This caseE should be unreachable, but got $c")
```
Felalolf, line 113: typo

Felalolf (commented):
some comments

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -416,12 +423,33 @@ case class PExpCompositeVal(exp: PExpression) extends PCompositeVal // exp is ne
 
 case class PLitCompositeVal(lit: PLiteralValue) extends PCompositeVal
 
-case class PFunctionLit(args: Vector[PParameter], result: PResult, body: PBlock) extends PLiteral with PCodeRootWithResult with PScope
+case class PFunctionLit(decl: PClosureNamedDecl) extends PLiteral with PActualExpression
+
+case class PClosureNamedDecl(id: Option[PIdnDef], decl: PClosureDecl) extends PGhostMisc {
+  val result: PResult = decl.result
+}
```
jcp19, line 430: this name might be a bit misleading: I believe that all other nodes that have "Named" in the name have a non-optional identifier of type `PIdnDef`. Additionaly, the extra indirection through `decl` is maybe unnecessary. Maybe one could make the class structure a bit simpler and follow the patterns already existent in the codebase (e.g., with `PNamedParameter` and `PUnnamedParameter`) with the following code:

```scala
// replaces PFunctionLit and PClosureDecl
sealed trait PClosureDecl extends PLiteral with PCodeRootWithResult { // "extends PActualExpression" is redundant here
    def args: Vector[PParameter]
    def result: PResult
    def spec: PFunctionSpec
    def body: Option[(PBodyParameterInfo, PBlock)])
}

// replaces PClosureNamedDecl when id != None
case class PNamedClosureDecl(
    id: PIdnDef,
    override val args: Vector[PParameter],
    override val result: PResult
    override val spec: PFunctionSpec
    override val body: Option[(PBodyParameterInfo, PBlock)]) extends PClosureDecl

// replaces PClosureNamedDecl when id == None
case class PUnnamedClosureDecl(
    override val args: Vector[PParameter],
    override val result: PResult
    override val spec: PFunctionSpec
    override val body: Option[(PBodyParameterInfo, PBlock)]) extends PClosureDecl
```
Additionally, this avoids making closure declarations subtypes of `PMisc` which I think makes more sense.

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -416,12 +423,33 @@ case class PExpCompositeVal(exp: PExpression) extends PCompositeVal // exp is ne
...
+case class PClosureDecl(args: Vector[PParameter],
+                  result: PResult,
+                  spec: PFunctionSpec,
+                  body: Option[(PBodyParameterInfo, PBlock)]) extends PFunctionOrClosureDecl with PCodeRootWithResult with PActualMisc
+
+case class PClosureSpecInstance(func: PNameOrDot, params: Vector[PClosureSpecParameter]) extends PGhostMisc
+
+case class PClosureSpecParameter(key: Option[PClosureSpecParameterKey], exp: PExpression) extends PGhostMisc
+
+case class PClosureSpecParameterKey(name: String) extends PGhostMisc
```
jcp19, line 441: Do we really need `PClosureSpecParameter` and `PClosureSpecParameterKey`? Couldn't we use `PKeyedElement` here instead? Or does it become cumbersome because we would need to deal with `PCompositeVal`s?

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -243,6 +244,38 @@ object Desugar {
       in.MPredicateProxy(id.name, name)(meta(id, context))
     }
 
+    def functionLitProxyD(lit: PFunctionLit, context: TypeInfo): in.FunctionLitProxy = {
```
jcp19, line 247: did you consider using `in.FunctionProxy` instead of introducing `in.FunctionLitProxy`? If so, why did you decide against it?

src/main/antlr4/GobraParser.g4
```diff
@@ -337,6 +348,15 @@ primaryExpr:
   ) L_PAREN expression R_PAREN #builtInCallExpr // Remove this alternative when predeclared functions are properly handled
   ;
 
+// Added function literal with specification
+operand: literal | operandName | L_PAREN expression R_PAREN;
+
+literal: basicLit | compositeLit | functionLit;
```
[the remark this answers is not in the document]
stefanomil: That is true. I have removed them

src/main/antlr4/GobraParser.g4
```diff
@@ -155,6 +155,15 @@ assertion:
 
 blockWithBodyParameterInfo: L_CURLY (SHARE identifierList eos)? statementList? R_CURLY;
 
+// Closures
+closureSpecInstance: (qualifiedIdent | IDENTIFIER) (L_CURLY (closureSpecParams COMMA?)? R_CURLY)?;
+
+closureSpecParams: closureSpecParam (COMMA closureSpecParam)*;
+
+closureSpecParam: (IDENTIFIER COLON)? expression;
```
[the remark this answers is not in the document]
stefanomil: Yes, I think it's fine because this always comes after "as" or "implements". Defining it as `qualifiedIdent | IDENTIFIER | compositeLit` might work, but I think it would cause complications (it would allow some invalid code to be parsed, so there would need to be additional checks). Maybe I will try, after resolving the other comments, to see what happens

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -355,6 +360,7 @@ case class FunctionCall(targets: Vector[LocalVar], func: FunctionProxy, args: Ve
 case class MethodCall(targets: Vector[LocalVar], recv: Expr, meth: MethodProxy, args: Vector[Expr])(val info: Source.Parser.Info) extends Stmt with Deferrable
 case class GoFunctionCall(func: FunctionProxy, args: Vector[Expr])(val info: Source.Parser.Info) extends Stmt
 case class GoMethodCall(recv: Expr, meth: MethodProxy, args: Vector[Expr])(val info: Source.Parser.Info) extends Stmt
+case class CallWithSpec(targets: Vector[LocalVar], closure: Expr, args: Vector[Expr], spec: ClosureSpec)(val info: Source.Parser.Info) extends Stmt
```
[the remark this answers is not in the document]
stefanomil: I will do this tomorrow, I don't expect it to be a big change

src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
```diff
@@ -1063,10 +1063,48 @@ class ParseTreeTranslator(pom: PositionManager, source: Source, specOnly : Boole
...
+  override def visitClosureSpecParams(ctx: ClosureSpecParamsContext): Vector[PClosureSpecParameter] =
+    if (ctx == null) Vector.empty else super.visitClosureSpecParams(ctx) match {
+      case v: Vector[Any] => v collect { case p: PClosureSpecParameter => p }
+      case p: PClosureSpecParameter => Vector(p)
+    }
+
+  override def visitClosureSpecParam(ctx: ClosureSpecParamContext): PClosureSpecParameter = {
+    val id = if(ctx.IDENTIFIER() == null) None else Some(PIdnNodeEx(PClosureSpecParameterKey, _ => None).get(ctx.IDENTIFIER()))
+    PClosureSpecParameter(id, visitNode[PExpression](ctx.expression()))
+  }
```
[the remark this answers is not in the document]
stefanomil: I have changed them to look more similar to `visitCompositeLit`.  `visitClosureSpecInstance` is a bit longer now, but probably easier to read

src/main/scala/viper/gobra/ast/frontend/Ast.scala
stefanomil: One indirection is (almost) necessary: the id of the literal needs to be outside of the scoped node, so that it can be referenced outside (as spec). Without this indirection, we would need to add ad-hoc rules in the Name Resolution, which I think would be worse. 

The second indirection was there so that the id of the literal would be considered ghost, so that it cannot be used as a normal closure in actual expressions. But it is possible to do without too much additional effort, so I have opted for removing the second indirection: there is now only `PFunctionLit` and `PClosureDecl`.

Hopefully the two remaining names are clear enough. PClosureDecl is still a `Misc`, but at least there is only one

src/main/scala/viper/gobra/ast/frontend/Ast.scala
stefanomil: I remember trying to use `PKeyedElement` and that caused some problems with name resolution, but I can't remember exactly so I will double check and see what happens if I try using that

src/main/scala/viper/gobra/frontend/info/implementation/resolution/NameResolution.scala
stefanomil: Calling `resolve` on `proof.impl.spec.func` does not work because it causes a circular dependency (the method is called within `entity`)

I have now set the addressability of all parameters to `false`

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
stefanomil: I think this is a bit less restrictive than having a separate type `spec`.

A rule for a `spec` type, in my understanding, would be that operands can never be of type `spec`, and a `spec` cannot be the same as a `func` (tell me if that's wrong). But, within a closure declaration itself, it is useful to assert `c == name`, where `c` is some closure (either captured or argument), and `name` is the name of the declaration. This means that the verifier knows that the variables captured by `c` are the same as the `current` closure - this is necessary for recursion.

Initially, I also allowed the closure declaration name to be used outside of the declaration itself (in ghost code). But that turned out to be a lot of additional code in the desugarer, and not too useful - a user can always alias the literal by themselves, if they need to.

I have changed the message to `expected valid operand, got closure declaration name $n` to make it a bit more consistent with other error messages.

src/main/scala/viper/gobra/ast/frontend/Ast.scala
stefanomil: I tried; using `PKeyedElement` does cause problems with name resolution, but those could be solved relatively simply by adding a new rule in the `entity` attribute, so I opted for it.

So I have removed `PClosureSpecParameter` and `PClosureSpecParameterKey`

src/main/scala/viper/gobra/frontend/info/implementation/resolution/NameResolution.scala
stefanomil: I have now split the method into 2 simpler ones, and made some other small changes

src/main/scala/viper/gobra/ast/internal/Program.scala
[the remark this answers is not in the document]
stefanomil: Done, I have also deleted PCallWithSpec and added a spec option to PInvoke, since that made things easier

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -882,6 +888,7 @@ case class MapValues(exp : Expr, expUnderlyingType: Type)(val info : Source.Pars
 case class PureFunctionCall(func: FunctionProxy, args: Vector[Expr], typ: Type)(val info: Source.Parser.Info) extends Expr
 case class PureMethodCall(recv: Expr, meth: MethodProxy, args: Vector[Expr], typ: Type)(val info: Source.Parser.Info) extends Expr
 case class DomainFunctionCall(func: DomainFuncProxy, args: Vector[Expr], typ: Type)(val info: Source.Parser.Info) extends Expr
+case class PureCallWithSpec(closure: Expr, args: Vector[Expr], spec: ClosureSpec, typ: Type)(val info: Source.Parser.Info) extends Expr
```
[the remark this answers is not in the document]
stefanomil: Done

stefanomil:
@Felalolf @jcp19 I was thinking again about the changes related to `CallWithSpec` and I'm not convinced, as I think it adds redundancy.

What we have now, after the changes, is:

```
case class FunctionCall(targets: Vector[LocalVar], func: Either[Expr, FunctionProxy], args: Vector[Expr], spec: Option[ClosureSpec])(val info: Source.Parser.Info) extends Stmt with Deferrable {
  require(func.isLeft || spec.isEmpty)
}
case class MethodCall(targets: Vector[LocalVar], recv: Expr, meth: MethodProxy, args: Vector[Expr], spec: Option[ClosureSpec])(val info: Source.Parser.Info) extends Stmt with Deferrable
```
I find the fact that the `func` parameter can both be a proxy and an expression a bit ugly. Also, if the spec is not `None`, any of these will be turned into a `FunctionCall` with a function/method object as closure, so the same thing can basically be expressed in 2 different ways. What I propose, instead, is to do these transformations at the desugarer level, so in the internal AST we have:

```
case class FunctionCall(targets: Vector[LocalVar], func: FunctionProxy, args: Vector[Expr])(val info: Source.Parser.Info) extends Stmt with Deferrable
case class MethodCall(targets: Vector[LocalVar], recv: Expr, meth: MethodProxy, args: Vector[Expr])(val info: Source.Parser.Info) extends Stmt with Deferrable
case class ClosureCall(targets: Vector[LocalVar], closure: Expr, args: Vector[Expr], spec: ClosureSpec)(val info: Source.Parser.Info) extends Stmt
```

For Go closure calls, we would only need to add a new `case class GoClosureCall`, so it's not much. Or we could make all 3 extend a `sealed trait CallStmt` and have a general `GoCall` which takes it as parameter.

The same applies to `PureFunctionCall`s

Felalolf:
I think your suggestion is ok. @jcp19 At some point, we should unify the nodes, e.g. something like:

```
sealed trait CallBase
case class FunctionBase(func: FunctionProxy) extends CallBase
case class MethodBase(recv: Expr, meth: MethodProxy) extends CallBase
case class ClosureBase(closure: Expr, spec: ClosureSpec) extends CallBase

sealed trait Call extends Node
case class CallStmt(targets: Vector[LocalVar], base: CallBase, args: Vector[Expr])(val info: Source.Parser.Info) extends Call with Stmt
case class CallExpr(base: CallBase, args: Vector[Expr])(val info: Source.Parser.Info) extends Call with Expr

case class Go(call: Call)(val info: Source.Parser.Info) extends Stmt
```

Felalolf (commented):
I looked at the encoding files. I have some comments, but overall looks quite positive. You definitely need more documentation. I still need to review the desugarer and the encoding parts (after you add documentation).

src/main/scala/viper/gobra/reporting/VerifierError.scala
```diff
@@ -285,6 +286,16 @@ case class GeneratedImplementationProofError(subT: String, superT: String, error
   }
 }
 
+case class MethodObjectGetterPreconditionError(info: Source.Verifier.Info) extends VerificationError {
+  override def localId: String = "method_object_nil_error"
+  override def localMessage: String = s"Cannot get method object ${info.origin.tag}"
```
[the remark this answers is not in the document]
stefanomil: Changed to s"The receiver of ${info.origin.tag} might be nil"

src/main/scala/viper/gobra/translator/encodings/closures/MethodObjectEncoder.scala
```diff
@@ -0,0 +1,109 @@
...
+      vpr.EqCmp(interfaces.dynTypeOf(recvExp)(pos, info, errT)(ctx), types.typeToExpr(t)(pos, info, errT)(ctx))(pos, info, errT),
+      vpr.EqCmp(vpr.Result(domain.vprType)(), defTCall(t))())()
+    val posts = ctx.table.implementations(recvType).toSeq.map(t => valueMatchesIfTypeIs(t.asInstanceOf[in.DefinedT]))
+    ml.unit(vpr.Function(proxy.uniqueName, Seq(ctx.variable(receiver)), domain.vprType, Seq(recvNotNil), posts, None)())
+  }
+
+  private def receiverNilErr(m: in.MethodObject, call: vpr.Exp): ErrorTransformer = {
+    case vprerr.PreconditionInAppFalse(offendingNode, _, _) if call eq offendingNode =>
+      val info = m.vprMeta._2.asInstanceOf[Source.Verifier.Info]
+      val recvInfo = m.recv.vprMeta._2.asInstanceOf[Source.Verifier.Info]
```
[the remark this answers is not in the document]
stefanomil: That doesn't work because `m.info` has type `Source.Parser.Info`, while the required one is `Source.Verifier.Info`.

src/main/scala/viper/gobra/translator/encodings/closures/MethodObjectEncoder.scala
```diff
@@ -0,0 +1,109 @@
...
+import viper.gobra.translator.context.Context
+import viper.gobra.translator.encodings.interfaces.{InterfaceComponent, InterfaceComponentImpl, InterfaceUtils, PolymorphValueComponent, PolymorphValueComponentImpl, TypeComponent, TypeComponentImpl}
+import viper.gobra.translator.util.ViperWriter.CodeLevel.errorT
+import viper.gobra.translator.util.ViperWriter.{CodeWriter, MemberWriter, MemberLevel => ml}
+import viper.gobra.util.Violation.violation
+import viper.silver.verifier.{errors => vprerr}
+import viper.silver.{ast => vpr}
+
+
+class MethodObjectEncoder(domain: ClosureDomainEncoder) {
```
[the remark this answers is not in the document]
stefanomil: Looks good, the only thing I've changed is that I've added a separate case where a `DefinedT` is an interface. In that case, the conversion should be the same as for an interface type (i.e. the method object should be the one corresponding to the concrete implementation).

I have also added a `decreases _` precondition, which allows to get a method object within a method that checks termination (I've done a similar thing for the getters and implements functions).

src/main/scala/viper/gobra/translator/encodings/closures/ClosureEncoding.scala
```diff
@@ -0,0 +1,170 @@
...
+  /** Encodes a spec implementation proof as:
+    *
+    * if (*) {
+    *   initialize args and results
+    *   assign parameters
+    *   while(true) {
+    *     inhale precondition of spec
+    *     [ body ]
+    *     exhale postcondition of spec
+    *   }
```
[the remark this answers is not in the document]
stefanomil: I don't understant what the problem with an infinite loop (with a `decreases _` invariant) is. What do you mean with "it does not break in the presence of inlining"? Inlining of what?

For now, I'm changing it to the following.
```
[ args, results already initialised ]
[ parameters already assigned ]
if (*) {
       while(true)
       decreases _ // assume that the loop terminates
      {
          inhale precondition of spec
         [ body ]
         exhale postcondition of spec
      }
    assume false
}
```
Changes:
- Args, results and parameters are initialised and assigned an a block enclosing the if-statement generated by the desugarer. It already worked like this, I had forgotten to reflect this in the comment.
- I added the `decreases _` invariant
- I added the `assume false` at the end
- The `*` variable is generated here instead of the desugarer

I have also added the testcase, which passes. To make it pass, I also had to add empty termination measures to implements functions, getters and conversions

src/main/scala/viper/gobra/translator/encodings/closures/ClosureEncoding.scala
```diff
@@ -0,0 +1,170 @@
...
+          case reason: reasons.InsufficientPermission => reporting.SpecImplementationPostconditionError(info, proof.spec.info.tag)
+            .dueTo(reporting.InsufficientPermissionError(reason.offendingNode.info.asInstanceOf[Source.Verifier.Info]))
+          case _ => reporting.SpecImplementationPostconditionError(info, proof.spec.info.tag)
+        }
+    }
+
+    val (pos, info, errT) = proof.vprMeta
+
+    for {
+      ndBoolTrue <- ctx.assertion(in.ExprAssertion(proof.ndBool)(proof.info))
```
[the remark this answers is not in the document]
stefanomil: Oh cool, I didn't know that. Done

src/main/scala/viper/gobra/translator/encodings/closures/ClosureDomainEncoder.scala
```diff
@@ -0,0 +1,40 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2011-2022 ETH Zurich.
+
+package viper.gobra.translator.encodings.closures
+
+import viper.gobra.translator.Names
+import viper.silver.{ast => vpr}
+
+class ClosureDomainEncoder(specs: ClosureSpecsEncoder) {
```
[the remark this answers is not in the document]
stefanomil: Done

src/main/scala/viper/gobra/translator/encodings/closures/ClosureEncoding.scala
```diff
@@ -0,0 +1,170 @@
...
+      addr match {
+        case Exclusive => domain.vprType
+        case Shared    => vpr.Ref
+      }
+  }
+
+  /**
+    * Encodes literal function expressions and calls to the default closure generator.
+    */
+  override def expression(ctx: Context): in.Expr ==> CodeWriter[vpr.Exp] = default(super.expression(ctx)){
```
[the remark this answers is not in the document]
stefanomil: Done

src/main/scala/viper/gobra/translator/encodings/closures/ClosureEncoding.scala
```diff
@@ -0,0 +1,170 @@
...
+
+    case e@in.DfltVal(_) :: ctx.Function(t) / Exclusive => dfltVal(e, t)(ctx)
+
+    case e@in.NilLit(_) :: ctx.Function(t) => dfltVal(e, t)(ctx)
+  }
+
+  private def dfltVal(e: in.Expr, fTyp: in.FunctionT)(ctx: Context): CodeWriter[vpr.Exp] =
+    ctx.expression(in.PureFunctionCall(Right(in.FunctionProxy(Names.closureDefaultFunc)(e.info)), Vector.empty, None, fTyp)(e.info))
+
+  override def statement(ctx: Context): in.Stmt ==> CodeWriter[vpr.Stmt] = default(super.statement(ctx)) {
```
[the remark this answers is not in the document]
stefanomil: Done

src/main/scala/viper/gobra/translator/encodings/closures/ClosureSpecsEncoder.scala
```diff
@@ -0,0 +1,228 @@
...
+import viper.gobra.translator.Names
+import viper.gobra.translator.context.Context
+import viper.gobra.translator.util.ViperWriter.CodeLevel.errorT
+import viper.gobra.translator.util.ViperWriter.MemberKindCompanion.ErrorT
+import viper.gobra.translator.util.ViperWriter.{CodeWriter, MemberWriter}
+import viper.gobra.util.Violation
+import viper.silver.verifier.{reasons, errors => vprerr}
+import viper.silver.{ast => vpr}
+
+protected class ClosureSpecsEncoder {
```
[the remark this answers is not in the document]
stefanomil: Done

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -164,21 +165,21 @@ object Desugar {
...
 
-    type Identity = Meta
+    type Identity = (Meta, TypeInfo)
 
-    private def abstraction(id: PIdnNode): Identity = {
+    private def abstraction(id: PIdnNode, info: TypeInfo = info): Identity = {
       val entity = info.regular(id)
-      meta(entity.rep)
+      (meta(entity.rep, entity.context.getTypeInfo), entity.context.getTypeInfo)
     }
```
jcp19, line 173: why do we now need to pass `info` as a parameter?

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -433,10 +471,31 @@ object Desugar {
...
 
+    private case class FunctionInfo(args: Vector[in.Parameter.In],
+                                    captured: Vector[(in.Expr, in.Parameter.In)],
+                                    results: Vector[in.Parameter.Out],
+                                    pres: Vector[in.Assertion],
+                                    posts: Vector[in.Assertion],
+                                    terminationMeasures: Vector[in.TerminationMeasure],
+                                    body: Option[in.MethodBody])
+
+    private def functionMemberOrLitD(decl: PFunctionOrClosureDecl, fsrc: Meta, outerCtx: FunctionContext = null): FunctionInfo = {
```
jcp19, line 488: I wouldn't use `null` here because it makes it tricky to understand all places where this `null` might propagate to. If you really need the possibility of not passing a `FunctionContext`, I would make this of type `Option[FunctionContext]`

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -926,7 +1034,7 @@ object Desugar {
                 for {
                   re <- goE(r)
                   dL <- leftOfAssignmentD(l)(re.typ)
-                  le = in.Assignee.Var(dL)
+                  le = dL
```
jcp19, line 1037: why declare `le`? It appears useless here

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -164,21 +165,21 @@ object Desugar {
 
     private val nm = new NameManager
 
-    type Identity = Meta
+    type Identity = (Meta, TypeInfo)
 
-    private def abstraction(id: PIdnNode): Identity = {
+    private def abstraction(id: PIdnNode, info: TypeInfo = info): Identity = {
```
jcp19, line 170: If the reason to pass the `info` object as a parameter is that we are running into `node not in tree exception`, then I would not use a default value for this parameter, as doing so might still hide some bugs from before where we pass the wrong info. Having to think explicitly about every `info` that we pass, while annoying, is probably safer.

jcp19 (commented):
Some minor comments

src/main/scala/viper/gobra/frontend/Desugar.scala
stefanomil: It's to avoid getting a `node not in tree exception` (as you said in the other comment)

src/main/scala/viper/gobra/frontend/Desugar.scala
stefanomil: Good point, I opted for always requiring one (as a caller can always make a dummy FunctionContext anyway)

src/main/scala/viper/gobra/frontend/Desugar.scala
stefanomil: Agreed, I have removed the default value from all functions that had it

src/main/scala/viper/gobra/frontend/Desugar.scala
stefanomil: I added that to help with debugging (the debugger wouldn't show me the value otherwise), and I forgot to remove if after. Fixed

src/main/scala/viper/gobra/translator/encodings/closures/ClosureEncoding.scala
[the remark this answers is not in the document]
stefanomil: I have now changed it to the second version. The updated comment is this:

```
    [ args, results already initialised ]
    [ parameters already assigned ]
    if (*) {
      var numIterations: Int
      numIterations := proofIterator()
      while(numIterations > 0)
        decreases numIterations
      {
        inhale precondition of spec
        [ body ]
        exhale postcondition of spec
        numIterations := numIterations - 1
      }
      assume false
    }
    inhale closureImplements$[spec]([closure], [params])
```

src/main/scala/viper/gobra/frontend/info/implementation/resolution/NameResolution.scala
```diff
@@ -301,9 +309,56 @@ trait NameResolution { this: TypeInfoImpl =>
...
+    }
+
+    // Within a spec implementation proof or closure instance, consider all arguments of the spec non-ghost and all results ghost.
+    // This is to be as permissive as possible at this point, since all ghostness-related checks are done later on.
+    lazy val inParam = func.args.map(namedParam).find(p => p.nonEmpty && p.get.id.name == id.name)
+      .map(p => InParameter(p.get, ghost = false, addressable = false, func.context))
+    if (inParam.nonEmpty) Some(inParam.get)
+    else  func.result.outs.map(namedParam).find(p => p.nonEmpty && p.get.id.name == id.name)
+      .map(p => OutParameter(p.get, ghost = true, addressable = false, func.context))
+  }
```
jcp19, line 343: could this be simplified to the following?
```scala
    lazy val inParam = func.args.flatMap(namedParam)
      .find(_.id.name == id.name)
      .map(InParameter(_, ghost = false, addressable = false, func.context))
    
    inParam.orElse{
      func.result.outs.flatMap(namedParam)
        .find(_.id.name == id.name)
        .map(OutParameter(_, ghost = true, addressable = false, func.context))
    }
  }
```

jcp19: Also, why did you make `inParam` lazy? It seems that the value needs to be computed anyway right after the declaration, so lazy here may not be needed

stefanomil: Yes, good suggestion

stefanomil: I'm not sure why I made `inParam` lazy, I probably implemented this in a slightly different way at the beginning which made it being lazy useful, and when I changed it I didn't notice it and forgot to remove it

stefanomil: Done

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1467,20 +1577,99 @@ object Desugar {
...
+              } yield Left((targets, methodCall(targets, recv, mproxy, convertedArgs, resT, spec)))
             }
           }
           case sym => Violation.violation(s"expected symbol with arguments and result or a built-in entity, but got $sym")
         }
       }
     }
 
+    /** Returns either a tuple with targets and call statement or, if the call is pure, the call expression directly. */
+    def closureCallDAux(ctx: FunctionContext, info: TypeInfo)(expr: PInvoke)(src: Meta): Writer[Either[(Vector[in.LocalVar], in.Stmt), in.Expr]] = {
```
Felalolf, line 1597: To me it looks like `functionCallDAux` is already taking care of the case that the call has a spec. What is this method doing that `functionCallDAux` is not already doing? If it is not doing anything additionally, I would remove this method to avoid code duplication.

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1467,20 +1577,99 @@ object Desugar {
...
+        closure <- exprD(ctx, info)(expr.base.asInstanceOf[PExpression])
+        targets = func.result.outs map (o => freshExclusiveVar(typeD(func.context.symbType(o.typ), Addressability.exclusiveVariable)(src), expr, info)(src))
+      } yield Left((targets, in.ClosureCall(targets, closure, arguments(args zip params), spec)(src)))
+    }
+
+
+    /** Desugars the arguments to a function call or a call with spec.
+      * @param args The expressions passed as arguments
+      * @param params The types of the arguments of the callee (or the spec) */
+    private def functionCallArgsD(args: Vector[PExpression], params: Vector[Type])(ctx: FunctionContext, info: TypeInfo, src: Meta): Writer[Vector[in.Expr]] = {
```
Felalolf, line 1625: Here you moved code from the `funcationCallAux` method, right? This refactoring seems like a good ideal.

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -2729,18 +2983,116 @@ object Desugar {
...
             case e => Violation.violation(s"Expected a magic wand, but got $e")
           }
-        case PExplicitGhostStatement(actual) => stmtD(ctx)(actual)
+        case p: PClosureImplProof => closureImplProofD(ctx)(p)
+        case PExplicitGhostStatement(actual) => stmtD(ctx, info)(actual)
         case _ => ???
       }
     }
 
+    private def closureImplProofD(ctx: FunctionContext)(proof: PClosureImplProof): Writer[in.Stmt] = {
```
Felalolf, line 2992: Please, add more documentation of what is happening inside this method. On first glance, the body seems fine to me, but we need proper documentation so that it is maintainable for us.

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -2520,23 +2768,24 @@ object Desugar {
...
-    def varD(ctx: FunctionContext)(id: PIdnNode): in.Var = {
+    def varD(ctx: FunctionContext, info: TypeInfo)(id: PIdnNode): in.Expr = {
       require(info.regular(id).isInstanceOf[st.Variable])
-
-      ctx(id) match {
+      ctx(id, info) match {
         case Some(v : in.Var) => v
-        case Some(_) => violation("expected a variable")
-        case None => localVarContextFreeD(id)
+        case Some(d@in.Deref(v, _)) if v.isInstanceOf[in.Var] => d
```
Felalolf, line 2775:
```suggestion
        case Some(d@in.Deref(v: in.Var, _)) => d
```
?

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -2520,23 +2768,24 @@ object Desugar {
...
-    def assignableVarD(ctx: FunctionContext)(id: PIdnNode) : in.AssignableVar = {
+    def assignableVarD(ctx: FunctionContext, info: TypeInfo)(id: PIdnNode) : in.Expr = {
       require(info.regular(id).isInstanceOf[st.Variable])
 
-      ctx(id) match {
+      ctx(id, info) match {
         case Some(v: in.AssignableVar) => v
-        case Some(_) => violation("expected an assignable variable")
-        case None => localVarContextFreeD(id)
+        case Some(d@in.Deref(v, _)) if v.isInstanceOf[in.Var] => d
```
Felalolf, line 2786:
```suggestion
        case Some(d@in.Deref(v: in.Var, _)) => d
```
?

Felalolf:
I have now looked at everything. I have a few more comments, but nothing should be particularly big. After my comments are addressed, I think we can merge it and then do later changes in new pull requests. Some parts in the ghost checker seem scuffed, but I have the impression that the fault lies with how we have handled interface implementation proofs so far.

src/main/scala/viper/gobra/translator/encodings/closures/ClosureDomainEncoder.scala
```diff
@@ -0,0 +1,68 @@
...
+    if (domainNeeded) {
+      addMemberFn(vprDomain)
+    }
+  }
+
+  private var domainNeeded: Boolean = false
+  lazy val vprType: vpr.DomainType = {
+    domainNeeded = true
+    vpr.DomainType(Names.closureDomain, Map.empty)(Vector.empty)
+  }
```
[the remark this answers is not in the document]
Felalolf: There are some bits that require additional investigation. In particular, termination. Until these things are all cleared up, I would keep this warning.

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -1080,6 +1089,55 @@ case class StringLit(s: String)(val info: Source.Parser.Info) extends Lit {
 
 case class NilLit(typ: Type)(val info: Source.Parser.Info) extends Lit
 
+/* ** Closures */
+sealed trait FunctionLikeLit extends Lit with FunctionLikeMemberOrLit  {
```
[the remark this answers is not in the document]
stefanomil: Changed

src/main/scala/viper/gobra/frontend/Desugar.scala
stefanomil: Yes

src/main/scala/viper/gobra/frontend/Desugar.scala
stefanomil: Done

src/main/scala/viper/gobra/frontend/Desugar.scala
stefanomil: Yes

src/main/scala/viper/gobra/frontend/Desugar.scala
stefanomil: `closureCallDAux` is for closure calls, i.e. calls c(_) as spec where c is a closure expression (and not a function or method name). These need to be handled separately (the same way as function calls and method calls are handled separately within `functionCallDAux`). I have added a new method for this since there is not much in `functionCallDAux` that would be reused by this (except the code that I have factored out in `functionCallArgsD`)

stefanomil: I have added a line to the comment to make this clearer

src/main/scala/viper/gobra/translator/encodings/closures/ClosureDomainEncoder.scala
```diff
@@ -0,0 +1,68 @@
...
+    * for T=1..M
+    *   for X=1..N(T)
+    *     function captVarXClosure_T(closure: Closure): TypeT
+    */
+  private def capturedVarsDomainFuncs: Seq[vpr.DomainFunc] = {
+    specs.captVarsTypeMap.flatMap {
+      case (typ, (tid, num)) =>
+        (1 to num) map { i =>  vpr.DomainFunc(Names.closureCaptVarDomFunc(i, tid), Seq(vpr.LocalVarDecl(Names.closureArg, vprType)()), typ)(domainName = Names.closureDomain) }
+    }.toSeq
+  }
```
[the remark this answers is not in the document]
stefanomil: That is true, done. Funnily enough, serialising type names actually makes things simpler implementation-wise.

src/main/scala/viper/gobra/translator/encodings/closures/ClosureDomainEncoder.scala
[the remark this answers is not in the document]
stefanomil: Add suggested `Experimental feature` warning

stefanomil:
@Felalolf I have now addressed all of the comments

Felalolf (approved):

jcp19:
Looking at the tests, it looks like the warning that closures are an experimental feature is being printed on every test, even those that do not use the feature at all - this suggests that the closure encoding is always generated, regardless of whether it is used. Could we generate the closures' viper code only when we need them?

stefanomil:
@jcp19 @Felalolf The Closure domain is now only generated only if it is used

Issue #475, opened 2022-07-04
Implicit conversion of constants to pointer types


gavinleroy:
Consider the following Go program:
```go
package main

import "fmt"

type A struct {
    F int
}

type B struct {
    As []A
}

//@ preserves acc(a)
func (a *A) get_f() int {
    return a.F
}

//@ requires acc(b)
//@ requires acc(b.As)
//@ requires len(b.As) > 1
func (b *B) sum() int {
    first := b.As[0] // <--- NOTE Gobra requires &b.As[0]
    second := b.As[1]
    return first.get_f() + second.get_f()
}

//@ trusted
func main() {
    arr := make([]A, 2)
    arr[0] = A{F: 5}
    arr[1] = A{F: 10}
    b := &B{As: arr}
    fmt.Println("hello: ", b.sum())
}
```

The above is a valid Go program, and when run, will print `15`. This is because when calling `first.get_f()` first is implicitly cast to type `*A`. 
However, a verification attempt will result in `error: got unknown identifier get_f`. If the slice accesses are changed to read `&b.As[i]` then this will verify. It is often useful to fall back on this implicit conversion and I've seen it used in the SCION project.

PR #483, opened 2022-07-13
Fix 475

jcp19:
This PR addresses the complaints in issue #475. In particular, I identified some problems that make it hard to understand why Gobra rejects the example in issue 475, even though it is correct to reject it:
- the error message is bad - it says that the identifier of the method is unknown even though a method with that name is clearly there. This has to do with the fact that Gobra performs method lookup differently when the receiver is addressable and when it is not. Here, it would be expected of Gobra to complain that the receiver in the example is not shared.
- once this error is understood, the user might try to make the expression addressable with `share r` - this is problematic because `r` will not be shared when `r` is a receiver - instead, receivers should be marked with a `@`. This might be confusing because Gobra does not complain about sharing a receiver. This PR introduces support for sharing a receiver, as well as allowing it to be marked as shared in its declaration.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
@@ -288,16 +288,16 @@ trait MemberResolution { this: TypeInfoImpl =>
           tryFieldLookup(exprType(expr), id)
         } else if (isEffAddressable && addr.nonEmpty) {
           addr
+        } else if (!isEffAddressable && nonAddr.nonEmpty) {
+          nonAddr
         } else if (isGoEffAddressable && addr.nonEmpty && nonAddr.isEmpty) {
           val errEntity = ErrorMsgEntity(error(id, s"$id requires a shared receiver ('share' or '@' annotations might be missing)."))
           Some((errEntity, Vector()))
-        } else if (isEffAddressable && addr.isEmpty && nonAddr.nonEmpty) {
-          val errEntity = ErrorMsgEntity(error(id, s"$id expects a non-effectively addressable receiver, but got $expr instead"))
-          Some((errEntity, Vector()))
-        } else if (!isEffAddressable && nonAddr.nonEmpty) {
-          nonAddr
         } else if (!isEffAddressable && addr.nonEmpty) {
           val errEntity = ErrorMsgEntity(error(id, s"$id requires the receiver to be effectively addressable, but got $expr instead"))
+          Some((errEntity, Vector()))
+        } else if (isEffAddressable && addr.isEmpty && nonAddr.nonEmpty) {
+          val errEntity = ErrorMsgEntity(error(id, s"$id expects a non-effectively addressable receiver, but got $expr instead"))
           Some((errEntity, Vector()))
         } else {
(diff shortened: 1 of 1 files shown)


src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
```diff
@@ -279,9 +279,18 @@ trait MemberResolution { this: TypeInfoImpl =>
...
     exprOrType(b) match {
       case Left(expr) =>
-        val methodLikeAttempt = tryMethodLikeLookup(expr, id)
-        if (methodLikeAttempt.isDefined) methodLikeAttempt
-        else tryFieldLookup(exprType(expr), id)
+        val methodLikeAttempts = tryMethodLikeLookup(expr, id)
+        methodLikeAttempts match {
+          case (Some(v), _) if effAddressable(expr) => Some(v)
+          case (_, Some(v)) if !effAddressable(expr) => Some(v)
+          case (Some(_), _) =>
```
Felalolf, line 286:
```suggestion
          case (Some(_), _) if goEffAddressable(expr) =>
```

src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
```diff
@@ -279,10 +279,29 @@ trait MemberResolution { this: TypeInfoImpl =>
...
+          val errEntity = ErrorMsgEntity(error(id, s"$id expects a non-effectively addressable receiver, but got $expr instead"))
+          Some((errEntity, Vector()))
+        } else if (!isEffAddressable && nonAddr.nonEmpty) {
+          nonAddr
+        } else if (!isEffAddressable && addr.nonEmpty) {
+          val errEntity = ErrorMsgEntity(error(id, s"$id requires the receiver to be effectively addressable, but got $expr instead"))
+          Some((errEntity, Vector()))
+        } else {
+          Violation.violation(s"unexpected case reached: $expr")
+        }
```
Felalolf, line 304: I would rearrange the cases a bit.
```suggestion
        if (addr.isEmpty && nonAddr.isEmpty) {
          // could not find the corresponding member
          tryFieldLookup(exprType(expr), id)
        } else if (isEffAddressable && addr.nonEmpty) {
          addr
        } else if (!isEffAddressable && nonAddr.nonEmpty) {
          nonAddr
        } else if (isGoEffAddressable && addr.nonEmpty && nonAddr.isEmpty) {
          val errEntity = ErrorMsgEntity(error(id, s"$id requires a shared receiver ('share' or '@' annotations might be missing)."))
          Some((errEntity, Vector()))
        } else if (!isEffAddressable && addr.nonEmpty) {
          val errEntity = ErrorMsgEntity(error(id, s"$id requires the receiver to be effectively addressable, but got $expr instead"))
          Some((errEntity, Vector()))
        } else if (isEffAddressable && addr.isEmpty && nonAddr.nonEmpty) {
          val errEntity = ErrorMsgEntity(error(id, s"$id expects a non-effectively addressable receiver, but got $expr instead"))
          Some((errEntity, Vector()))
        } else {
          Violation.violation(s"unexpected case reached: $expr")
        }
```

Felalolf (approved):
one small remark

PR #488, opened 2022-07-18
Fix issue 486 (V2)

Felalolf:
Title


jcp19 (approved):
LGTM!
we can merge after the implementation goes through

PR #496, opened 2022-07-28
Range

Dspil:
This PR adds range for slices and arrays.
Slices and arrays have the same behavior regarding range. The general transformation to support that is the following:
Given a simple range loop in gobra:

```
<invariants>
for i, j := range x {
	<body>
}
```
It is transformed into:

```
var i int
var j <type of elements of x>
<invariants>
invariant 0 <= i && i <= len(x)
invariant 0 <= i && i < len(x) ==> j == x[i]
while i < len(x) {
    <body>
    i += 1
    if (i < len(x)) j = x[i]
}
```

The extra invariants are added after the user provided invariants as the second one refers to the expression in range and the user has to have provided an invariant granting access to that before.

In the case where we have a normal assignment `for i, j = range x` and not a short one, the transformation is the same, minus the variable declarations. In this case i and j could be any lvalues.


src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1126,6 +1126,174 @@ object Desugar {
...
+              indexLeft <- leftOfAssignmentD(shorts(0))(in.IntT(Addressability.exclusiveVariable))
+              indexVar = in.Assignee.Var(indexLeft)
+              indexSrc = meta(shorts(0))
+              rangeSrc = meta(range)
+              indexAss = singleAss(indexVar, in.IntLit(0)(rangeSrc))(indexSrc)
+              incrIndex = singleAss(indexVar, in.Add(indexLeft, in.IntLit(1)(indexSrc))(indexSrc))(indexSrc)
+
+              rangeExpSrc = meta(range.exp)
+
+              // in go the range expression is only computed once before the iteration begins
```
jcp19, line 1151: In Go, the expression inside range (let's call it `e`) is not necessarily evaluated if `len(e)` is constant (does not contain function calls and channel receives), which is probably meant as an optimization. I think that your solution is fine as long as in Gobra we guarantee that if len(src) is constant then src does not have side effects. Is this the case?

Furthermore, your solution is stricter then what Go allows. E.g., in Go, the following is totally fine:
```go
func main() {
	var x *[3][3]int
	fmt.Println(len(x[0]))
}
```
But your solution rejects it because it forces the evaluation of `x[0]` and `x` is `nil` here. Nonetheless, I am ok with this, as it looks to me that it is bad style to have these "ill-formed" expressions as parts of the program.

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1126,6 +1126,174 @@ object Desugar {
...
+              // we do that by storing it in copiedVar
```
jcp19, line 1152: I would enrich this comment to address at least the fact that the elements you iterate through do not change, even if you reassign to the slice or modify it by appending to it for example

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1126,6 +1126,174 @@ object Desugar {
...
+              // get the length of the expression
+              length = in.Length(copiedVar)(rangeSrc)
+              cond = in.LessCmp(indexLeft, length)(rangeSrc)
+
+              // this invariant states that the index is between 0 and the length of the array/slice (both inclusive)
+              indexBoundsInv = in.ExprAssertion(in.And(
+                in.AtMostCmp(in.IntLit(0)(rangeSrc), indexLeft)(rangeSrc),
+                in.AtMostCmp(indexLeft, length)(rangeSrc))(rangeSrc))(rangeSrc)
+
+              // standard for loop desugaring
```
jcp19, line 1165: how much does this code overlap with the code for standard loops? Would it be possible to factor out their shared functionality in a method?

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1126,6 +1126,174 @@ object Desugar {
...
+                  Vector(copyAss, indexAss) ++ dInvPre ++ dTerPre ++ Vector(
+                    in.While(cond, dInv ++ Vector(indexBoundsInv), dTer, in.Block(Vector(continueLoopLabelProxy),
+                      Vector(dBody, continueLoopLabel, incrIndex) ++ dInvPre ++ dTerPre
+                    )(src))(src), breakLoopLabel
+                  )
+                )(src)
+              }
+            } yield wh))
+
+          case n@PAssForRange(range, ass, spec, body) =>
```
jcp19, line 1219: a similar comment as before applies here - how much of this logic is repeated? would it make sense to refactor this to avoid repetition?

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1126,6 +1126,174 @@ object Desugar {
...
+                in.Block(
+                  Vector(valueLeft.asInstanceOf[in.LocalVar]),
+                  Vector(copyAss, indexAss, valueAss) ++ dInvPre ++ dTerPre ++ Vector(
+                    in.While(cond, dInv ++ Vector(indexBoundsInv, indexValueEq), dTer, in.Block(Vector(continueLoopLabelProxy),
+                      Vector(dBody, continueLoopLabel, incrIndex, updateValue) ++ dInvPre ++ dTerPre
+                    )(src))(src), breakLoopLabel
+                  )
+                )(src)
+              }
+              else {
```
jcp19, line 1206: looks awkward, usually you see `} else {` in the same line

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1126,6 +1126,174 @@ object Desugar {
...
+              continueLabelName = nm.continueLabel(n, info)
+              continueLoopLabelProxy = in.LabelProxy(continueLabelName)(src)
+              continueLoopLabel = in.Label(continueLoopLabelProxy)(src)
+
+              breakLabelName = nm.breakLabel(n, info)
+              breakLoopLabelProxy = in.LabelProxy(breakLabelName)(src)
+              _ <- declare(breakLoopLabelProxy)
+              breakLoopLabel = in.Label(breakLoopLabelProxy)(src)
+
+              wh = if (shorts.length == 2) {
```
jcp19, line 1180: might not be immediately obvious here why we do this case analysis. I would add a very small skeleton of a loop where this case is supposed to apply (in comments) and I would do the same for the `else` case

src/main/scala/viper/gobra/frontend/Desugar.scala
jcp19: Additionally, here you rely on the implicit assumption that the length of `shorts` is either 1 or 2. If for some reason, that assumption is violated (e.g., len(shorts) = 3), the code will still pass instead of reporting what is obviously an erroneous situtaion. Could we add a third case that reports a violation here?

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -3329,10 +3497,10 @@ object Desugar {
     }
 
     /** returns the relativeId with the CONTINUE_LABEL_SUFFIX appended */
-    def continueLabel(loop: PForStmt, info: TypeInfo) : String = relativeId(loop, info) + CONTINUE_LABEL_SUFFIX
+    def continueLabel(loop: PNode, info: TypeInfo) : String = relativeId(loop, info) + CONTINUE_LABEL_SUFFIX
```
jcp19, line 3500: why did you make the type of loop so much weaker? is it to be able to deal with range and normal loops? If so, wouldn't it make more sense to have a trait that both types of loops implement, and require this param to be of that trait type?

src/main/scala/viper/gobra/frontend/info/implementation/typing/MiscTyping.scala
```diff
@@ -51,12 +51,12 @@ trait MiscTyping extends BaseTyping { this: TypeInfoImpl =>
...
-      case SliceT(elem) => InternalSingleMulti(elem, InternalTupleT(Vector(IntT(config.typeBounds.Int), elem)))
-      case GhostSliceT(elem) => InternalSingleMulti(elem, InternalTupleT(Vector(IntT(config.typeBounds.Int), elem)))
-      case MapT(key, elem) => InternalSingleMulti(key, InternalTupleT(Vector(key, elem)))
-      case ChannelT(elem, ChannelModus.Recv | ChannelModus.Bi) => elem
+      case ArrayT(_, elem) => InternalTupleT(Vector(IntT(config.typeBounds.Int), elem))
+      case PointerT(ArrayT(_, elem)) => InternalTupleT(Vector(IntT(config.typeBounds.Int), elem))
+      case SliceT(elem) => InternalTupleT(Vector(IntT(config.typeBounds.Int), elem))
+      case GhostSliceT(elem) => InternalTupleT(Vector(IntT(config.typeBounds.Int), elem))
+      case MapT(key, elem) => InternalTupleT(Vector(key, elem))
+      case ChannelT(elem, ChannelModus.Recv | ChannelModus.Bi) => InternalTupleT(Vector(elem))
```
jcp19, line 59: why did you change these types? I think this will just make your life harder

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
```diff
@@ -103,14 +104,32 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
...
-
-    case n@PAssForRange(exp, lefts, _) =>
-      multiAssignableTo.errors(Vector(miscType(exp)), lefts map exprType)(n) ++
-        lefts.flatMap(t => addressable.errors(t)(t))
+    case _@PShortForRange(range, shorts, _, _) =>
+      if (shorts.forall(pointsToData)) {
+        val typ = miscType(range).asInstanceOf[Type.InternalTupleT]
+        if (shorts.length > 0 && shorts.length <= typ.ts.length) {
+          if (typ.ts.take(shorts.length).zip(shorts map idType).forall(assignableTo))
+            noMessages
```
jcp19, line 112: couldn't you use `multiAssignableTo` here?

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
```diff
@@ -103,14 +104,32 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
...
+          else error(range, "Can not assign range types to identifiers")
+        }
+        else error(range.exp, "Wrong number of identifiers for range")
+      }
+      else error(range.exp, s"at least one assignee in $shorts points to a type")
```
jcp19, line 117: what is this supposed to mean?

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
jcp19: also, instead of returning all these errors explicitly, you could probably just return the errors found in multiAssignableTo (usually, the idiom that you see in code is `multiAssignableTo.errors(???, ???)(???)`)

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
```diff
@@ -103,14 +104,32 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
...
+      })
+      val typ = miscType(range).asInstanceOf[Type.InternalTupleT]
+      if (ass.length > 0 && ass.length <= typ.ts.length) {
+        if (typ.ts.take(ass.length).zip(typs).forall(assignableTo))
+          noMessages
+        else error(range, "Can not assign range types to identifiers")
+      }
+      else error(range.exp, "Wrong number of identifiers for range") ++
+      ass.flatMap(a => assignable.errors(a)(a))
+    }
```
jcp19, line 132: same here

src/main/scala/viper/gobra/frontend/Desugar.scala
Dspil: it is about 10 lines of code consisting of handling the break label, the continue label and the invariant and termination measure desugaring. I didn't extract this into a separate method because it seemed that it does many distinct and not complete stuff (since it does not generate the viper while) and it is not that big of a chunk of code.

src/main/scala/viper/gobra/frontend/Desugar.scala
Dspil: The logic is the same, but now instead of a vector of PIdnUnk objects that I have to create variables from I have a bunch of lvalues so the way they are handled is different. I am not aware of a smart way to handle them identically.

src/main/scala/viper/gobra/frontend/Desugar.scala
Dspil: I didn't include it because I have handled that case in the type checker. Is there a situation where I shouldn't depend on the type checker to catch that case?

src/main/scala/viper/gobra/frontend/Desugar.scala
Dspil: Fixed for the functions continueLabel and breakLabel, left it as is for relativeId since it is used by outline too and it seems like something generally useful

src/main/scala/viper/gobra/frontend/info/implementation/typing/MiscTyping.scala
Dspil: Initially I tried to use multiAssignableTo but in the case where I have one identifier, that tried to assign the whole tuple returned by the range expression. I just wanted to easily extract as many elements as I want to account easily for that case.

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
Dspil: It was there before I started altering the type checking and I thought it was needed, after further inspection it is not, removed.

src/main/scala/viper/gobra/reporting/VerifierError.scala
```diff
@@ -266,6 +266,11 @@ case class ChannelMakePreconditionError(info: Source.Verifier.Info) extends Veri
   override def localMessage: String = s"The provided length to ${info.origin.tag.trim} might be negative"
 }
 
+case class RangeVariableMightNotExistError(info: Source.Verifier.Info)(varName: String, rangeExpr: String) extends VerificationError {
+  override def localId: String = "range_variable_might_not_exist"
+  override def localMessage: String = s"Range variable '$varName' might not exist if the length of expression '$rangeExpr' is 0"
```
jcp19, line 271:
```suggestion
  override def localMessage: String = s"Range variable '$varName' might not have been declared because the length of the expression '$rangeExpr' might be 0"
```

src/test/resources/regressions/features/loops/range-fail2.gobra
```diff
@@ -0,0 +1,11 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package pkg
+
+func foo_arr() {
+	x := []uint{1, 2, 3, 4}
+	//:: ExpectedOutput(invariant_not_well_formed)
```
jcp19, line 8: Why is the invariant not well formed?

jcp19 (approved):
Apart for the small comments we discussed in person, LGTM! I am in favor of merging it but I would like the opinion of @Felalolf or @ArquintL before doing so - in particular, I wonder whether we should just require the user to write the invariants by hand at all the times instead of trying to generate the obvious ones. The disadvantage of doing so is that (1) we have more code, (2) it might make error messages a bit obscure as shown in test case src/test/resources/regressions/features/loops/range-fail2.gobra

src/test/resources/regressions/features/loops/range-fail2.gobra
Dspil: One of the invariants that is added automatically is:
len(x) > 0 ==> 0 <= i && i < len(x) ==> j == x[i]
which fails because we don't have acc(x).
The user wouldn't be able anyway to do a range over this loop because inside the loop body, an assignment from x[i] would happen to j. The error message is not super clarifying though.

Dspil: Error message improved

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
```diff
@@ -217,8 +217,12 @@ trait TypeEncoding extends Generator {
   final def invariant(ctx: Context): in.Assertion ==> (CodeWriter[Unit], vpr.Exp) = {
     def invErr(inv: vpr.Exp): ErrorTransformer = {
       case e@ vprerr.ContractNotWellformed(Source(info), reason, _) if e causedBy inv =>
-        LoopInvariantNotWellFormedError(info)
-          .dueTo(DefaultErrorBackTranslator.defaultTranslate(reason))
+        info.origin match {
+          case Source.AnnotatedOrigin(_, _:Source.NoPermissionToRangeExpressionAnnotation) =>
+            NoPermissionToRangeExpressionError(info).dueTo(DefaultErrorBackTranslator.defaultTranslate(reason))
```
Felalolf, line 222: Can this not be done where the other annotations are handled?

src/main/scala/viper/gobra/reporting/VerifierError.scala
```diff
@@ -266,6 +266,16 @@ case class ChannelMakePreconditionError(info: Source.Verifier.Info) extends Veri
...
 
+case class RangeVariableMightNotExistError(info: Source.Verifier.Info)(varName: String, rangeExpr: String) extends VerificationError {
+  override def localId: String = "range_variable_might_not_exist"
+  override def localMessage: String = s"Range variable '$varName' might not have been declared because the length of the expression '$rangeExpr' might be 0"
+}
+
+case class NoPermissionToRangeExpressionError(info: Source.Verifier.Info) extends VerificationError {
+  override def localId: String = "no_permission_to_range_expression"
+  override def localMessage: String = s"Generated invariant for the loop with range clause cannot hold. Missing permissions for range expression"
+}
```
Felalolf, line 277: I find both error messages a bit confusing. I would go with something along the lines of "Might not have read permissions to range expression." since the user does not have to know that the generated invariant causes this, and "The length of the range expression might be 0." since the current statement about the existence of ranged variables seems missleading.

Felalolf:
I do not like that the transformation is handled by the desugarer, but since it is urgent, we can go with this solution until we might change it.

Felalolf (approved):

PR #501, opened 2022-08-06
Make closure calls deferrable

stefanomil:


Felalolf (approved):

src/main/scala/viper/gobra/translator/util/Core.scala
```diff
@@ -62,11 +63,19 @@ object Core {
     } yield x.copy(args = args)(x.info)
   }
 
+  def deepCore(x: in.ClosureSpec): Core[in.ClosureSpec] = {
+    for {
+      paramValues <- vector(x.paramValues.map(y => option(y.map((core)))))
```
jcp19, line 68: nitpick: duplicated pair of parentheses around `core`

jcp19 (approved):
LGTM

src/main/scala/viper/gobra/translator/util/Core.scala
stefanomil: Parentheses removed

PR #514, opened 2022-08-30
Introduce ghost equality (`===`) and inequality (`!==`)

jcp19:
This PR introduces support for the operators `===` and `!==` which can be used in ghost contexts to perform comparisons between slices and between non-comparable interfaces.

The big size of this PR is due to the changes in the parser.


Aurel300:
Interesting syntax you've got there :eyes:

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -2263,6 +2263,24 @@ object Desugar {
...
             }
 
+          case PGhostEquals(left, right) =>
+            if (info.typ(left) == PermissionT || info.typ(right) == PermissionT) {
+              // TODO: inherits the same check as for PEquals and PUnequals. Could possibly be dropped or
+              //       refactored in the future
+              for { l <- permissionD(ctx, info)(left); r <- permissionD(ctx, info)(right) } yield in.GhostEqCmp(l, r)(src)
+            } else {
+              for { l <- exprD(ctx, info)(left); r <- exprD(ctx, info)(right) } yield in.GhostEqCmp(l, r)(src)
+            }
```
Felalolf, line 2273: I do not understand why the extra permissions cse is necessary since they should also just be expressions. Does something break, when you remove the case distinction?

Felalolf (approved):
One question maybe needs to be addressed.

src/main/scala/viper/gobra/frontend/Desugar.scala
jcp19: This guarantees that, when you are comparing a perm expression to some literal, both operands are treated as permissions. Otherwise, encoding the following comparison would fail if the right hand side is treated as an int (integer division is used here):
```
func test() {
    assert perm(1/2) == 1/2
}
```
Another technical point is that this expression would be encoded as a comparison between an integer and a perm, for which we do not implement a case (although we could).

Felalolf: Please add a comment explaining the first part "This guarantees that, when you are comparing a perm expression to some literal, both operands are treated as permissions. E.g. the right-hand side of perm(1/2) == 1/2 is treated as a permission.

To me this seems like bad language design on our part, but we can discuss this in a future meeting and decide on a solution then.

PR #499, opened 2022-08-05
Update Gobra to the newest Viper version

jcp19:
This PR updates Gobra to use the latest changes as of today (01.09.2022). Initially, it was meant to update Gobra to use the viper version from the `22.07` release, but instead it uses a more recent version than that due to nice-to-have changes made recently (in particular, https://github.com/viperproject/silver/pull/604).

Furthermore, in this PR, I removed `ViperChopper` - instead, we now use `Chopper` provided by silver. Furthermore, I changed Gobra to use submodules, which makes the process of upgrading and installing Gobra a lot easier! Thanks @ArquintL  for changing viperserver to use modules.

PR #515, opened 2022-08-31
Rename 'ReceiverNotInjective' -> 'QPAssertionNotInjective'

gavinleroy:
In the most recent version of Silver `viper.silver.verifier.reasons.ReceiverNotInjective` was renamed to `QPAssertionNotInjective`. I'm not sure if this is worth merging into master just yet for compatibility.


jcp19:
This does not work without upgrading the version of viper that we are using. This will be done in PR #499 soon

gavinleroy:
Whoops, I did not see that PR. I'll close this PR, there is no sense in having it merged separately.

PR #522, opened 2022-09-19
Remove redundant checks when comparing interfaces with the `nil` literal

jcp19:
It is very common to compare an interface value with the literal `nil`. Currently, this comparison is encoded as an equality, preceded by a check that one of the operands is comparable. This check is redundant and I removed it in this PR.


ArquintL:
I would generalize your optimization such that it does not matter whether `nil` is on the rhs or lhs

jcp19:
> I would generalize your optimization such that it does not matter whether `nil` is on the rhs or lhs

I don't know how I forgot this xD thanks for reminding me!

Felalolf (approved):

PR #525, opened 2022-09-21
Verify chopped parts in parallel

jcp19:
Currently, chopping a program in multiple parts causes all parts to be verified sequentially. This PR enables concurrent verification of those parts.

### DIFF
--- a/src/main/scala/viper/gobra/backend/BackendVerifier.scala
+++ b/src/main/scala/viper/gobra/backend/BackendVerifier.scala
@@ -59,25 +59,20 @@ object BackendVerifier {
         val num = programs.size
         var counter = 0 // verification progress counter
 
-        // TODO: check if this is safe for carbon too
-        // TODO: doc
+        // Starts verifying all chopped programs in parallel
         val partialVerificationResults = Future.traverse(programs.zipWithIndex) { case (program, idx) =>
           val programID = s"${config.taskName}_$idx"
           verifier.verify(programID, reporter, program)(executor).andThen { _ =>
-            // TODO: explain purpose
-            this.synchronized {
-              counter += 1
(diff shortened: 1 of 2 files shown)


src/main/scala/viper/gobra/backend/BackendVerifier.scala
```diff
@@ -57,23 +57,22 @@ object BackendVerifier {
...
         val programs = ChopperUtil.computeChoppedPrograms(task, pkgInfo)(config)
         val num = programs.size
+        var counter = 0 // verification progress counter
+
+        // Starts verifying all chopped programs in parallel
+        val partialVerificationResults = Future.traverse(programs.zipWithIndex) { case (program, idx) =>
+          val programID = s"${config.taskName}_$idx"
+          verifier.verify(programID, reporter, program)(executor).andThen { _ =>
+            // this block ensures that progress messages are printed in order
+            this.synchronized { counter += 1; config.reporter report ChoppedProgressMessage(counter, num) }
```
Felalolf, line 67: This way you do not know which chopped program has completed, right?
I think it would be usefull to know if some chopped parts are very quick or very slow. Could you integrate the idx into the ChoppedProgressMessage?

Felalolf (approved):

ArquintL:
Are concurrent verifications now possible? I thought @marcoeilers "only" added that branches within a method are verified concurrently

marcoeilers:
@ArquintL Yes, concurrent use of several verifier instances is possible now, I did both.

src/main/scala/viper/gobra/backend/BackendVerifier.scala
jcp19: should be done now

PR #529, opened 2022-09-29
Remove node and strategy

Felalolf:
I simplified the Nodes.subnodes implementation so that we do not have to extend it for every new internal node.
Similarly, I added a general solution for the GobraStrategy.


src/main/scala/viper/gobra/ast/internal/Node.scala
```diff
@@ -115,23 +112,42 @@ trait Node extends Rewritable with Product {
     }
   }
 
-}
+  def transform(pre: PartialFunction[Node, Node] = PartialFunction.empty,
+                recurse: Traverse = Traverse.Innermost)
+  : this.type = {
```
jcp19, line 118: Why are we using `this.type` here? If this is guaranteed to return `this` and we are modifying `this` directly, why not make this of type `Unit`?

Dspil (approved):

jcp19 (approved):

src/main/scala/viper/gobra/ast/internal/Node.scala
Felalolf: This is from the Viper code. We often return this instead of making the function have no results since it fits better with the scala programming style. But this is my best bet.

PR #531, opened 2022-09-29
Fix issue 491

### LINKED ISSUE #491: Dereferencing feature
Gobra accepts the following 2 examples:

```
type A struct {
	x int
}

func main() {
	var x *A
	var y *A = &(*x)
	assert y == x
}
```

and 

```
func main() {
	var x *[3]int
	var y []int = (*x)[:]
}
```

Felalolf:
This PR: 

- Fixes a bug in the encoding context, where using the copy constructor would cause the freshname generator to break.
- Makes the length argument of array type expressions constant. This makes our treatment of types more consistent.
- Fixes an unsoundness with length.
- Fixes an unsoundness where panics that are caused by referencing L-values were not detected. Instead of checking that every L-value is safe immediately, we check that all usages of L-values are safe. See `safeReference` for more details.
- Fixes several unsoundnesses in the encoding of zero sized types.
- Adds an `UncheckedRef` node. The node is necessary to make our encoding modular.
- Renames `InterfaceReceiverIsNilReason` to the more general `ReceiverIsNilReason`. 
- Fixes some error codes. Previously, assignments and dereferencs sometimes failed with exhale and call error, respectively, and now fail with assignment and deref error, respectively.

There is one failing test. The test fails due to an error in the desugaring of closures. I will most likely add an `IgnoreFile` annotation and open a separate issue. The currently generated internal representation should fail since it contains an unsafe dereference.


src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
```diff
@@ -121,7 +122,7 @@ trait TypeEncoding extends Generator {
       for {
         footprint <- addressFootprint(ctx)(loc, in.FullPerm(loc.info))
         eq1 <- ctx.equal(loc, in.DfltVal(t.withAddressability(Exclusive))(loc.info))(loc)
-        eq2 <- ctx.equal(in.Ref(loc)(loc.info), in.NilLit(in.PointerT(t, Exclusive))(loc.info))(loc)
+        eq2 <- ctx.equal(in.UncheckedRef(loc)(loc.info), in.NilLit(in.PointerT(t, Exclusive))(loc.info))(loc)
```
jcp19, line 125: Can you elaborate on why you picked `UncheckedRef` here instead of `Ref`?

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
```diff
@@ -403,3 +433,52 @@ trait TypeEncoding extends Generator {
...
+    else {
+      ctx.expression(in.UneqCmp(
+        in.UncheckedRef(loc)(annotatedInfo),
+        in.NilLit(in.PointerT(loc.typ, Exclusive))(annotatedInfo)
+      )(annotatedInfo))
+    }
+  }
+
+  @tailrec
+  private def cannotBeNil(l: in.Expr): Boolean = l match {
```
jcp19, line 478: Is this meant to only check locations? If so, I would probably make the param type more restrictive (i.e., `in.Location`)

jcp19 (approved):

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
Felalolf: No, it also checks expressions (in particular, see the body of the method, where we use it to check whether the receiver of an index and field-ref expression cannot be nil).

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
Felalolf: The way we model memory, taking a reference never fails (as it is a pure operation). Conversely, taking a reference in Go fails if the argument is nil. This mismatch caused the unsoundness. To fix the mismatch, I added the (Safe)Ref node, which artificially fails if the argument is nil. Whenever there is an actual reference-taking happening in the Go program, we will have to use `Ref`. However, in many cases, such as here, we take a reference in the encoding even though there is no reference-taking in Go. In these cases, we can use `UncheckedRef`. In such cases, we use references to modularly split the encoding.

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
Felalolf: I could change it to something along the lines of:
```
private def cannotBeNil(l: in.Location): Boolean = {
  def aux(e: in.Expr): Boolean = {
    // body of cannotBeNil
  }
  aux(l)
}
```

Felalolf:
This PR still has issues because of which it is not merged

jcp19:
> This PR still has issues because of which it is not merged

For reference, the issue mentioned here is that Gobra is often not able to prove non-nilness of memory locations that are obviously non-nil. Consider the simple example below
```go
requires  8 <= len(raw)
preserves forall i int :: { &raw[i] } 0 <= i && i < len(raw) ==>
	acc(&raw[i])
func DecodeFromBytes(raw []byte) {
	assert forall i int :: 0 <= i && i < len(raw[2:4]) ==>
		&raw[2:4][i] == &raw[2 + i]
}
```
In the current status of the PR, we get the following error:
```go
[info] Error at: </Users/joao/000491-bug.gobra:7:58> Reading might fail. 
[info] The receiver raw[2:4][i] might be nil
```

ArquintL:
The handling of zero-sized data has been extracted and merged via #1022 and #1023. This PR just got updated by merging in master to (resolve its conflicts &) reduce its diff

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
```diff
@@ -10,18 +10,20 @@ import org.bitbucket.inkytonik.kiama.==>
 import viper.gobra.ast.{internal => in}
 import viper.gobra.ast.internal.theory.Comparability
 import viper.gobra.reporting.BackTranslator.{ErrorTransformer, RichErrorMessage}
-import viper.gobra.reporting.{AssignmentError, DefaultErrorBackTranslator, LoopInvariantNotWellFormedError, MethodContractNotWellFormedError, NoPermissionToRangeExpressionError, Source}
+//  import viper.gobra.reporting.{AssignmentError, DefaultErrorBackTranslator, DerefError, LoopInvariantNotWellFormedError, MethodContractNotWellFormedError, NoPermissionToRangeExpressionError, Source}
```
ArquintL:
```suggestion
```

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
```diff
@@ -471,3 +500,52 @@ trait TypeEncoding extends Generator {
...
+    }
+  }
+
+  @tailrec
+  private def cannotBeNil(l: in.Expr): Boolean = l match {
+    case _: in.Var => true
+    case l: in.FieldRef => cannotBeNil(l.recv)
+    case l: in.IndexedExp => cannotBeNil(l.base)
+    case _ => false
+  }
```
ArquintL: Implementing the suggestion that emerged from the discussion above:

```suggestion
  private def cannotBeNil(l: in.Location): Boolean = {
    @tailrec
    def aux(e: in.Expr): Boolean = e match {
      case _: in.Var => true
      case l: in.FieldRef => aux(l.recv)
      case l: in.IndexedExp => aux(l.base)
      case _ => false
    }
    aux(l)
  }
```

src/main/scala/viper/gobra/translator/encodings/structs/StructEncoding.scala
```diff
@@ -186,7 +186,7 @@ class StructEncoding extends TypeEncoding {
     * R[ (base: Struct{F})[f = e] ] -> ex_struct_upd([base], f, [e], F)
     * R[ dflt(Struct{F}) ] -> create_ex_struct( [T] | (f: T) in F )
     * R[ structLit(E) ] -> create_ex_struct( [e] | e in E )
-    * R[ loc: Struct{F}@ ] -> convert_to_exclusive( Ref[loc] )
+    * R[ loc: Struct{F}@ ] -> convert_to_exclusive( Ref[loc] ) // assert [&loc != nil] if Struct{F} has size zero
```
ArquintL: I find the comment a bit confusing as the body does not implement any nilness check

```suggestion
    * R[ loc: Struct{F}@ if size != 0] -> convert_to_exclusive( Ref[loc] ) // zero-sized structs are handled by the default implementation
```

ArquintL (approved):

jcp19:
@ArquintL feedback incorporated. Can be merged together with #1077, otherwise, it leads to very bad incompletenesses.

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
ArquintL: This was a wrong suggestion, sorry!

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
```diff
@@ -471,3 +499,55 @@ trait TypeEncoding extends Generator {
...
+  private def cannotBeNil(l: in.Location): Boolean = {
+    @tailrec
+    def aux(e: in.Expr): Boolean = e match {
+      case _: in.Var => true
+      case l: in.FieldRef => cannotBeNil(l.recv)
+      case l: in.IndexedExp => cannotBeNil(l.base)
+      case _ => false
+    }
+    aux(l)
+  }
```
ArquintL, line 552:
```suggestion
  private def cannotBeNil(l: in.Location): Boolean = {
    @tailrec
    def aux(e: in.Expr): Boolean = e match {
      case _: in.Var => true
      case l: in.FieldRef => aux(l.recv)
      case l: in.IndexedExp => aux(l.base)
      case _ => false
    }
    aux(l)
  }
```

PR #535, opened 2022-09-30
Use let expressions in encoding to reduce redundancy

Felalolf:
This PR is a first attempt to make use of let expressions in the encoding more often.
The main contribution of this PR are the value function of the type encoding and the two bind functions.

Currently, it is unclear at which points we should transform an expression writer to an expression, and thus "discharge" the accumulated let expressions. For now, my idea was to use pure whenever one of the new bind methods is used in the encoding of expressions. An alternative is, similar to statements, to use pure at finalExpression in the type encoding.

### DIFF
--- a/src/main/scala/viper/gobra/translator/util/ViperUtil.scala
+++ b/src/main/scala/viper/gobra/translator/util/ViperUtil.scala
@@ -48,7 +48,7 @@ object ViperUtil {
   }
 
   /**
-    * TODO: should be removed once the corresponding silver function is fixed.
+    * TODO: should be removed once the corresponding silver function is fixed, see /silver/issues/610.
     *
     * Returns a list of all undeclared local variables used in this statement.
     * If the same local variable is used with different


src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
```diff
@@ -272,6 +272,22 @@ trait TypeEncoding extends Generator {
...
+    * Value[ loc: T@ ] = R[loc]
+    *
+    * */
+  final def value(ctx: Context): in.Expr ==> CodeWriter[vpr.Exp] = {
+    val liftedResult: in.Expr => Option[CodeWriter[vpr.Exp]] = {
+      case (l: in.Location) :: _ / Shared => reference(ctx).lift(l)
+      case e => finalExpression(ctx).lift(e)
+    }
+    liftedResult.unlift
+  }
```
jcp19, line 289: why are we applying `lift` and then `unlift`?

src/main/scala/viper/gobra/translator/util/ViperUtil.scala
```diff
@@ -46,4 +46,53 @@ object ViperUtil {
       case _ => Violation.violation(s"expected vpr variable or field access, but got $left")
     }
   }
+
+  /**
+    * TODO: should be removed once the corresponding silver function is fixed.
+    *
```
jcp19, line 52: Can you please add the link to the `silver` issue that pertains to this comment? Otherwise, it is hard to know what the issue is and to understand if it has been fixed.

jcp19 (approved):
LGTM, I just have two minor comments. Also, from my experience, this reduced the size (in bytes) of the viper output for a large package in SCION in ~4%, but did not lead to any observable performance improvements. Nonetheless, the changes here look reasonable to me and I would be happy to merge it.

ArquintL (approved):
I'm wondering about the same two things but otherwise looks good

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
Felalolf: Writing `liftedResult` as a partial function from `in.Expr` to `CodeWriter[vpr.Exp]` is messy (For instance, I would have to match on `reference(ctx)` and `in.Location` in one expression which requires first widening the `reference` function etc). The key insight is that I do not just unlift something I have previously lifted, but that there is an extension inbetween.

src/main/scala/viper/gobra/translator/util/ViperUtil.scala
```diff
@@ -46,4 +46,53 @@ object ViperUtil {
       case _ => Violation.violation(s"expected vpr variable or field access, but got $left")
     }
   }
+
+  /**
+    * TODO: should be removed once the corresponding silver function is fixed.
```
Felalolf, line 51:
```suggestion
    * TODO: should be removed once the corresponding silver function is fixed, see /silver/issues/610
```

src/main/scala/viper/gobra/translator/util/ViperUtil.scala
Felalolf: I added a silver issue: https://github.com/viperproject/silver/issues/610

PR #537, opened 2022-10-03
Fixes conversion to interfaces and interfaces in struct literals

jcp19:
This PR fixes the explicit conversions to interfaces and adds an implicit conversion when there is a struct literal with a field of interface type.

### DIFF
--- a/src/main/scala/viper/gobra/ast/internal/Program.scala
+++ b/src/main/scala/viper/gobra/ast/internal/Program.scala
@@ -1063,6 +1063,11 @@ case class ShiftRight(left: Expr, right: Expr)(val info: Source.Parser.Info) ext
 }
 case class BitNeg(op: Expr)(val info: Source.Parser.Info) extends IntOperation
 
+/*
+ * Convert 'expr' to non-interface type 'newType'. If 'newType' is
+ * an interface type, then 'ToInterface' should be used instead.
+ */
+ // TODO: maybe unify with ToInterface at some point
 case class Conversion(newType: Type, expr: Expr)(val info: Source.Parser.Info) extends Expr {
   override def typ: Type = newType
 }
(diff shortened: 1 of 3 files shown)


src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -1063,6 +1063,11 @@ case class ShiftRight(left: Expr, right: Expr)(val info: Source.Parser.Info) ext
 }
 case class BitNeg(op: Expr)(val info: Source.Parser.Info) extends IntOperation
 
+/*
+ * Convert 'expr' to non-interface type 'newType'. If 'newType' is
+ * an interface type, then 'ToInterface' should be used instead.
```
ArquintL, line 1068: Could we add an assertion checking that `newType` is not an interface?

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -2541,6 +2541,11 @@ object Desugar {
                 conv: in.EffectfulConversion = in.EffectfulConversion(target, resT, dArg)(src)
                 _ <- write(conv)
               } yield target
+            case (t: InterfaceT, _) =>
```
ArquintL, line 2544: Is this already covered by another unit test? It seems that the newly added unit test does not trigger this code at all

ArquintL (approved):
I only have 2 minor comments

src/main/scala/viper/gobra/ast/internal/Program.scala
jcp19: To do that, we would need to either be able to compute the underlyingType in this class (which we cannot because we do not have typing information) or to pass the underlyingType as a parameter (which is cumbersome and may not perform as intended if one forgets to compute the actual underlyingType).

src/main/scala/viper/gobra/frontend/Desugar.scala
jcp19: `client1` covers this code. Nonetheless, I added another test (`client3`) that more clearly relies on this feature.

src/main/scala/viper/gobra/ast/internal/Program.scala
ArquintL: I see thanks

PR #539, opened 2022-10-04
Auto-update ViperServer submodule

ArquintL:


jcp19:
I'm a bit on the fence about having this by default. On one hand, its it nice to not have to do it by hand 90% of the time. On the other, we often decide to postpone updating to the latest Viper release because it may break some important assumption that Gobra (and SCION) rely on.

jcp19:
Maybe the best option is for the bot to open the PR automatically but wait for one of us to merge it

jcp19 (approved):

PR #540, opened 2022-10-04
Create gobrafied debug files

ArquintL:
If `unparse` is enabled, `.gobrafied` files will be created with the gobrafied content of the input files.
This is useful to count the number of lines of specification in the following way:
LOC = lines of code of a `.go` file counted by the usual tools
LOS = lines of code of a `.gobrafied` file - LOC + hybrid lines

A hybrid line is a line that contains actual code and annotations. The following command can be used to count these hybrid lines:
```
awk 'fname != FILENAME { printf "%s: %s\n", fname, count; fname = FILENAME; count = 0 } /.*\/\*@.*@\*\/.*/ { count++; total++ } END { print total }' $(find . -name '*.go' -or -name '*.gobra')
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ tmp/
 sync/
 
 logger.log
+.DS_Store
 
 *.go.*
 *.gobra.*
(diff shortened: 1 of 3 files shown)


ArquintL:
That would indeed be very nice. Would you look at ‚ghostness‘ of statements & expressions together with their positional information to compute this or integrate similar metrics as we use externally  today? For the latter one, a big question to me seems to be how we count .gobra files, i.e. as spec-only (& maybe even check this)

PR #542, opened 2022-10-04
Update Submodules

viper-admin:
* Updates ViperServer from `87089e6c56a3f5b5ef7258fcd258af42bc870775` to `c2622f973135f8a25cd0680ae6f18e74530d9135`.


ArquintL:
@jcp19 I've noticed that Gobra currently references a non-master commit of ViperServer (in which Felix tried out the sbt config changes we recently discussed). This PR updates ViperServer to the current commit on master.

The diff is mostly related to ViperServer's LSP frontend, i.e. necessary changes to make Viper-IDE work that should not affect ViperServer's behavior when only ViperServer's core is used (as is the case for Gobra). Furthermore, there are no changes to the referenced Silicon and Carbon commits

jcp19 (approved):

PR #545, opened 2022-10-10
Fix #544

### LINKED ISSUE #544: gobrafier: annotations are not Go 1.19 formatting compatible
Go has improve their support for doc comments with Go 1.19. As a result, the gofmt tool was adapted to support new rewrite rules and also adds spaces where it deems it appropriate.

Unfortunately, gofmt now adds spaces in front of the @ sign in the gobra annotations of Go code for doc comments.
This breaks the annotations as defined here;

https://github.com/viperproject/gobra/blob/01c38e0310eeee13ed0a4e5e86023119cb202312/src/main/scala/viper/gobra/frontend/Gobrafier.scala#L17-L18

We encountered this problem in when trying to upgrade our Go version to 1.19 in scionproto/scion#4271
When formatting the code, with gofmt 1.19, we encounter the following Gobra check error:

https://github.com/scionproto/scion/actions/runs/3197147933/jobs/5219993099

image

Dropping the space results in a passing check. But of course that is not really an option because then the file is not formatted according to gofmt -> it will be overwritten immediately in any editor that does auto formatting.

jcp19:
This PR introduces support for an extra space between `//` and `@` in Gobra comments in order to comply with the new rewriting rules of gofmt introduced in Go 1.19

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Gobrafier.scala
+++ b/src/main/scala/viper/gobra/frontend/Gobrafier.scala
@@ -14,8 +14,8 @@ import scala.util.matching.Regex
 
 object Gobrafier {
 
-  private def multilineComment(str: String): String = s"/\\*@\\s*$str\\s*@\\*/"
-  private def singlelineComment(str: String): String = s"//@\\s*$str\\s*"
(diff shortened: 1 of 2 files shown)


ArquintL (approved):
Thanks a lot for the fix!

PR #547, opened 2022-10-11
Fix lazy generation of the string domain

jcp19:
Currently, the `String` domain is always generated, even if it is not used because function `strSlice` is marked as non-lazy. This PR addresses that.


ArquintL (approved):

ArquintL:
It's a bit strange that the unit tests are now running for more than 2h. I'm restarting them but suspect that there might be a different issue

PR #548, opened 2022-10-12
Improving the Import of Termination Axioms

ArquintL:
This PR optimizes the import of axioms for termination measures to only import those that are actually needed based on the type of termination measure tuples.

Furthermore, this PR sorts the Viper members reported by `GeneratedViperMessage`

### DIFF
--- a/src/main/scala/viper/gobra/translator/Translator.scala
+++ b/src/main/scala/viper/gobra/translator/Translator.scala
@@ -15,6 +15,8 @@ import viper.gobra.translator.context.DfltTranslatorConfig
 import viper.gobra.translator.encodings.programs.ProgramsImpl
 import viper.gobra.translator.transformers.{AssumeTransformer, TerminationTransformer, ViperTransformer}
 import viper.gobra.util.Violation
+import viper.silver.ast.pretty.FastPrettyPrinter
+import viper.silver.{ast => vpr}
 
 object Translator {
 
@@ -22,6 +24,8 @@ object Translator {
     val translationConfig = new DfltTranslatorConfig()
     val programTranslator = new ProgramsImpl()
     val task = programTranslator.translate(program)(translationConfig)
+    // we report the un-transformed program as this is much more readable
+    config.reporter report GeneratedViperMessage(config.taskName, config.packageInfoInputMap(pkgInfo).map(_.name), () => sortAst(task.program), () => task.backtrack)
 
     if (config.checkConsistency) {
       val errors = task.program.checkTransitively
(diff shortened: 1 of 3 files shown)


src/main/scala/viper/gobra/translator/Translator.scala
```diff
@@ -37,8 +39,36 @@ object Translator {
       case (t, transformer) => transformer.transform(t)
     }
 
-    config.reporter report GeneratedViperMessage(config.taskName, config.packageInfoInputMap(pkgInfo).map(_.name), () => transformedTask.program, () => transformedTask.backtrack)
+    config.reporter report GeneratedViperMessage(config.taskName, config.packageInfoInputMap(pkgInfo).map(_.name), () => sortAst(transformedTask.program), () => transformedTask.backtrack)
     transformedTask
   }
 
+  def sortAst(program: vpr.Program): vpr.Program = {
```
jcp19, line 46: I would add a small comment here explaining that we are doing this to make the generated code easier to navigate

src/main/scala/viper/gobra/translator/Translator.scala
```diff
@@ -37,8 +39,36 @@ object Translator {
...
+    implicit def domainOrdering: Ordering[vpr.Domain] = Ordering.by(_.name)
+    implicit def domainFnOrdering: Ordering[vpr.DomainFunc] = Ordering.by(_.name)
+    implicit def domainAxOrdering: Ordering[vpr.DomainAxiom] = Ordering.by(FastPrettyPrinter.pretty(_))
+    implicit def fieldOrdering: Ordering[vpr.Field] = Ordering.by(_.name)
+    implicit def functionOrdering: Ordering[vpr.Function] = Ordering.by(_.name)
+    implicit def predicateOrdering: Ordering[vpr.Predicate] = Ordering.by(_.name)
+    implicit def methodOrdering: Ordering[vpr.Method] = Ordering.by(_.name)
+    implicit def extensionOrdering: Ordering[vpr.ExtensionMember] = Ordering.by(_.name)
```
jcp19, line 54: could we have only one `Ordering` for members and only introduce other orderings for other types which are not subtypes of members? In particular, I guess that `domainOrdering`, `functionOrdering`, `predicateOrdering`, `methodOrdering` and `extensionOrdering` could be merged into a single one.

jcp19 (approved):
LGTM

src/main/scala/viper/gobra/translator/Translator.scala
ArquintL: Great idea! Scala implicits seem to not work for subtypes but still better than how it is now

ArquintL:
bors merge

PR #550, opened 2022-10-13
Fixes Pure Function Calls whose Result is Unused

ArquintL:
@jcp19 I'm not feeling super confident with the writer pattern, so please take a look.


ArquintL:
@jcp19 and I have discussed different solutions yesterday and the one currently implemented is the best one we came up with. @Felalolf Do you have ideas for alternative solutions?

jcp19 (approved):
LGTM but I would wait for @Felalolf's opinion before merging

Felalolf (approved):
I would have solved it the same way (probably without the optimization).

PR #551, opened 2022-10-15
Range

Dspil:
Range for slices/arrays is reworked as discussed.
Range for maps is also implemented.
Variables declared with := in range expressions can be shared now (this is the case with Go)


src/test/scala/viper/gobra/parsing/GoParserUnitTests.scala
```diff
@@ -31,19 +31,19 @@ class GoParserUnitTests extends AnyFunSuite with Matchers with Inside {
     "package p; func f(x ...int,) {};",
     "package p; type T []int; var a []bool; func f() { if a[T{42}[0]] {} };",
     "package p; type T []int; func g(int) bool { return true }; func f() { if g(T{42}[0]) {} };",
-    "package p; type T []int; func f() { for _ = range []int{T{42}[0]} {} };",
+    "package p; type T []int; func f() { for _ = range []int{T{42}[0]} with _ {} };",
```
jcp19, line 34: Does this mean that we do not support a range loop without the `with` clause?

Dspil:
> Why is `genparser.sh` part of the changeset?

I think it is because I added execute permissions to it to run it and git captured that too.

src/test/scala/viper/gobra/parsing/GoParserUnitTests.scala
Dspil: I would think that to prove the useful stuff you would need the enumerator. Without it you can't even prove termination easily, I can make it optional though.

src/main/java/viper/gobra/frontend/GobraLexer.java
```diff
@@ -1,4 +1,4 @@
-// Generated from /Users/joao/Code/gobraHome/gobra/src/main/antlr4/GobraLexer.g4 by ANTLR 4.9.2
+// Generated from /home/shit/repos/GobraHome/gobra/src/main/antlr4/GobraLexer.g4 by ANTLR 4.9.1
```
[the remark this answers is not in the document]
Dspil: There doesn't seem to be such a command line option. I removed my path though

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -797,7 +797,7 @@ sealed trait PMisc extends PNode
 
 sealed trait PActualMisc extends PMisc
 
-case class PRange(exp: PExpression) extends PActualMisc
+case class PRange(exp: PExpression, enumerated: PUnkLikeId) extends PActualMisc
```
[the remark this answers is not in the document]
Dspil: Ok, makes sense, I will make the 'with' clause optional

Felalolf (commented):
A few smaller comments. I did not check the desugaring thoroughly, but just that it does not break existing stuff.

src/main/scala/viper/gobra/reporting/DefaultErrorBackTranslator.scala
```diff
@@ -182,6 +179,12 @@ class DefaultErrorBackTranslator(
           case (err, reason) => err dueTo reason
         }
 
+      case _ / InsufficientPermissionToRangeExpressionAnnotation() =>
+        x.reasons.foldLeft(InsufficientPermissionToRangeExpressionError(x.info): VerificationError){ case (err, reason) => err dueTo reason }
```
jcp19, line 183: Why do we need an extra annotation and error? What would be the message shown without this new annotation?

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1288,113 +1300,87 @@ object Desugar {
...
-            *     j = c[i]
-            * }
-            *
-            * invariant len(c) > 0 ==> 0 <= i0 && i0 <= len(c) // these invariants do not
-            * invariant len(c) > 0 ==> i == i0                 // require access to anything so they
-            * invariant len(c) > 0 ==> 0 <= i && i <= len(c)   // are added before the user invariants
-            * <invariants were i is replaced with f(i, len(c) > 0) and j with f(j, len(c) > 0)>
-            * invariant len(c) > 0 ==> (0 <= i && i < len(c) ==> j == c[i]) // [v]
-            * for i < len(c) {
+            * if (len(x) == 0) {
```
jcp19, line 1313: How do you guarantee here that `x` is evaluated, at most, once?

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1404,163 +1390,341 @@ object Desugar {
...
-            * var i0 int = 0 // since 'expIndex' can change in the iteration we store the true index in i0
-            *
-            * c := x // save the value of the slice/array since changing it doesn't change the iteration
+            * The following code is produced. Note
+            * that everything is in a new block so it can shadow variables with the same
+            * name declared outside. This is also the go behaviour.
             *
-            * if (len(c) > 0) {
-            *     expIndex = 0
+            * if (|x| == 0) {
```
jcp19, line 1585: same here

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1404,163 +1390,341 @@ object Desugar {
...
-              rangeSrc = meta(range, info)
-              indexSrc = meta(ass(0), info)
-              rangeExpSrc = meta(range.exp, info)
 
-              copiedVar <- freshDeclaredExclusiveVar(exp.typ, n, info)(rangeExpSrc)
-              copyAss = singleAss(in.Assignee.Var(copiedVar), exp)(rangeExpSrc)
-              length = in.Length(copiedVar)(rangeSrc)
+              keyType = exp.typ.asInstanceOf[in.MapT].keys.withAddressability(Addressability.exclusiveVariable)
+              valType = exp.typ.asInstanceOf[in.MapT].values.withAddressability(Addressability.exclusiveVariable)
+              visType = in.SetT(keyType, Addressability.exclusiveVariable)
```
jcp19, line 1623: I think that the declaration of `keyType` and `valType` here could be a bit safer by using a case analysis:
```
              val (keyType, valType) = underlyingType(exp.typ) match {
                case in.MapT(k, v, _) => (k.withAddressability(???), v.withAddressability(???))
                case _ => violation(???)
              }
```
Also, notice that I'm using `underlyingType` here applied to `exp.type`. I'm not sure if we need `underlyingType` here but my impression is that, without it, we may run into exceptions when ranging over a defined type.  I also wonder why we are creating `exclusive` types no matter what. Shouldn't it depend on the addressability vector in `n`?

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1404,163 +1390,341 @@ object Desugar {
...
-              incrIndex = singleAss(indexLeft, copiedIndexVar)(indexSrc)
+              visited <- leftOfAssignmentD(range.enumerated, info)(in.SetT(keyType, Addressability.exclusiveVariable))
 
-              cond = in.LessCmp(indexLeft.op, length)(rangeSrc)
+              perm <- freshDeclaredExclusiveVar(in.PermissionT(Addressability.exclusiveVariable), n, info)(src)
 
-              copiedIndexEqualsIndexInv = in.Implication(
-                in.LessCmp(in.IntLit(0)(indexSrc), length)(indexSrc),
-                in.ExprAssertion(in.EqCmp(indexLeft.op, copiedIndexVar)(rangeSrc))(rangeSrc))(rangeSrc)
+              initPerm = singleAss(in.Assignee.Var(perm), in.PermLit(1, 1000000)(src))(src)
```
jcp19, line 1631: Is this hardcoded in other places too? If so, I would move it to a constant to make it easier to change

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1404,163 +1390,341 @@ object Desugar {
...
+                      in.LessCmp(in.Length(visited.op)(src), in.Length(exp)(src))(src),
+                      dInv ++ addedInvariants, dTer, in.Block(Vector(continueLoopLabelProxy, k) ++ (if (hasValue) Vector(v) else Vector()),
+                    Vector(exhalePerm, updateKeyVal, dBody, continueLoopLabel, inhalePerm, updateVisited) ++ dInvPre ++ dTerPre
+                  )(src))(src), visitedEqDomain, breakLoopLabel
+                ))(src)
+              )(src)
+
+            } yield ifstmt))
+
+          case n@PAssForRange(range, ass, spec, body) if underlyingType(info.typ(range.exp)).isInstanceOf[SliceT] || underlyingType(info.typ(range.exp)).isInstanceOf[MapT] =>
```
jcp19, line 1694: we are computing `underlyingType` multiple times just to be able to deal with different types in multiple cases. Why not have a single case for `PAssForRange` and then, do, as the first thing, a case analysis on underlyingType of range.exp? What I would even do is to introduce separate methods to handle these cases and just call them from the case analysis.

src/main/scala/viper/gobra/frontend/Desugar.scala
jcp19: I think that there are other places where my comments may apply

jcp19 (commented):
It is a bit hard for me to follow the Desugarer, as it seems very long and repetitive. Is there a way we can nicely refactor this into smaller, easy to grok methods?

src/main/scala/viper/gobra/reporting/DefaultErrorBackTranslator.scala
Dspil: It would be "invariant may not be established" and it would point to the whole range-loop structure since that invariant is automatically added and does not exist in the original ast. It would be confusing

src/main/scala/viper/gobra/frontend/Desugar.scala
Dspil: At that point though that I am doing the cast, we have already pattern matched on the type of the expression and we know it is a map.

src/main/scala/viper/gobra/frontend/Desugar.scala
Dspil: I have updated the comment to show that only c is used. Also, the length of the slice/array is also saved in the beginning because it might change but Go retains the original length

jcp19 (approved):
We can merge it by me

PR #560, opened 2022-11-01
Add distinction between shared and exclusive global variables

jcp19:


Dspil (approved):

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
```diff
@@ -108,7 +109,7 @@ trait Assignability extends BaseProperty { this: TypeInfoImpl =>
 
   lazy val assignable: Property[PExpression] = createBinaryProperty("assignable") {
     case PIndexedExp(b, _) => underlyingType(exprType(b)) match {
-      case _: ArrayT => assignable(b)
+      case _: ArrayT => assignable(b) && isMutableArray(b)
```
Felalolf, line 112: Do you catch everything with that. I would have expected some case in the beginning like `case e _ if !isMutable(e) => ...`

src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
```diff
@@ -50,16 +51,16 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
         })
 
     case n@PAssignment(rights, lefts) =>
-      rights.flatMap(isExpr(_).out) ++ lefts.flatMap(isExpr(_).out) ++
+      rights.flatMap(isExpr(_).out) ++ lefts.flatMap(isExpr(_).out) ++ lefts.flatMap(isNotExclusiveGlobal) ++
```
Felalolf, line 54: this should be caught by assingable (i.e. `isNotExclusiveGlobal` should not be there).

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
```diff
@@ -195,12 +195,20 @@ trait TypeEncoding extends Generator {
...
     case v: in.GlobalConst if typ(ctx) isDefinedAt v.typ => unit(ctx.fixpoint.get(v)(ctx))
     case (v: in.BodyVar) :: t / Exclusive if typ(ctx).isDefinedAt(t) => unit(variable(ctx)(v).localVar)
+    case (v: in.GlobalVar) :: t / Exclusive if typ(ctx).isDefinedAt(t) =>
+      val (pos, info, errT) = v.vprMeta
+      val typ = ctx.typ(v.typ)
+      val vprExpr = vpr.FuncApp(
+        funcname = v.name.uniqueName,
+        args = Seq.empty
+      )(pos, info, typ, errT)
+      unit(vprExpr)
```
Felalolf, line 211: I am not sure right now, but should this not be in the file with the globals encoding?

Felalolf (approved):
I have some comments. After they are addressed, it is fine.

Issue #234, opened 2021-03-11
Tests failing non-deterministically


jcp19:
The test `regressions/examples/impl_errors/parallel_search_replace.gobra` has been failing non-deterministically in our CI with a NPE in silicon. Below, I attached the stack trace.  Running it locally leads to either non-termination or to the expected verification results.

```
[info]   java.util.concurrent.ExecutionException: java.util.concurrent.ExecutionException: java.lang.NullPointerException
[info]   at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122)
[info]   at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:191)
[info]   at viper.silicon.Silicon.verify(Silicon.scala:199)
[info]   at viper.silicon.Silicon.verify(Silicon.scala:157)
[info]   at viper.gobra.backend.Silicon.$anonfun$verify$1(Silicon.scala:27)
[info]   at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:672)
[info]   at scala.concurrent.impl.Promise$Transformation.run(Promise.scala:431)
[info]   at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
[info]   at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
[info]   at java.base/java.lang.Thread.run(Thread.java:834)
[info]   ...
[info]   Cause: java.util.concurrent.ExecutionException: java.lang.NullPointerException
[info]   at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122)
[info]   at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:191)
[info]   at viper.silicon.verifier.DefaultMasterVerifier.$anonfun$verify$9(DefaultMasterVerifier.scala:240)
[info]   at scala.collection.immutable.List.flatMap(List.scala:293)
[info]   at scala.collection.immutable.List.flatMap(List.scala:79)
[info]   at viper.silicon.verifier.DefaultMasterVerifier.verify(DefaultMasterVerifier.scala:240)
[info]   at viper.silicon.Silicon.viper$silicon$Silicon$$runVerifier(Silicon.scala:235)
[info]   at viper.silicon.Silicon$$anon$1.call(Silicon.scala:193)
[info]   at viper.silicon.Silicon$$anon$1.call(Silicon.scala:192)
[...]

PR #563, opened 2022-11-03
Disable problematic tests

jcp19:
As reported in issue #234, some of our tests seem to not terminate non-deterministically. This PR disables a few of them that are stalling our CI


Felalolf (approved):

jcp19:
Only bors failed for some reason, I will merge it

PR #570, opened 2022-11-10
Optimize map encoding

jcp19:
Changes:
- Generates as many fields as possible to represent different map types, instead of generating a single field called `underlyingMapField`. Should have performance implications like PR #568. (This is somewhat confirmed by our experiments with SCION, where verifying a pacakge that uses maps heavily went down from 17min to 12)
- Removes one level of indirection which, in hindsight, does not seem to be necessary (through the `getMap` function)

In a separate PR, I plan to generate more auxiliary functions for `len` and `lookup`, which should expose better triggers at the surface language

PS: some trivial changes like inserting spaces and what not happened because my IDE decided randomly that it is good to automatically apply a code style now -.-

### DIFF
--- a/src/main/scala/viper/gobra/translator/encodings/maps/MapEncoding.scala
+++ b/src/main/scala/viper/gobra/translator/encodings/maps/MapEncoding.scala
@@ -24,13 +24,6 @@ import viper.silver.verifier.reasons.AssertionFalse
 import viper.silver.verifier.{ErrorReason, errors => err}
 import viper.silver.{ast => vpr}
 
-// TODO:
-// - use one ref instead of id
-// - use different field names when the maps are different
-// - use function generator for lookup functions (create this too, for better triggers); maybe create more functions
-//   to allow for better triggers in general
-// - delete GobraMap domain
-
 /**
   * Encoding for Go maps. Unlike slices, maps are not thread-safe;
   * thus, all concurrent accesses to maps must be synchronized. In particular,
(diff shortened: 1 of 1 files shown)


src/main/scala/viper/gobra/translator/encodings/maps/MapEncoding.scala
```diff
@@ -250,40 +247,22 @@ class MapEncoding extends LeafTypeEncoding {
     * [ mapExp[idx] = newVal ] ->
     *     var res: Ref
```
Felalolf, line 248:
```suggestion
```

src/main/scala/viper/gobra/translator/encodings/maps/MapEncoding.scala
```diff
@@ -250,40 +247,22 @@ class MapEncoding extends LeafTypeEncoding {
...
     *     var m: Map[ [k], [v] ]
-    *     m = getCorrespondingMap(mapExp)
-    *     inhale getMap(res) == m[ [idx] = [newVal] ]
-    *     m.underlyingField = res
+    *     m = [mapExp].underlyingMapField
+    *     [mapExp].underlyingMapField == m[ [idx] = [newVal] ]
```
Felalolf, line 251:
```suggestion
    *     [mapExp].underlyingMapField := m[ [idx] = [newVal] ]
```

Felalolf (approved):

Dspil (approved):

PR #574, opened 2022-11-13
Fix issue 573

jcp19:


src/main/scala/viper/gobra/frontend/info/implementation/resolution/Enclosing.scala
```diff
@@ -72,6 +72,12 @@ trait Enclosing { this: TypeInfoImpl =>
   lazy val tryEnclosingFunction: PNode => Option[PFunctionDecl] =
     down[Option[PFunctionDecl]](None) { case m: PFunctionDecl => Some(m) }
 
+  lazy val tryEnclosingFunctionOrMethod: PNode => Option[PFunctionOrMethodDecl] =
+    down[Option[PFunctionOrMethodDecl]](None) {
+      case f: PFunctionDecl => Some(f)
+      case m: PMethodDecl => Some(m)
```
ArquintL, line 78: Wouldn't it make more sense to match here on PFunctionOrMethodDecl?

ArquintL (approved):
One small comment otherwise looks good to me

src/main/scala/viper/gobra/frontend/info/implementation/resolution/Enclosing.scala
jcp19: Good point! I will change it

PR #588, opened 2023-01-04
Add support for ADTs as termination measures

jcp19:
Add support for using ADT instances as termination measures. This is just a draft, as I will likely adapt the PR such that:
- we introduce support for `len` for ADTs, which compute the rank of an ADT
- we encode the rank function
The advantage is that we expose `rank` directly to the user, allowing us to prove theorems about it more easily

### DIFF
--- a/src/main/scala/viper/gobra/translator/encodings/adts/AdtEncoding.scala
+++ b/src/main/scala/viper/gobra/translator/encodings/adts/AdtEncoding.scala
@@ -227,10 +227,81 @@ class AdtEncoding extends LeafTypeEncoding {
           vpr.Forall(Seq(variableDecl), Seq(trigger), vu.bigOr(equalities)(aPos, aInfo, aErrT))(aPos, aInfo, aErrT)
         )(aPos, aInfo, adtName, aErrT)
       }
+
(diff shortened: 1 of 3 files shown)


src/main/scala/viper/gobra/translator/encodings/adts/AdtEncoding.scala
```diff
@@ -227,10 +227,109 @@ class AdtEncoding extends LeafTypeEncoding {
...
+        }
+
+        // forall x, y, z: X :: { decreasing(x,y), decreasing(y,z) } decreasing(x,z)
+        val decreasingTransitive = {
+          val varX = vpr.LocalVarDecl("X", adtT)(aPos, aInfo, aErrT)
+          val varY = vpr.LocalVarDecl("Y", adtT)(aPos, aInfo, aErrT)
+          val varZ = vpr.LocalVarDecl("Z", adtT)(aPos, aInfo, aErrT)
+          val decrApp1 = applyDecreasing(varX.localVar, varY.localVar)
+          val decrApp2 = applyDecreasing(varY.localVar, varZ.localVar)
+          val decrApp3 =  applyDecreasing(varX.localVar, varZ.localVar)
```
jcp19, line 289:
```suggestion
          val decrApp3 = applyDecreasing(varX.localVar, varZ.localVar)
```

PR #590, opened 2023-01-05
Streaming errors

Dspil:
Instead of printing the  verification errors after the verification for each package ends, stream them.

### DIFF
--- a/src/main/scala/viper/gobra/backend/BackendVerifier.scala
+++ b/src/main/scala/viper/gobra/backend/BackendVerifier.scala
@@ -9,7 +9,7 @@ package viper.gobra.backend
 import viper.gobra.backend.ViperBackends.{CarbonBackend => Carbon}
 import viper.gobra.frontend.{Config, PackageInfo}
 import viper.gobra.reporting.BackTranslator.BackTrackInfo
-import viper.gobra.reporting.{BackTranslator, BacktranslatingReporter, ChoppedProgressMessage}
+import viper.gobra.reporting.{BackTranslator, BacktranslatingReporter, StreamingReporter, ChoppedProgressMessage}
 import viper.gobra.util.{ChopperUtil, GobraExecutionContext}
 import viper.silver
 import viper.silver.verifier.VerificationResult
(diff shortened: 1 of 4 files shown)


src/main/scala/viper/gobra/backend/BackendVerifier.scala
```diff
@@ -49,7 +49,7 @@ object BackendVerifier {
 
     val verificationResults: Future[VerificationResult] =  {
       val verifier = config.backend.create(exePaths, config)
-      val reporter = BacktranslatingReporter(config.reporter, task.backtrack, config)
+      val reporter = BacktranslatingReporter(StreamingReporter(config.reporter), task.backtrack, config)
```
ArquintL, line 52: @Dspil Would it be possible to set the `StreamingReporter` in the config instead of wrapping the reporter here?
What I image is that the reporter (or default reporter) for Gobra via CLI is the streaming reporter (maybe streaming reporter wrapping the current reporter). In this way, Gobra-IDE could use the existing reporter (i.e. a different one) without observing any changes

src/main/scala/viper/gobra/reporting/StreamingReporter.scala
```diff
@@ -0,0 +1,24 @@
...
+
+import com.typesafe.scalalogging.StrictLogging
+
+case class StreamingReporter(reporter: GobraReporter) extends GobraReporter with StrictLogging {
+  override val name: String = reporter.name
+
+  def report(msg: GobraMessage): Unit = {
+    msg match {
+      case m:GobraEntityFailureMessage => m.result match {
+        case VerifierResult.Failure(errors) => errors.foreach(err => logger.error(s"Error: ${err.formattedMessage}"))
```
jcp19, line 17: This way, we will only log the errors if we are using the `StreamingReporter`, which will probably not be the case with the IDE right? Could we do the reporting somewhere else, like in `BacktranslatingReporter`?

ArquintL: Although this is correct, I don't think errors printed to StdOut should be of concern for the IDE

src/main/scala/viper/gobra/reporting/StreamingReporter.scala
```diff
@@ -0,0 +1,24 @@
...
+        case _ => ;
+      }
+      case _ => ;
```
jcp19, line 20:
```suggestion
        case _ =>
      }
      case _ =>
```

jcp19 (approved):
LGTM but I would wait for Linard's approval to confirm that this has no impact in the GobraIDE

PR #594, opened 2023-01-06
More precise triggers for smake

jcp19:
(as suggested by Peter)


Felalolf (approved):

jcp19:
I have tested this with in all verified packages in SCION and did not observe any differences in verification results. I did observe a small change in verification time for the router (an increase in ~2min30s with the parallelizeBranches flag - the package takes around 30 min to verify) but, for a few problematic examples, we have seen a major reduction in verification time with these changes.

PR #597, opened 2023-01-13
print early errors that don't go through the streaming reporter

Dspil:
Errors that do not get produced by the backend (typechecking and parsing errors) didn't pass through the streaming reporter so they were not reported. Now the verificationFailure instead of a vector of errors takes an Either as a parameter to distinguish between errors produced by the backend or not so they are printed if the streaming reporter hasn't printed them already. A simple example is the following:

```
package test

import "context"

func f() 
```
If context is not in the directories passed as arguments to Gobra, Gobra produces an error saying that this qualifier could not be derived. This kind of error did not pass through the reporter so it was not printed.


src/main/scala/viper/gobra/reporting/VerifierResult.scala
```diff
@@ -13,7 +13,7 @@ sealed trait VerifierResult
 object VerifierResult {
 
   case object Success extends VerifierResult
-  case class Failure(errors: Vector[VerifierError]) extends VerifierResult
+  case class Failure(errors: Either[Vector[VerifierError], Vector[VerifierError]]) extends VerifierResult
```
jcp19, line 16: It is not clear here in which case we should put our errors on `Left` and when we should put them on `Right`

src/main/scala/viper/gobra/reporting/DefaultMessageBackTranslator.scala
```diff
@@ -11,9 +11,18 @@ import viper.gobra.frontend.Config
 import viper.gobra.reporting.BackTranslator.BackTrackInfo
 import viper.silver.reporter.{EntityFailureMessage, EntitySuccessMessage, Message, OverallFailureMessage, OverallSuccessMessage}
 import viper.silver.verifier.VerificationResult
+import com.typesafe.scalalogging.StrictLogging
 
-class DefaultMessageBackTranslator(backTrackInfo: BackTrackInfo, config: Config) extends MessageBackTranslator {
+class DefaultMessageBackTranslator(backTrackInfo: BackTrackInfo, config: Config) extends MessageBackTranslator with StrictLogging {
   override def translate(msg: Message): GobraMessage = {
+    // TODO: Remove match when issue https://github.com/viperproject/gobra/issues/556 is fixed
```
Felalolf, line 18: add to the issue a reminder to adapt this code please.

Felalolf (approved):
From my side, looks ok, but let joao also have a look

PR #600, opened 2023-01-22
Allow non-side-effectful conversions in spec

jcp19:
Currently, type conversions cannot have ghost child nodes. This is unnecessarily restrictive. This PR lifts such a restriction and allows non-effectful conversions to be used in spec.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
@@ -1055,7 +1055,7 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
       case t => violation(s"unexpected argument ${expr.exp} of type $t passed to len")
     }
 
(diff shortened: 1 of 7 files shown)


src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -1054,4 +1054,13 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
       case _: SequenceT | _: SetT | _: MultisetT | _: MathMapT | _: AdtT => UNTYPED_INT_CONST
       case t => violation(s"unexpected argument ${expr.exp} of type $t passed to len")
     }
+
+  override def isEffectfulConversion(c: ap.Conversion): Boolean = {
```
Felalolf, line 1058: please add a comment, whether these are all cases, or just all cases that are relevant for now.

Felalolf (approved):
Small comment otherwise ok.

PR #604, opened 2023-01-24
Better error spec

jcp19:
The version of error introduced in the previous spec caused a big performance degradation in verifiedSCION (likely, due to the use of wildcards in the spec of `Error`. This PR addresses that and recovers the original expressivity of the `error` type (in particular, implementations of `Error` that mutate the underlying data structure are still allowed)

### DIFF
--- a/src/main/resources/builtin/builtin.gobra
+++ b/src/main/resources/builtin/builtin.gobra
@@ -13,7 +13,18 @@ type any = interface{}
 type error interface {
 	pred ErrorMem()
 
-	requires acc(ErrorMem(), _)
+    ghost
+    requires acc(ErrorMem(), _)
+    decreases
+    pure IsDuplicableMem() bool
+
+    ghost
+	preserves ErrorMem()
+	ensures   IsDuplicableMem() ==> ErrorMem() 
+    decreases
+    Duplicate()
+
+	preserves ErrorMem()
 	decreases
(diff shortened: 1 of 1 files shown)


src/main/resources/builtin/builtin.gobra
```diff
@@ -13,8 +13,19 @@ type any = interface{}
 type error interface {
 	pred ErrorMem()
 
+	ghost
 	requires acc(ErrorMem(), _)
 	decreases
+	pure IsDuplicableMem() bool
+
+	ghost
+	preserves ErrorMem()
```
ArquintL, line 22: Given that `IsDuplicableMem` takes a wildcard permission amount, it might make sense to add a `perm` parameter to `Duplicate`. Otherwise, `IsDuplicableMem`'s contract could be changed to take full permission

jcp19: I would rather not complicate the signature of `Duplicate` with a perm ammount. Furthermore, I think that this will still result in the same performance degradation as we currently have with the wildcard version. Regarding the second suggestion, if I understand it correctly, any permission to a predicate amount in a pure function's precondition is now encoded as a wildcard - as such, there is no point in changing the syntax to something that is interpreted in the same way

ArquintL (approved):

PR #615, opened 2023-02-06
Allow go calls of closures

jcp19:


src/main/scala/viper/gobra/translator/encodings/closures/ClosureSpecsEncoder.scala
```diff
@@ -81,6 +83,31 @@ protected class ClosureSpecsEncoder {
...
+        case e => Violation.violation(s"expected a function or method, but found $e instead.")
+      }
+      formalParams = formalParamDecls.map(_.localVar)
+      substitutions = (formalParams zip vprargs).toMap
+      instantiatedPres = pres.map(_.replace(substitutions))
+      and = ViperUtil.bigAnd(instantiatedPres)(pos, info, errT)
+      exhale = vpr.Exhale(and)(pos, info, errT)
+      _ <- errorT {
+        case err.ExhaleFailed(Source(info), _, _) => PreconditionError(info).dueTo(GoCallPreconditionReason(info))
+      }
```
Felalolf, line 107: I believe the code writer has a method to generate an exhale together with adding an error transformation. I recommend using that one to avoid errors where the error transformer is hocked incorrectly. That said, I am also fine if you prefer this style.

Felalolf (approved):
Looks ok to me.

PR #628, opened 2023-02-20
Add `--conditionalizePermissions`

jcp19:


ArquintL (approved):
Looks good to me but I was wondering why we do not stick to `conditionalizePermissions`? It would cause less confusion on our end because it would be more obvious that this indeed corresponds to the same option in silicon. For Gobra users, I do not see a benefit for going with `conditionalizeConditions` because (based on the good example) it seems to apply to permissions only

jcp19:
> Looks good to me but I was wondering why we do not stick to `conditionalizePermissions`? It would cause less confusion on our end because it would be more obvious that this indeed corresponds to the same option in silicon. For Gobra users, I do not see a benefit for going with `conditionalizeConditions` because (based on the good example) it seems to apply to permissions only

This was an accident, I mixed the two names by mistake xD

PR #633, opened 2023-03-06
Support outlines in closures

jcp19:


Felalolf (approved):
Did you make sure that the offset computation that is used for the name of the outline block is fine too?

Felalolf (changes requested):
I checked it myself. The relativeId function is incorrect. It uses `codeRoot`, which returns the enclsoing closure. Instead, it must use some function that returns the enclosing function or method.

Felalolf (approved):

PR #634, opened 2023-03-09
Parallelizing Gobra

ArquintL:
This PR introduces several optimizations to Gobra:
- Parallel parsing
  - A "fast parser" (ANTLR parsing the 'preamble' of a file) quickly identifies package imports (per file). This introduces some overhead as a file ends up being parsed twice. I see quite a variance for parsing this preamble ranging from <1% to 50% of the time it takes to parse the entire file.
  - Since I had to re-generate the parser, I've updated to the latest ANTLR release will also required updating the Go lexer and Go parser because the previous lexer and parser caused an error.
  - There are currently some problems left related to lazily parsing files but the speeding up parsing makes a huge difference.
-  Explicit type-check dependencies
  - `TypeInfo` (i.e. the type-check instance per package) gets a `dependentTypeInfo` map that maps the package's (direct) imports to the corresponding type-check result.
  - Since `dependentTypeInfo` makes the dependencies for type-checking explicit, this enables us to cache (in memory) type-checking results because we no longer depend on a `Context` instance that changes for every verification job.
  - Type-checking can be performed lazily, sequentially or in parallel, which is configurable via CLI
- Parallel desugaring
  - Although impact is low (since desugaring is fast compared to other steps performed by Gobra), the main package (i.e. the one that is getting verified) and imported packages are now desugared in parallel


src/main/scala/viper/gobra/frontend/TaskManager.scala
```diff
@@ -0,0 +1,100 @@
...
+class TaskManager[K, I, R](mode: TaskManagerMode)(implicit executor: GobraExecutionContext) {
+  private val jobs: ConcurrentMap[K, Job[I, R]] = new ConcurrentHashMap()
+
+  def addIfAbsent(id: K, job: Job[I, R]): Unit = {
+    var isAbsent = false
+    jobs.computeIfAbsent(id, _ => {
+      isAbsent = true
+      job
+    })
+    if (isAbsent) {
```
Felalolf, line 66: I am not familiar with ConcurrentMap, but this looks a bot racey to me. Is the body given to `computeIfAbsent` executed in the same thread? Otherwise, `isAbsent` might be set to true after the check on line 66.

src/main/scala/viper/gobra/frontend/TaskManager.scala
```diff
@@ -0,0 +1,100 @@
...
+  def getResultBlocking(id: K): R = {
+    Await.result(getResult(id), Duration.Inf)
+  }
+
+  def getAllResultsWithKeys: Future[Map[K, R]] = {
+    val futs = jobs.asScala.toVector.map { case (key, job) =>
+      mode match {
+        case Lazy => job.execute() // now we need the job's result
+        case _ =>
+      }
```
Felalolf, line 95: I do not fully understand why the lazy jobs are executed here. Could you not just pass the future to it?

src/main/scala/viper/gobra/frontend/info/TypeInfo.scala
```diff
@@ -8,14 +8,16 @@ package viper.gobra.frontend.info
...
 import viper.gobra.frontend.info.base.SymbolTable.{MethodImpl, MethodSpec, Regular, TypeMember}
 import viper.gobra.frontend.info.base.Type.{InterfaceT, Type}
 import viper.gobra.frontend.info.implementation.resolution.{AdvancedMemberSet, MemberPath}
+import viper.gobra.reporting.VerifierError
 import viper.gobra.theory.Addressability
 
 trait TypeInfo extends ExternalTypeInfo {
 
-  def context: Info.Context
+  def dependentTypeInfo: Map[AbstractImport, () => Either[Vector[VerifierError], ExternalTypeInfo]]
```
Felalolf, line 20: Is this not already inherited via `ExternalTypeInfo`?

src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
```diff
@@ -403,44 +402,19 @@ trait MemberResolution { this: TypeInfoImpl =>
...
-        errs => context.addErrenousPackage(importTarget, errs)(config),
-        info => context.addPackage(importTarget, info)(config)
-      )
-      res
-    }
-
     def createImportError(errs: Vector[VerifierError]): Messages = {
       // create an error message located at the import statement to indicate errors in the imported package
-      // we distinguish between parse and type errors, cyclic imports, and packages whose source files could not be found
+      // we distinguish between regular errors and packages whose source files could not be found (not that cyclic
```
Felalolf, line 408:
```suggestion
      // we distinguish between regular errors and packages whose source files could not be found (note that cyclic
```

src/main/scala/viper/gobra/frontend/info/Info.scala
```diff
@@ -6,82 +6,268 @@
...
+import viper.gobra.util.{GobraExecutionContext, Violation}
 
-import scala.collection.immutable.ListMap
+import java.security.MessageDigest
+import java.util.concurrent.atomic.AtomicLong
+import java.util.concurrent.{ConcurrentHashMap, ConcurrentMap}
+import scala.concurrent.Future
 
-object Info {
+object Info extends LazyLogging {
```
Felalolf, line 30: I have seen that you extneded LazyLogging in some places. I only have a rough idea why. Please add a comment to the code explaining why this is necessary.

Felalolf (approved):
I did not look at the Parser and Info in detail. I would need more time to look at that. Please add some more documentation to each of the jobs that gives an overview of what the job is doing, in particular before the ParseJob trait and TypeCheckJob.

src/main/antlr4/GoParser.g4
```diff
@@ -194,7 +195,7 @@ commCase: CASE (sendStmt | recvStmt) | DEFAULT;
 
 recvStmt: (expressionList ASSIGN | identifierList DECLARE_ASSIGN)? recvExpr = expression;
 
-forStmt: FOR (expression | forClause | rangeClause)? block;
+forStmt: FOR (expression? | forClause | rangeClause?) block;
```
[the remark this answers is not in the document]
ArquintL: I've simply taken this over from ANTLR. However, I've added a unit test checking that empty for clause are still correctly parsed

src/main/scala/viper/gobra/frontend/TaskManager.scala
ArquintL: It's the same thread but made it clearer now (and also use `putIfAbsent` since we do not really compute anything in the closure)

src/main/scala/viper/gobra/frontend/TaskManager.scala
ArquintL: While a future exists for every job (which is created at job construction time), this future will never complete unless `execute` is called to actually start the processing of the job. In the case of `SEQUENTIAL` and `PARALLEL`, jobs are started on insert to the concurrent map but `LAZY` jobs aren't so we have to trigger their execution here

PR #646, opened 2023-04-12
Print time-per-step when running with log level `DEBUG`

jcp19:
When running with the option `--logLevel=DEBUG` or `--debug`, Gobra will print information about how long each step of Gobra takes, as exemplified below:
```
[info] Verifying package .. - router [20:11:02]
[info] Parsing finished in 9.318s.
[info] Type-checking finished in 14.857s.
[info] Desugaring finished in 2.17s.
[info] Internal transformations performed in 0.643s.
[info] Encoding performed in 19.759s.
```

Furthermore, this PR drops the annoying error message that we get when using closures.


ArquintL:
PR #634 will introduce the same time information. One short-coming of your PR is that the parsing time does exclude parsing of all imported packages and likewise the time reported for type-checking contains the parsing & type-checking time for all imported packages. While I agree that this information is useful nevertheless, it could be confusing though

jcp19:
@ArquintL In that case, we can close this PR. I wonder if we can merge the time reports from your PR in a separate PR already

PR #660, opened 2023-07-04
Fix #659

### LINKED ISSUE #659: Quantified Let Expressions in Encoding Unsoundness
The following code snippet successfully verifies although a verification error is expected for the `assert false` statement.
This is a duplicate of [Silicon #736](https://github.com/viperproject/silicon/issues/736) because Gobra encodes the following program using several quantifiers and let expressions.
This bug has recently been introduced in Viper and older Viper versions (e.g. used by the current stable release) are not affected.

```
package issue659

type Node struct {
    ok     bool
}

pred (s *Set)mem(){
    acc(s) &&
    acc(s.nodes) &&
    forall i int:: i in domain(s.nodes) ==> acc(s.nodes[i])
}

type Set struct {
    nodes map[int]*Node
}

requires s.mem()
requires acc(n)
requires !(k in unfolding s.mem() in domain(s.nodes))
func (s *Set) add2(n *Node, k int){
    unfold s.mem()
    _,ok := s.nodes[0];
    if ok {
        s.nodes[k] = n
        fold s.mem()
        assert false // should fail
        return
    }
}
```

ArquintL:
Fixes #659
Includes changes of #658

### DIFF
--- a/src/test/resources/regressions/issues/000659.gobra
+++ b/src/test/resources/regressions/issues/000659.gobra
@@ -1,0 +1,33 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package issue659
+
+type Node struct {
+    ok     bool
+}
+
+pred (s *Set)mem(){
+    acc(s) &&
+    acc(s.nodes) &&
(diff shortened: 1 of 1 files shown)


ArquintL:
bors merge

PR #666, opened 2023-07-28
Add flag to use Z3 via API

Dspil:
Markus noticed a speedup when running silicon with Z3 via API in scion generated files.

### DIFF
--- a/src/main/scala/viper/gobra/backend/Silicon.scala
+++ b/src/main/scala/viper/gobra/backend/Silicon.scala
@@ -36,12 +36,10 @@ class Silicon(commandLineArguments: Seq[String]) extends ViperVerifier {
 
         result
       } catch {
-        case _: java.lang.UnsatisfiedLinkError => println("Couldn't find Z3 java API. No libz3java in java.library.path")
+        case _: java.lang.UnsatisfiedLinkError => System.err.println("Couldn't find Z3 java API. No libz3java in java.library.path")
           new Failure(Seq.empty)
         case e: Throwable =>
-          println("Couldn't verify package due to internal error:")
-          println(e.getMessage())
-          new Failure(Seq.empty)
+          throw e
       }
     }
   }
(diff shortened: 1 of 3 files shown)


jcp19 (approved):
LGTM. I wonder what happens when the Z3 API cannot be found though. Does it show an ugly exception or an informative error message? Would there be a way to pass a path to the Z3 API from Gobra?

src/main/scala/viper/gobra/backend/Silicon.scala
```diff
@@ -23,18 +22,27 @@ class Silicon(commandLineArguments: Seq[String]) extends ViperVerifier {
       val backend: silicon.Silicon = silicon.Silicon.fromPartialCommandLineArguments(commandLineArguments, reporter)
       
       val startTime = System.currentTimeMillis()
-      backend.start()
-      val result = backend.verify(program)
-      backend.stop()
+      try {
+        backend.start()
+        val result = backend.verify(program)
```
jcp19, line 27: We should minimize the amount of code inside the try block.
Why don't we put the try/catch around `backend.verify(program)` instead of putting all the code inside it?
Sth like
```scala
val result = 
    try {
        backend.verify(program)
    } catch ...
```
try/catch blocks are [expressions too](https://docs.scala-lang.org/overviews/scala-book/functional-error-handling.html).

src/main/scala/viper/gobra/backend/Silicon.scala
```diff
@@ -23,18 +22,27 @@ class Silicon(commandLineArguments: Seq[String]) extends ViperVerifier {
...
 
-      result
+        result
+      } catch {
+        case _: java.lang.UnsatisfiedLinkError => println("Couldn't find Z3 java API. No libz3java in java.library.path")
+          new Failure(Seq.empty)
+        case e: Throwable =>
+          println("Couldn't verify package due to internal error:")
+          println(e.getMessage())
+          new Failure(Seq.empty)
```
jcp19, line 44: Is this case required to make the pattern matching in the `catch` block exhaustive? If it is not required, I would drop it and preserve the old behavior for when an exception is thrown. If it is necessary, I would just re-throw the exception without adding additional text

jcp19 (commented):
I have a few minor comments. After addressing them, we can merge it

src/main/scala/viper/gobra/backend/Silicon.scala
Dspil: The error is thrown at the backend.start() line because it is there where the Z3 process starts.

PR #676, opened 2023-09-21
Improvements in maps

jcp19:
This PR extends the `TypeEncoding` with a method `triggerExpr` that is now used to encode expressions that occur in triggers (instead of encoding them with `expression`). This allows us to select well-formed triggers whenever the `expression` encoding generates ill-formed triggers. With this, we solve a long-standing issue where triggers containing a a map (like `m[i]` or `i in range(m)`) are translated to invalid triggers at the Viper level.

Besides, this PR introduces two quality of life improvements
- like in Viper, we now support the idiom `key in m`, where `m` is a map, whereas before we would have had to write `key in domain(m)`. The previous idiom is still supported.
- the `assert` method in `ViperWriter` has been simplified to not generate a call to `assert` whenever the condition is `TrueLit`. Instead, it simplifies to the expression. This makes the resulting Viper code look much cleaner.

In the future, we may re-organize the map encoding as done in https://github.com/viperproject/gobra/pull/606.

TODO:
- [x] doc
- [x] use default encoding for the fallback case instead of having it in the type combiner
- [x] test with the slayers/path package in SCION

### DIFF
--- a/src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
+++ b/src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
@@ -233,7 +233,11 @@ trait TypeEncoding extends Generator {
     case in.Conversion(t2, expr :: t) if typ(ctx).isDefinedAt(t) && typ(ctx).isDefinedAt(t2) => ctx.expression(expr)
   }
 
-  // TODO: doc
+  /**
+    * Encodes expressions when they occur as the top-level expression in a trigger. The default implements
+    * an encoding for predicate instances that and defers the encoding of all expressions occurring in
+    * a trigger to the expression encoding.
+    */
   def triggerExpr(@unused ctx: Context): in.TriggerExpr ==> CodeWriter[vpr.Exp] = PartialFunction.empty
 
   /**
--- a/src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
+++ b/src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
@@ -233,6 +233,22 @@ trait TypeEncoding extends Generator {
     case in.Conversion(t2, expr :: t) if typ(ctx).isDefinedAt(t) && typ(ctx).isDefinedAt(t2) => ctx.expression(expr)
   }
 
+  // TODO: doc
+  // TODO: enable consistency checks on triggers when --checkConsistency
+  def triggerExpr(ctx: Context): in.TriggerExpr ==> CodeWriter[vpr.Exp] = PartialFunction.empty
+
+  /*
+  {
+    // use predicate access encoding but then take just the predicate access, i.e. remove `acc` and the permission amount:
+    case in.Accessible.Predicate(op) =>
+      for {
+        v <- ctx.assertion(in.Access(in.Accessible.Predicate(op), in.FullPerm(op.info))(op.info))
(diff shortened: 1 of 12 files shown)


src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
```diff
@@ -233,6 +233,13 @@ trait TypeEncoding extends Generator {
     case in.Conversion(t2, expr :: t) if typ(ctx).isDefinedAt(t) && typ(ctx).isDefinedAt(t2) => ctx.expression(expr)
   }
 
+  /**
+    * Encodes expressions when they occur as the top-level expression in a trigger. The default implements
+    * an encoding for predicate instances that and defers the encoding of all expressions occurring in
+    * a trigger to the expression encoding.
```
Felalolf, line 239: some words seem missing. also the sentence is very long, maybe try to cut it.

src/main/scala/viper/gobra/translator/encodings/defaults/DefaultTriggerExprEncoding.scala
```diff
@@ -0,0 +1,27 @@
...
+import viper.gobra.ast.{internal => in}
+import viper.gobra.translator.context.Context
+import viper.gobra.translator.encodings.combinators.Encoding
+import viper.silver.{ast => vpr}
+
+class DefaultTriggerExprEncoding extends Encoding {
+  import viper.gobra.translator.util.ViperWriter._
+
+  override def triggerExpr(ctx: Context): in.TriggerExpr ==> CodeWriter[vpr.Exp] = {
+    // use predicate access encoding but then take just the predicate access, i.e. remove `acc` and the permission amount:
```
Felalolf, line 19:
```suggestion
    // use predicate access encoding but then take just the predicate access, i.e. without the access predicate:
```

src/main/scala/viper/gobra/translator/encodings/maps/MapEncoding.scala
```diff
@@ -111,6 +127,50 @@ class MapEncoding extends LeafTypeEncoding {
     }
   }
 
+  /**
+    * Encodes expressions when they occur as the top-level expression in a trigger.
+    * Notice that if we had used the expression encoding for the following triggers,
+    * they would have resulted in ill-formed triggers at the Viper level (e.g., because
+    * they would have ternary operations).
```
Felalolf, line 134:
```suggestion
    * Encodes expressions when they occur as the top-level expression in a trigger.
    * Notice that using the expression encoding for the following triggers,
    * results in ill-formed triggers at the Viper level (e.g., because
    * they have ternary operations).
```

src/main/scala/viper/gobra/translator/util/ViperWriter.scala
```diff
@@ -399,8 +399,12 @@ object ViperWriter {
     /* Can be used in expressions. */
     def assert(cond: vpr.Exp, exp: vpr.Exp, reasonT: (Source.Verifier.Info, ErrorReason) => VerificationError)(ctx: Context): Writer[vpr.Exp] = {
       // In the future, this might do something more sophisticated
-      val (res, errT) = ctx.condition.assert(cond, exp, reasonT)
-      errorT(errT).map(_ => res)
+      if (cond.isInstanceOf[vpr.TrueLit]) {
```
Felalolf, line 402: use match

Felalolf (approved):
Please do the changes, then you can merge

PR #694, opened 2023-10-26
support nested dicts, maps and dicts-maps in options

Dspil:
The new two tests added in this PR were reported by Markus as not type-checking properly.
The problem was with the nested dicts where with a dict `m`, the expression `m[1][1]` would not typecheck since `m[1]` instead of type `MathMapT` has the `InternalSingleMulti` type.
Similar problem existed with options.


jcp19 (commented):
I am not sure if adding all these case distinctions is the way to go because of all the redundancy that it introduces. I think that a more maintainable way is to use `Single` like I did in https://github.com/viperproject/gobra/pull/699 (I was gonna add this solution as a code suggestion, but it actually became harder to explain that way :D)

PR #696, opened 2023-10-26
Fixes #695

### LINKED ISSUE #695: Exception when sending a constant to a channel with a defined channel type
The following program causes a violation during encoding:
```
type ChannelMsgType int

const ChannelMsg ChannelMsgType = 42

pred sendInvariant(v ChannelMsgType) {
    true
}

requires acc(channel.SendChannel(), 1/2)
requires channel.SendGivenPerm() == sendInvariant!<_!>
requires channel.SendGotPerm() == PredTrue!<!>
func sendMsg(channel chan ChannelMsgType) {
    fold sendInvariant!<_!>(ChannelMsg)
    fold PredTrue!<!>()
    assert acc(channel.SendChannel(), _) && sendInvariant!<_!>(ChannelMsg)
    assert acc(channel.SendChannel(), _) && channel.SendGivenPerm()(ChannelMsg)
    channel <- ChannelMsg
}
```

The channel encoding checks whether the message type is identical to the channel type. This is however not the case here because constant propagation replaces `ChannelMsg` with `42`, i.e., an integer literal.

ArquintL:
This PR fixes issue #695.

We drop the violation in the channel encoding as the type-checker is anyway more lenient and allows messages of types that are *assignable* (and not necessarily equal) to the channel type.
However, this change requires a small change to the channel encoding as we can no longer use the message's type when constructing the pure method calls to the channel's invariant getters (`SendGivenPerm`, `SendGotPerm`, `RecvGivenPerm`, and `RecvGotPerm`) but have to use the channel type instead (as they are not necessarily equal anymore).

### DIFF
--- a/src/main/scala/viper/gobra/translator/encodings/channels/ChannelEncoding.scala
+++ b/src/main/scala/viper/gobra/translator/encodings/channels/ChannelEncoding.scala
@@ -67,7 +67,7 @@ class ChannelEncoding extends LeafTypeEncoding {
           )
 
           // exhale [c].RecvGivenPerm()()
-          recvGivenPermInst = getChannelInvariantAccess(channel, recvGivenPerm, Vector())(exp.info)
+          recvGivenPermInst = getChannelInvariantAccess(channel, recvGivenPerm, Vector.empty, Vector.empty)(exp.info)
           vprRecvGivenPermInst <- ctx.assertion(recvGivenPermInst)
           _ <- exhale(vprRecvGivenPermInst,
             (info, _) => ChannelReceiveError(info) dueTo InsufficientPermissionFromTagError(s"${channel.info.tag}.RecvGivenPerm()()")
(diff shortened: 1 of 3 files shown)


ArquintL:
I've observed a non-deterministic behavior of test case `same_package/import-fail01`. E.g., it failed in [this](https://github.com/viperproject/gobra/actions/runs/6661030283/job/18103206957) CI run.
However, I double-checked the corresponding Viper encoding, which does not change due to this PR's code changes.

src/main/scala/viper/gobra/translator/encodings/channels/ChannelEncoding.scala
```diff
@@ -68,7 +67,7 @@ class ChannelEncoding extends LeafTypeEncoding {
           )
 
           // exhale [c].RecvGivenPerm()()
-          recvGivenPermInst = getChannelInvariantAccess(channel, recvGivenPerm, Vector())(exp.info)
+          recvGivenPermInst = getChannelInvariantAccess(channel, recvGivenPerm, Vector.empty, Vector.empty)(exp.info)
```
jcp19, line 70: using `Vector.empty` instead of `Vector()` sounds like a good idea! I wonder, does it actually save an allocation, or is there any specific optimisation for calling the constructor with an empty list?

jcp19 (approved):

src/main/scala/viper/gobra/translator/encodings/channels/ChannelEncoding.scala
ArquintL: I've actually no idea. My motivation was to slightly help readability

PR #699, opened 2023-10-28
Treat values of type InternalSingleMulti(t, ms) as t for nested dictionaries

jcp19:
Our treatment of expressions with type `InternalSingleMulti(t, ms)` is very iffy atm, leading to issues like those reported in #694. This PR addresses the issue in part. It allows progress on VerifiedSCION to continue, but I recognize that extending this solution to the entire type system is cumbersome. Ideally, we would have a more systematic way to deal with expressions of this type.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
@@ -305,49 +305,53 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
         val baseType = exprType(base)
         val idxType  = exprType(index)
         (underlyingType(baseType), underlyingType(idxType)) match {
-          case (ArrayT(l, _), IntT(_)) =>
-            val idxOpt = intConstantEval(index)
-            error(n, s"index $index is out of bounds", !idxOpt.forall(i => i >= 0 && i < l))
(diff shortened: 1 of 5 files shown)


src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -305,49 +305,53 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
+              // Assignability in Go is a property between a value and and a type. In Gobra, we model this as a relation
+              // between two types, which is less precise. Because of this limitation, and with the goal of handling
+              // untyped literals, we introduce an extra condition here. This makes the type checker of Gobra accept Go
+              // expressions that are not accepted by the compiler.
+              val assignableToIdxType = error(n, s"$idxType is not assignable to map key of $key", !assignableTo(idxType, key))
+              if (assignableToIdxType.nonEmpty) {
+                error(n, s"$underlyingIdxType is not assignable to map key of $key", !assignableTo(underlyingIdxType, key))
+              } else {
+                assignableToIdxType
+              }
```
jcp19, line 351: This code is unchanged, I only indented it one level

jcp19:
@mlimbeck this should solve the problems you are having with dicts

PR #708, opened 2023-12-01
Add option to disable set axiomatization

dnezam:
Adds a flag to disable set axiomatization by passing a custom set axiomatization file without any axioms to Silicon.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Config.scala
+++ b/src/main/scala/viper/gobra/frontend/Config.scala
@@ -72,6 +72,7 @@ object ConfigDefaults {
   lazy val DefaultNoStreamErrors: Boolean = false
   lazy val DefaultParseAndTypeCheckMode: TaskManagerMode = TaskManagerMode.Parallel
   lazy val DefaultRequireTriggers: Boolean = false
+  lazy val DefaultDisableSetAxiomatization: Boolean = false
 }
 
 // More-complete exhale modes
@@ -137,6 +138,7 @@ case class Config(
                    parseAndTypeCheckMode: TaskManagerMode = ConfigDefaults.DefaultParseAndTypeCheckMode,
                    // when enabled, all quantifiers without triggers are rejected
                    requireTriggers: Boolean = ConfigDefaults.DefaultRequireTriggers,
+                   disableSetAxiomatization: Boolean = ConfigDefaults.DefaultDisableSetAxiomatization,
 ) {
 
   def merge(other: Config): Config = {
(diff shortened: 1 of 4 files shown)


src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -769,6 +780,18 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
       Right(())
     }
   }
+  
+  // `disableSetAxiomatization` can only be provided when using a silicon-based backend
+  // since, at the time of writing, we rely on Silicon's setAxiomatizationFile for the
+  // implementation
```
jcp19, line 786: assuming that the axiomatisations in carbon define the same symbols, why can't we use this approach with carbon?

dnezam: I don't think Carbon has an equivalent to this flag. Instead, the set axiomatization seems to be [hardcoded](https://github.com/viperproject/carbon/blob/2e12ccdd636f8e51de2d1719338613dd98bc6dd5/src/main/scala/viper/carbon/modules/impls/sequence_axioms/SetAxiomatization.scala). Additionally, it appears to be written in Boogie (instead of Viper), and also contains axiomatizations beyond sets - namely multisets.

jcp19 (commented):
This can be merged after we introduce a test that shows that the feature is working. Ideally, we would have some function that operates on sets that verifies when running "normal" Gobra and that fails with the new flag. For the CI to go through, we need a test annotation in the assertion or postcondition that is supposed to fail (you can find how to do this by looking at our regressions).

dnezam:
I think check-license-headers fails because src/main/resources/noaxioms/sets.vpr is missing a copyright header. I adapted [Silicon's src/main/resources/dafny_axioms/sets.vpr](https://github.com/viperproject/silicon/blob/cef593b9892e5f9bd062412cb35e4dfe2159357e/src/main/resources/dafny_axioms/sets.vpr) which doesn't have one either.

For the newly added test, I went with the same copyright header as the rest of the files in the directory.

jcp19:
> I think check-license-headers fails because src/main/resources/noaxioms/sets.vpr is missing a copyright header. I adapted [Silicon's src/main/resources/dafny_axioms/sets.vpr](https://github.com/viperproject/silicon/blob/cef593b9892e5f9bd062412cb35e4dfe2159357e/src/main/resources/dafny_axioms/sets.vpr) which doesn't have one either.
> 
> For the newly added test, I went with the same copyright header as the rest of the files in the directory.

Can you add the same header that we add on every file? From our perspective, this is totally fine. After this, the pr is ready to merge, assuming that the tests go through

jcp19 (approved):

PR #709, opened 2023-12-01
Flag for non-linear integer arithmetic

Dspil:
Introduces a flag to disable non-linear integer arithmetic.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Config.scala
+++ b/src/main/scala/viper/gobra/frontend/Config.scala
@@ -654,7 +654,7 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
 
   val disableNL: ScallopOption[Boolean] = opt[Boolean](
     name = "disableNL",
-    descr = "Disable non-linear integer arithmetics. Non compatible using Z3 via API or Carbon",
+    descr = "Disable non-linear integer arithmetics. Non compatible with using Z3 via API or Carbon",
     default = Some(ConfigDefaults.DefaultDisableNL),
     noshort = true,
   )
--- a/src/main/scala/viper/gobra/frontend/Config.scala
+++ b/src/main/scala/viper/gobra/frontend/Config.scala
@@ -66,6 +66,7 @@ object ConfigDefaults {
   lazy val DefaultParallelizeBranches: Boolean = false
   lazy val DefaultConditionalizePermissions: Boolean = false
   lazy val DefaultZ3APIMode: Boolean = false
+  lazy val DefaultDisableNLIR: Boolean = false
   lazy val DefaultMCEMode: MCE.Mode = MCE.Enabled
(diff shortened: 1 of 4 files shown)


src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -646,6 +652,13 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
     noshort = true,
   )
 
+  val disableNL: ScallopOption[Boolean] = opt[Boolean](
+    name = "disableNL",
+    descr = "Disable non-linear integer arithmetics. Non compatible using Z3 via API or Carbon",
```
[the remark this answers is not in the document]
jcp19: Why doesn't it work with Z3 through API?

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -770,6 +783,18 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
...
+  addValidation {
+    if (disableNL.toOption.contains(true)) {
+      if (z3APIMode.toOption.contains(true) || !isSiliconBasedBackend) {
+        Left("--disableNL is not compatible with Z3 via API or Carbon")
+      } else {
+        Right(())
+      }
+    } else {
+      Right(())
+    }
```
jcp19, line 795:
```suggestion
    if (disableNL.toOption.contains(true) && !z3APIMode.toOption.contains(true) && isSiliconBasedBackend) {
      Right(())      
    } else {
      Left("--disableNL is not compatible with Z3 via API or Carbon")
    }
```

src/test/resources/regressions/examples/disableNL.gobra
```diff
@@ -0,0 +1,11 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package pkg
+
+requires 0 <= x && x <= 10
+requires 0 <= y && y <= 10
+ensures 0 <= res && res <= 100 // would fail with --disableNL
```
jcp19, line 8: Here, we can definitely add an annotation making this run with `--disableNL` and then we can add an expected post-condition failure

jcp19 (changes requested):

src/main/scala/viper/gobra/frontend/Config.scala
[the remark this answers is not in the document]
Dspil: We don't have a good way to pass these kinds of settings using the API.

src/main/scala/viper/gobra/frontend/Config.scala
jcp19: oops, you are right, should be fixed now

jcp19 (approved):
LGTM

PR #718, opened 2024-01-11
Check by default that all pure and ghost functions have termination measures

jcp19:
In the past, we run into unsoundnesses and tricky corner cases of Gobra because, so far, we were able to use non-terminating functions in specifications. This PR changes the default behaviour of Gobra to expect a termination measure in any pure function or method. Please let me know if you agree with this being the default behaviour. Notice that I did not adapt all of the current tests. This can be done but it would take more time than I am willing to spend atm.

### DIFF
--- a/src/main/resources/stubs/strings/strings.gobra
+++ b/src/main/resources/stubs/strings/strings.gobra
@@ -151,6 +151,7 @@ ensures len(elems) == 0 ==> res == ""
 ensures len(elems) == 1 ==> res == elems[0]
 // (joao) Leads to precondition of call might not hold (permission to elems[i] might not suffice)
 // ensures len(elems) > 1 ==> res == elems[0] + sep + Join(elems[1:], sep)
+decreases _
 pure func Join(elems []string, sep string) (res string) /*{
 	switch len(elems) {
 	case 0:
@@ -177,6 +178,7 @@ pure func Join(elems []string, sep string) (res string) /*{
 // HasPrefix tests whether the string s begins with prefix.
 pure
 ensures ret == (len(s) >= len(prefix) && s[0:len(prefix)] == prefix)
+decreases _
 func HasPrefix(s, prefix string) (ret bool) {
 	return len(s) >= len(prefix) && (s[0:len(prefix)] == prefix)
 }
@@ -184,6 +186,7 @@ func HasPrefix(s, prefix string) (ret bool) {
 // HasSuffix tests whether the string s ends with suffix.
 pure
 ensures ret == (len(s) >= len(suffix) && s[len(s)-len(suffix):] == suffix)
+decreases _
 func HasSuffix(s, suffix string) (ret bool) {
 	return len(s) >= len(suffix) && s[len(s)-len(suffix):] == suffix
 }
--- a/src/main/resources/stubs/strings/strings.gobra
+++ b/src/main/resources/stubs/strings/strings.gobra
@@ -178,7 +178,7 @@ pure func Join(elems []string, sep string) (res string) /*{
 // HasPrefix tests whether the string s begins with prefix.
 pure
 ensures ret == (len(s) >= len(prefix) && s[0:len(prefix)] == prefix)
-decreases _
+decreases
 func HasPrefix(s, prefix string) (ret bool) {
 	return len(s) >= len(prefix) && (s[0:len(prefix)] == prefix)
 }
@@ -186,7 +186,7 @@ func HasPrefix(s, prefix string) (ret bool) {
 // HasSuffix tests whether the string s ends with suffix.
 pure
 ensures ret == (len(s) >= len(suffix) && s[len(s)-len(suffix):] == suffix)
-decreases _
+decreases
 func HasSuffix(s, suffix string) (ret bool) {
 	return len(s) >= len(suffix) && s[len(s)-len(suffix):] == suffix
 }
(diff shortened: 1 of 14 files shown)


src/main/resources/stubs/strings/strings.gobra
```diff
@@ -177,13 +178,15 @@ pure func Join(elems []string, sep string) (res string) /*{
 // HasPrefix tests whether the string s begins with prefix.
 pure
 ensures ret == (len(s) >= len(prefix) && s[0:len(prefix)] == prefix)
+decreases _
```
ArquintL, line 181:
```suggestion
decreases
```

src/main/resources/stubs/strings/strings.gobra
```diff
@@ -177,13 +178,15 @@ pure func Join(elems []string, sep string) (res string) /*{
...
 func HasPrefix(s, prefix string) (ret bool) {
 	return len(s) >= len(prefix) && (s[0:len(prefix)] == prefix)
 }
 
 // HasSuffix tests whether the string s ends with suffix.
 pure
 ensures ret == (len(s) >= len(suffix) && s[len(s)-len(suffix):] == suffix)
+decreases _
```
ArquintL, line 189:
```suggestion
decreases
```

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -710,6 +715,13 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
     noshort = true,
   )
 
+  val disableCheckTerminationPureFns: ScallopOption[Boolean] = opt[Boolean](
+    name = "disableTermCheckPureFns",
+    descr = "Do not check that all pure functions are marked with termination measures,",
```
ArquintL, line 720:
```suggestion
    descr = "Do not enforce that all pure functions must have termination measures",
```

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -710,6 +715,13 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
     noshort = true,
   )
 
+  val disableCheckTerminationPureFns: ScallopOption[Boolean] = opt[Boolean](
+    name = "disableTermCheckPureFns",
```
ArquintL, line 719:
```suggestion
    name = "disablePureFunctsTerminationRequirement",
```

ArquintL: I'm definitely open to other suggestions but that's one I'd consider better than the current one.
Some of my arguments:
- `Term` is imho too short as one could confuse it with, e.g., a term algebra we use for our work on security protocols
- `Check` is imho confusing as enabling this option will not reduce proof obligations since pure functions with termination measures will still be checked for termination

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostMemberTyping.scala
```diff
@@ -58,13 +59,26 @@ trait GhostMemberTyping extends BaseTyping { this: TypeInfoImpl =>
...
 
+  private[typing] def wellFoundedIfNeededPureSpec(member: PMember): Messages = {
+    val spec = member match {
+      case m: PMethodDecl => m.spec
+      case f: PFunctionDecl => f.spec
+      case _ => Violation.violation("Unexpected member type")
+    }
+    val hasMeasureIfNeeded = config.disableCheckTerminationPureFns || spec.terminationMeasures.nonEmpty
+    val needsMeasureError =
+      error(member, "All pure functions must have termination measures, but none was found for this member.", !hasMeasureIfNeeded)
```
ArquintL, line 70:
```suggestion
      error(member, "All ghost methods and pure functions must have termination measures, but none was found for this member.", !hasMeasureIfNeeded)
```

ArquintL: While looking at this body, I thought that the error message only captures some of the cases but actually you're calling this function only for pure functions and methods. However, shouldn't we call this check also for ghost (impure) functions and methods?

src/main/resources/stubs/strings/strings.gobra
ArquintL: I think it would be good to be consistent, i.e., to use `decreases _` for abstract functions and `decreases` (or an appropriate non-constant measure) for non-abstract functions

ArquintL (commented):
I've left some comments

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostMemberTyping.scala
jcp19: I would be happy to make that change as well if the rest agrees. Tbh, I did not even consider that; instead, I was concerned about guaranteeing that our specifications are always meaningful, which may not be the case if they refer to non-terminating pure functions

src/main/resources/stubs/strings/strings.gobra
jcp19: I thought about using `decreases` and dismissed it because these functions are imported and re-verified in every client. It feels wasteful to reverify termination everywhere. Note that, while the same could be said about reverifying the body, including the definition of the function may be useful for clients (and, unfortunately, Gobra/Viper does not allow us to define a function without verifying its correctness).

ArquintL: I'm not sure whether we already do that but we could in principle turn decreases measures for imported pure functions into wildcard measures to avoid the additional overhead of re-verifying termination. I'd rather do this within Gobra as opposed to these stubs such that we can still verify the stubs (and thus check that they actually terminate)

ArquintL (approved):

PR #731, opened 2024-02-23
More aggressive chopper

Felalolf:
This PR adds an improved version of the chopper. For now, the changes are local to Gobra. In the future, these changes should be moved to Viper. 

## Improvement

The PR changes how dependencies for domain axioms are defined. Before, for an axiom A, we added dependencies from all uses in A to A and from A to all uses in A. This has one bug and one issue.

1) If the uses of A include the domain that A is defined in, then A and all its uses are always included when the domain is included.
2) A is always included if any expression in A is required, *even* if that expression is guarded by a trigger that is not included, i.e. never triggered.

The PR does the following changes:

1) Dependencies that have A's domain as source are removed from axiom dependencies. 
2) As before, the axiom has a dependency to all uses in its body. Conversely, all uses outside of quantified expressions and uses in triggers have a dependency to the axiom. 

Consider the following domain:
```javascript=
domain D {
  func foo(x: Int): Int
  func bar(x: Int): Int
  func baz(x: Int): Int
  axiom A { forall x: Int :: {bar(x)} bar(x) > baz(x) }
}
```

Previously, the dependencies were (omitting dependencies to D): A -> bar, A -> baz, bar -> A, baz -> A
With the PR the dependencies are instead: A -> bar, A -> baz, bar -> A


ArquintL:
Why does `LocalChopper` only add code? Doesn't this PR build on the existing chopper and modified its behavior?

PR #735, opened 2024-02-27
Fix #734

### LINKED ISSUE #734: Paranthesis omitted in toString of Internal AST
For instance, the statement `assert 0 == 5 - (2+3)` is printed as "assert 0 == 5 - 2 + 3".

Felalolf:
This PR fixes issue 734 with the simplest solution by always adding parantheses around binary expressions.
This solution results in many unnecessary parantheses. Please tell me if you think that this is not ok. In that case, the printer would require some state tracking whether a child expression may omit parantheses.

### DIFF
--- a/src/main/scala/viper/gobra/ast/internal/PrettyPrinter.scala
+++ b/src/main/scala/viper/gobra/ast/internal/PrettyPrinter.scala
@@ -580,7 +580,7 @@ class DefaultPrettyPrinter extends PrettyPrinter with kiama.output.PrettyPrinter
     case StructUpdate(base, field, newVal) => showExpr(base) <> brackets(showField(field) <+> ":=" <+> showExpr(newVal))
     case Negation(op) => "!" <> showExpr(op)
     case BitNeg(op) => "^" <> showExpr(op)
-    case BinaryExpr(left, op, right, _) => showExpr(left) <+> op <+> showExpr(right)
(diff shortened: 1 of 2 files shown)


Felalolf:
The failing tests are the printer tests. I will change them after we agree on the parantheses.

jcp19 (approved):
This is alright, but maybe we can be a bit more parsimonious with the parentheses. Maybe, we could just emit them for a few binary expressions like Subtraction and Division, and keep it as is for addition and the other associative operators

Felalolf:
I do not think this solves the problem, e.g. `a + (b * c)` and  `(a + b) * c` are also different. On first glance, it seems that we can get rid of the paranthesis for multiplication and the equality operators.

jcp19:
> I do not think this solves the problem, e.g. `a + (b * c)` and `(a + b) * c` are also different. On first glance, it seems that we can get rid of the paranthesis for multiplication and the equality operators.

Fine by me

jcp19:
(Closed by mistake, reopened again)

You are right, my suggestion is bad. I think that we still need parens for equality: `(a == b) == c` and `a == (b == c)` would both be pretty-printed as `a == b == c` otherwise

jcp19:
Let's proceed with your original suggestion then.

jcp19 (approved):

PR #740, opened 2024-03-12
Fix issue 511

### LINKED ISSUE #511: `&&` is not short-circuiting
The following fails, even though it shouldn't
```go
type X int

requires x != nil
func (x *X)test() bool

func clientX1(x *X) {
    if x != nil && x.test() {

    }
}
```

Error:
```
[info] Precondition of call x.test() might not hold. 
[info] Assertion x != nil might not hold.
```

Dually, it is likely that `||` is also not short-circuiting, although I haven't tested it.

jcp19:
Fixes #511

### DIFF
--- a/src/test/resources/regressions/issues/000511.gobra
+++ b/src/test/resources/regressions/issues/000511.gobra
@@ -11,12 +11,18 @@ func incF(x *int) (res bool) {
 	return false
 }
 
-func clientAdd() {
+func clientAnd1() {
 	x@ := 1
 	if incF(&x) && incT(&x) {
 		assert false
 	}
 	assert x == 2
+}
+
+func clientAnd2() {
+	x@ := 1
+	if incT(&x) && incT(&x) {}
+	assert x == 3
 }
 
 preserves acc(x)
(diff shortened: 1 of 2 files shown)


src/test/resources/regressions/issues/000511.gobra
```diff
@@ -0,0 +1,34 @@
...
+
+preserves acc(x)
+ensures   *x == old(*x) + 1
+ensures !res
+func incF(x *int) (res bool) {
+	*x += 1
+	return false
+}
+
+func clientAdd() {
```
Felalolf, line 14:
```suggestion
func clientAnd() {
```

src/test/resources/regressions/issues/000511.gobra
```diff
@@ -0,0 +1,34 @@
...
+ensures   *x == old(*x) + 1
+ensures res
+func incT(x *int) (res bool) {
+	*x += 1
+	return true
+}
+
+func clientOr() {
+	x@ := 1
+	if incT(&x) || incT(&x) {}
```
Felalolf, line 32: Please, also add cases where both calls happen, thus the variable is increased twice

Felalolf (approved):

Issue #745, opened 2024-03-15
Misleading error messages for assignments and dereferences


Felalolf:
(created for PR)

In the following code, the first and second assignment fail as exhale and call error respectively:
```
func assignment() int {
    var x@, y [10]int
    exhale acc(&x)
    x = y
}

func load() int {
    var x@, y [10]int
    exhale acc(&x)
    y = x
}

```

PR #746, opened 2024-03-15
Improves Error Messages for Assignments

Felalolf:
Check #745 for an example. As shown by the added file, the error messages are not perfect: 

The following snippet causes an assignment error,
```
var @x, y int
exhale acc(&x)
y = x
```
whereas the next snippet causes a load error,
```
var @x, y [5]int
exhale acc(&x)
y = x
```

The issue is caused by how Viper reports errors. The first snippets encodes to a field access whereas the second snippets encodes to a call. The call error is swapped with a load error. The field access error is a reason and not an error, so it cannot be swapped easily. 

Solutions are:

- To encode the first snippet also as a call
- To introduce reasons that can overwrite errors

Both solutions have drawbacks. The call encoding adds more complexity, which from my experience scales poorely for expressions that are all over the place. Overwriting reasons is rather error prone. Consider the encoding of `x = y`, which encodes `x` the same as `y = x`. If we overwrite errors, either both cause the same error message, i.e. a read/load error, or we need to remove the overwriting reasons somehow in one of the cases.

Long story short, in my opinion, the changes described above are not worth it. If we want these improvement, I am in favor of merging this PR and creating a new issue.

### DIFF
--- a/src/main/scala/viper/gobra/reporting/DefaultErrorBackTranslator.scala
+++ b/src/main/scala/viper/gobra/reporting/DefaultErrorBackTranslator.scala
@@ -6,8 +6,7 @@
 
 package viper.gobra.reporting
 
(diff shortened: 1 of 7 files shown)


jcp19 (approved):
LGTM

PR #749, opened 2024-03-21
Add support for backend annotations on functions and methods

jcp19:
With this PR, users can now attach annotations for the backend on functions, methods, outline blocks, and closures. I picked the syntax `#backend[...]`, where `...` is a comma-separated list of Viper annotations without the `@` prefix.


src/main/scala/viper/gobra/translator/util/VprInfo.scala
```diff
@@ -28,4 +29,19 @@ object VprInfo {
     val annotation = vpr.AnnotationInfo(Map(key -> values))
     vpr.ConsInfo(annotation, info)
   }
+
+  private def annotationToInfo(a: BackendAnnotation): vpr.AnnotationInfo = {
```
Felalolf, line 33: "annotation" feels a bit overloaded because we also already have the Annotations for error backtranslation. How about using the more verbose "backendAnnotation", also for the other added methods in this object.

Felalolf (commented):
Please change the position of the annotation.

Felalolf (approved):

PR #751, opened 2024-03-24
Viper data collection changes

Simon-Hostettler:
This PR contains changes to facilitate submissions to the [Viper data collection server](https://github.com/viperproject/viper-data-collection).

- Added a `submitForEvaluation` flag to the config
- Added submitters to `GoVerifier`

### DIFF
--- a/src/main/scala/viper/gobra/Gobra.scala
+++ b/src/main/scala/viper/gobra/Gobra.scala
@@ -17,6 +17,7 @@ import scalaz.Scalaz.futureInstance
 import viper.gobra.ast.internal.Program
 import viper.gobra.ast.internal.transform.{CGEdgesTerminationTransform, ConstantPropagation, InternalTransform, OverflowChecksTransform}
 import viper.gobra.backend.BackendVerifier
+import viper.gobra.backend.ViperBackends.{CarbonBackend, SiliconBackend}
 import viper.gobra.frontend.PackageResolver.{AbstractPackage, RegularPackage}
 import viper.gobra.frontend.Parser.ParseResult
 import viper.gobra.frontend.info.{Info, TypeInfo}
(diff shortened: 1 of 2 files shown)


src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -82,9 +89,18 @@ trait GoVerifier extends StrictLogging {
...
+    val viperBackendName: String = config.backend match {
+      case CarbonBackend => "Carbon"
+      case SiliconBackend => "Silicon"
+    }
+
     val timeFormatter = DateTimeFormatter.ofPattern("HH:mm:ss");
     config.packageInfoInputMap.keys.foreach(pkgInfo => {
       val pkgId = pkgInfo.id
+
+      //adds a submitter for this package to submitters if config.submitForEvaluation is set, key is pkgId
```
jcp19, line 101: this sounds like the wrong place for this comment. Sth like this should be in `createSubmitter` and and the description of what we store in keys/values should be on the declaration of `submitters`

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -143,6 +165,13 @@ trait GoVerifier extends StrictLogging {
     if (allErrors.isEmpty) VerifierResult.Success else VerifierResult.Failure(allErrors)
   }
 
+  private def createSubmitter(allowSubmission: Boolean, id: String, verifier: String): Unit = {
```
jcp19, line 168: we should make the signature of `createSubmitter` by having `verifier` of type `ViperBackend`. We could obtain a the name using the `name` field that I suggested.

jcp19, line 168: maybe a better name would be `registerSubmitter`

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -143,6 +165,13 @@ trait GoVerifier extends StrictLogging {
...
+    if (allowSubmission) {
+      val submitter = new ManualProgramSubmitter(true, "", "Gobra", verifier, Array())
```
jcp19, line 170: are we actually submitting anything if the second parameter is always `""`?

jcp19: Btw, I find it a bit odd that we have a flag `ProgramSubmitter` has a field `allowSubmission`. Maybe I misunderstand its purpose, but if this object was registered, it should be submitted. Otherwise, we should not register it. I don't see the need for this field.

jcp19 (changes requested):

jcp19:
@Dspil is this still relevant, or shall I close it?

Dspil:
It is, just not very high in my todo list.

PR #747, opened 2024-03-20
Support for Ghost Pointers

ArquintL:
This PR adds support for ghost pointers. However, ghost fields will be added by a separate PR. Note that this PR does not address the issue of calling ghost functions in non-ghost code, which is instead part of PR #755.

PR #755, opened 2024-03-28
Disallows impure non-ghost calls in ghost code

### LINKED ISSUE #37: Disallow use of ghost members in non-ghost code

### LINKED ISSUE #420: non ghost assignment in ghost if-else block
Gobra verifies ghost if-else blocks that assign to non-ghost variables.
An example that verifies:
```
func foo() int {
    ghost bar := true
    var baz int
    ghost if bar {
        baz = 1
    } else {
        baz = 0
    }
    return baz
}
```

Configuration:
Gobra: 63ea747f56f7280106e349c0cd8dd51b2afb360d
silver: 7228e7144d41c91f02a70a68a93b6b3efae57d14
silicon: 22551b40509c381991a493da0108ea4c97fd602d
carbon: 4393d154a5ae24d994a0c2c578374bdd49c3a3b3
viperserver: 5907ce1744501b7949d25cc0d5356145431ab6f7

ArquintL:
This PR depends on PR #747. Thus, this PR targets `impure-calls-in-ghost-code-target-branch` (which corresponds to the branch used in #747) for review-purposes.

This PR checks that ghost code does not contain calls to non-ghost impure functions. Note that calling non-ghost pure functions is fine due to the absence of side effects.
To support implementation proofs, this PR makes interface & closure implementation proofs non-ghost. Technically, we would need to distinguish between implementation proofs for a ghost or non-ghost member but due to the syntax restrictions for implementation proofs, we do not have to make this distinction (e.g., assignments are disallowed).

Fixes #37, Fixes #420


src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
```diff
@@ -217,4 +224,238 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
...
+
+  // the body can only contain fold, unfold, and the call
+  private[typing] def implementationProofBodyHasRightShape(body: PBlock,
+                                                          isExpectedCall: PInvoke => Boolean,
+                                                          expectedCall: String,
+                                                          result: PResult): PropertyResult = {
+    val expectedResults = result.outs.flatMap(nameFromParam).map(t => PNamedOperand(PIdnUse(t)))
+
+    def isExpectedAssignment(ass: PAssignment): Boolean = ass match {
+      case PAssignment(Vector(i: PInvoke), left) if isExpectedCall(i) && expectedResults == left => true
```
jcp19, line 349:
```suggestion
      case PAssignment(Vector(i: PInvoke), left) => isExpectedCall(i) && expectedResults == left
```

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GhostWellDef.scala
```diff
@@ -134,24 +125,38 @@ trait GhostWellDef { this: TypeInfoImpl =>
...
-    case _: PNew => noMessages
-
-    case n@PMake(_, args) => error(
-      n,
-      "ghost error: make expressions may not contain ghost expressions",
-      args exists (x => !noGhostPropagationFromChildren(x))
-    )
+    case e: PNew =>
+      if (isEnclosingGhost(e)) {
+        Violation.violation(exprType(e).isInstanceOf[GhostPointerT], s"All memory allocated within ghost code must be located on the ghost heap")
```
jcp19, line 145: I think the following is more clear
```suggestion
        Violation.violation(exprType(e).isInstanceOf[GhostPointerT], s"Cannot allocate non-ghost memory in ghost code.")
```

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GhostWellDef.scala
```diff
@@ -134,24 +125,38 @@ trait GhostWellDef { this: TypeInfoImpl =>
...
+      } else {
+        Violation.violation(exprType(e).isInstanceOf[ActualPointerT], s"All memory allocated within actual code must be located on the actual heap")
```
jcp19, line 147: we can use a similar message here

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GhostWellDef.scala
```diff
@@ -134,24 +125,38 @@ trait GhostWellDef { this: TypeInfoImpl =>
...
+      noMessages
+
+    case e: PMake => (e, isEnclosingGhost(e)) match {
+      case (PMake(_: PGhostSliceType, _), true)  => noMessages
+      case (_, true) => error(e, "Allocating memory within ghost code is forbidden")
+      case (PMake(_, args), false) => error(
+        e,
+        "ghost error: make expressions may not contain ghost expressions",
+        args exists (x => !noGhostPropagationFromChildren(x))
+      )
```
jcp19, line 158: I wonder if we need a similar treatment for all built-in functions

jcp19 (commented):
I just have a few minor comments

src/test/resources/regressions/features/closures/closures-fail5-proofs.gobra
```diff
@@ -66,8 +66,7 @@ func test4() {
 
     //:: ExpectedOutput(type_error)
     proof cl implements pspec {
-        //:: ExpectedOutput(type_error)
-        x = 2 // assignment in ghost code to a non-ghost variable
+        x = 2
```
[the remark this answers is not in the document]
ArquintL: The entire implementation proof used to be ghost. I changed this such that the block is non-ghost as otherwise we couldn't call a non-ghost function as part of the implementation proof. As a consequence, the assignment does no longer cause a ghost error. However, the implementation proof itself still fails as this assignment is not permitted due to the restrictions imposed on statements occurring in an implementation proof

src/main/scala/viper/gobra/frontend/info/implementation/typing/MiscTyping.scala
```diff
@@ -41,6 +41,77 @@ trait MiscTyping extends BaseTyping { this: TypeInfoImpl =>
...
+              failedProp(
+                s"implementation proof and interface member have a different signature (should be '$specSig', but is $implSig nad ${implSig == specSig})",
+                cond = !identicalTypes(implSig, specSig)
+              )
+            }
+            // check that pure annotations match
+            val matchingPure = failedProp(
+              s"The pure annotation does not match with the pure annotation of the interface member",
+              cond = n.isPure != spec.isPure
+            )
```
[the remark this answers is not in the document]
ArquintL: This is checked in `Implements.scala` but I've added a testcase for this

jcp19 (approved):
LGTM

src/test/resources/regressions/features/ghost_pointer/ghost-pointer-receiver-fail01.gobra
```diff
@@ -0,0 +1,149 @@
...
+decreases
+func (r Foo) ActualNonPointerReceiver() int {
+    r.field = 42
+    return r.field
+}
+
+decreases
+preserves acc(r)
+//:: ExpectedOutput(type_error)
+func (r gpointer[Foo]) ActualGhostPointerReceiver() int { // ghostpointer is only allowed for ghost methods
```
Felalolf, line 138: gpointer is not allowed as the receiver, right? It could be a standard argument?

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GhostWellDef.scala
```diff
@@ -134,24 +125,38 @@ trait GhostWellDef { this: TypeInfoImpl =>
...
-      n,
-      "ghost error: make expressions may not contain ghost expressions",
-      args exists (x => !noGhostPropagationFromChildren(x))
-    )
+    case e: PNew =>
+      if (isEnclosingGhost(e)) {
+        Violation.violation(exprType(e).isInstanceOf[GhostPointerT], s"Cannot allocate non-ghost memory in ghost code.")
+      } else {
+        Violation.violation(exprType(e).isInstanceOf[ActualPointerT], s"Cannot allocate ghost memory in non-ghost code.")
+      }
```
Felalolf, line 148: Just to make sure, this is checked by the type checker beforehand, right? I.e. if I allocate non-ghost memory in ghost code, then I will get an error and not an exception?

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GhostWellDef.scala
```diff
@@ -134,24 +125,38 @@ trait GhostWellDef { this: TypeInfoImpl =>
...
+      noMessages
+
+    case e: PMake => (e, isEnclosingGhost(e)) match {
+      case (PMake(_: PGhostSliceType, _), true)  => noMessages
+      case (_, true) => error(e, "Allocating memory within ghost code is forbidden")
```
Felalolf, line 153:
```suggestion
      case (_, true) => error(e, "Allocating memory with make within ghost code is forbidden")
```

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GhostWellDef.scala
```diff
@@ -102,15 +102,6 @@ trait GhostWellDef { this: TypeInfoImpl =>
 
   private def exprGhostSeparation(expr: PExpression): Messages = expr match {
     case _: PGhostExpression => noMessages
```
Felalolf, line 104: As a sanity check, is the following example rejected?

```
requires acc(x)
func foo(x *int) {
  *x = 0
}

ghost 
requires acc(x)
func bar(x *int) {
  ghost foo(x)
}
```

Felalolf (approved):
I have 1-2 questions. If these questions do not show some issue, then you can merge.

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GhostWellDef.scala
ArquintL: It is and I've added it as another testcase. In particular, we have an assignment of a ghost variable to a non-ghost parameter and a call of a non-ghost method in a ghost context

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GhostWellDef.scala
ArquintL: The type-checker makes sure that `PNew` is typed correctly such that the violations should not occur. I.e., if you use `new` in ghost code, you're creating a ghost pointer while the use of `new` in non-ghost code returns a regular pointer

src/test/resources/regressions/features/ghost_pointer/ghost-pointer-receiver-fail01.gobra
ArquintL: Yes and yes. The parameter would however be explicitly marked as a ghost parameter.

PR #768, opened 2024-05-19
Fix purity classification of composite literals

jcp19:
Currently, the type system of Gobra disallows array literals like `[3]int{}` from occurring in strongly pure contexts, even though an array is a value type (similarly to structs). Furthermore, the type system (wrongly) allows slice and map literals in pure contexts, even though it shouldn't because slices and maps are reference types. This PR addresses this classification.

To implement this, I had to fix the function `underlyingTypeWithCtxP` to allow for ADTs

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
@@ -469,7 +469,7 @@ trait GhostExprTyping extends BaseTyping { this: TypeInfoImpl =>
         case _: PImplicitSizeArrayType => true
         case t: PType => underlyingTypeP(t) match {
           case Some(ut) => ut match {
-            case _: PSliceType | _: PMapType => false
+            case _: PSliceType | _: PGhostSliceType | _: PMapType => false
             case _ => true
           }
           case None => false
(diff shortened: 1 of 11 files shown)


src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
```diff
@@ -466,8 +466,14 @@ trait GhostExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
 
       case PCompositeLit(typ, _) => typ match {
-        case _: PArrayType | _: PImplicitSizeArrayType => !strong
-        case _ => true
+        case _: PImplicitSizeArrayType => true
+        case t: PType => underlyingTypeP(t) match {
+          case Some(ut) => ut match {
+            case _: PSliceType | _: PGhostSliceType | _: PMapType => false
+            case _ => true
+          }
```
ArquintL, line 474: I think I'd prefer an allow list instead of a block list. In particular, the ghost pointer PR adds `PGhostPointerType`, which implements `PGhostLiteralType`. One option I could see is to add a `PValueType` and `PReferenceType` in the PAST such that we can more conveniently case split here

ArquintL (changes requested):

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
jcp19: Given that "value type" and "reference type" are not used in the language spec ("reference types" were used until 2013 and then dropped - https://github.com/golang/go/commit/b34f0551387fcf043d65cd7d96a0214956578f94), I will refrain from labeling our types as such in the implementation. I added a case analysis which lists all acceptable literal types. While very explicit, this list is still reasonably sized and it is unlikely to grow a lot in the future

PR #776, opened 2024-07-01
Add refute statement

bruggerl:
This PR adds support for `refute` statements.


src/main/scala/viper/gobra/reporting/VerifierError.scala
```diff
@@ -11,7 +11,10 @@ import viper.gobra.ast.frontend.{PReceive, PSendStmt}
 import viper.gobra.reporting.Source.Verifier
 import viper.gobra.util.Constants
 import viper.gobra.util.Violation.violation
-import viper.silver.ast.SourcePosition
+import viper.silver.ast.{FalseLit, SourcePosition}
+import viper.silver.plugin.standard.refute.{Refute, RefuteError, RefuteErrorReason}
+import viper.silver.plugin.standard.smoke.SmokeDetectionInfo
+import viper.silver.verifier.{ErrorMessage, ErrorReason, errors, reasons}
```
jcp19, line 17: Do we need all of these imports? In particular, we do not seem to be using `FalseLit` or `SmokeDetectionInfo`

src/main/scala/viper/gobra/translator/transformers/RefuteTransformer.scala
```diff
@@ -0,0 +1,38 @@
...
+import viper.silver.{ast => vpr}
+
+class RefuteTransformer extends ViperTransformer {
+  override def transform(task: BackendVerifier.Task): Either[Seq[AbstractError], BackendVerifier.Task] = {
+    for {
+      transformedProg <- executeTerminationPlugin(task)
+    } yield transformedProg
+  }
+
+  private def executeTerminationPlugin(task: BackendVerifier.Task): Either[Seq[AbstractError], BackendVerifier.Task] = {
```
jcp19, line 22: Let's rename this to `executeRefutePlugin`

src/main/scala/viper/gobra/translator/transformers/RefuteTransformer.scala
```diff
@@ -0,0 +1,38 @@
...
+import viper.silver.plugin.SilverPlugin
+import viper.silver.plugin.standard.refute.RefutePlugin
+import viper.silver.verifier.AbstractError
+import viper.silver.{ast => vpr}
+
+class RefuteTransformer extends ViperTransformer {
+  override def transform(task: BackendVerifier.Task): Either[Seq[AbstractError], BackendVerifier.Task] = {
+    for {
+      transformedProg <- executeTerminationPlugin(task)
+    } yield transformedProg
```
jcp19, line 19: This thing is the same as 
```suggestion
    executeTerminationPlugin(task)
```
which makes me wonder if we need to have the logic on a separate function. Maybe we should just inline `executeTerminationPlugin` (a.k.a. `executeRefutePlugin`) here.

src/main/scala/viper/gobra/backend/BackendVerifier.scala
```diff
@@ -78,22 +80,31 @@ object BackendVerifier {
...
+        Failure(verificationError.toVector, backTrackInfo)
+    }
+
+
+  def convertVerificationResult(program: Program, result: VerificationResult, backTrackInfo: BackTrackInfo): Result = {
+    val refutePlugin = new RefutePlugin(null, null, null, null)
+    val transformedResult = refutePlugin.mapVerificationResult(program, result)
+
+    convertVerificationResult(transformedResult, backTrackInfo)
   }
```
jcp19, line 108: are you sure we need this explicit call to `mapVerificationResult`? I believe this is implicitly called already by the plugin machinery (IIRC, we do not add explicit calls to `mapVerificationResult` for the transformation plugin)

jcp19 (changes requested):
I think the structure is mostly ok, but please check my comments. This PR is missing tests though. I would add a few to folder `src/test/resources/regressions/features/refute` (which needs to be created). We usually write multiple types of tests:
- type-checking tests, e.g., what happens when we write `refute 1`? Does Gobra provide an error in the correct line?
- successful verification tests, i.e., tests that use the feature and should succeed.
- failed verification tests - these should indicate what is the expected verification error. 

To annotate the expected type errors and verification errors, we use an annotation system that is exemplified in https://github.com/viperproject/gobra/tree/master/src/test/resources/regressions/features/opaque. Essentially, we use `//:: ExpectedOutput(type_error)` in the line before the expected type error and `//:: ExpectedOutput(error:cause)`, where `error` and `cause` are replaced by the ids of the error kind and cause that you added

src/main/scala/viper/gobra/reporting/VerifierError.scala
bruggerl: -> [2bd1f22](https://github.com/viperproject/gobra/pull/776/commits/2bd1f22e9cf908143bae93c9a4730a6abcf094e6)

src/main/scala/viper/gobra/translator/transformers/RefuteTransformer.scala
bruggerl: -> [ed6fbf4](https://github.com/viperproject/gobra/pull/776/commits/ed6fbf44f1e9990a4c9331f60eb3ea38f1fc3c44)

src/main/scala/viper/gobra/translator/transformers/RefuteTransformer.scala
bruggerl: -> [ed6fbf4](https://github.com/viperproject/gobra/pull/776/commits/ed6fbf44f1e9990a4c9331f60eb3ea38f1fc3c44)

jcp19 (approved):
LGTM! I just have a few small nitpicks, please address them before merging.

For future reference, this PR has more cool things besides `refute`:
- It refactors Gobra to use the `FrontendAPI`, which besides other things, gives us pretty-printed Viper code that is easier to read e.g., it shows the code prior to applying the termination plugin. (This has a side effect at the moment of leading to ill-formed Viper code due to a bug in silver where decreases clauses are printed as `requires decreases ...`, but that is a bug that we need to fix in silver anyway)
- It fixes the error reporting in verification errors related to closures
- It deprecates part of the `TerminationTransformer` (we might be able to drop this entirely in the future)

PR #802, opened 2024-11-29
Hyper Gobra

jcp19:
Felix's implementation of the technique for checking SIF based on [this paper](https://pm.inf.ethz.ch/publications/WolfMueller24.pdf)


src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -68,7 +68,8 @@ object ConfigDefaults {
   val DefaultZ3APIMode: Boolean = false
   val DefaultDisableNL: Boolean = false
   val DefaultMCEMode: MCE.Mode = MCE.Enabled
-  val DefaultEnableLazyImports: Boolean = false
+  val DefaultHyperMode: Hyper.Mode = Hyper.Enabled
```
jcp19, line 71: TODO: change

src/main/scala/viper/gobra/util/Sound.scala
```diff
@@ -0,0 +1,68 @@
+package viper.gobra.util
+
+import javax.sound.sampled.{AudioFormat, AudioSystem}
```
jcp19, line 3: TODO: drop

src/main/scala/viper/gobra/frontend/Config.scala
ArquintL: It would be nice if we could leave this turned on but ensure that if a program does not make use of any SIF specs that we do not perform the product construction.
If this is not easily possible, we have to think about the `hyperMode` options as "off" currently results in turning all SIF specs to true, which is potentially dangerous if you are not aware of this. Imho, the default mode should make sure that non-SIF programs are encoded as today (i.e., without product construction) and either handle SIF programs soundly or report an error. However, having a mode that allows you to turn SIF checks off for a SIF program could be useful for debugging but this requires a warning / error that this is potentially unsound

src/main/scala/viper/gobra/translator/transformers/hyper/SIFExtendedTransformer.scala
```diff
@@ -0,0 +1,2165 @@
...
+    /** If true, add an 'assume p1' at the beginning of each method, to cut down on redundant paths the
+      * verification could consider */
+    var optimizeRestrictActVars: Boolean = true
+    /** Applications of the functions which have an entry here, will be replaced by the expression
+      * determined by the entry in the second execution. */
+    var primedFuncAppReplacements: mutable.HashMap[String, (FuncApp, Exp, Exp) => Exp] = new mutable.HashMap
+    /** Set this to only transform methods that contain relational assertions somewhere in their spec or body.
+      * May lead to invalid programs when such a method calls another methods that does contain such specs.
+      */
+    var onlyTransformMethodsWithRelationalSpecs: Boolean = false
```
jcp19, line 37: @ArquintL I guess setting this to true should allow us to keep the default for SIF. I will change it once I start working on this pR

ArquintL: That's not sufficient. Instead of checking which methods have relational spec and which do not, a first attempt would be to decide this for the entire input program. To keep this on a per-method level, we would first need to address the problem mentioned in the comment above namely that a non-SIF method calling a SIF method (or vice-versa) fails as the number of parameters / arguments does not match.
A similar issue occurs if we would mix methods encoded with optimized SIF encoding (enforcing low control flow) and the more complete SIF encoding as the latter one uses additional parameters (namely the activation variables) and also uses a different ordering for the parameters belonging to the 2nd execution (which is just an implementation detail)

jcp19:
The current version, together with the changes from #818 leads to an exception on this file:
```go
package bugs

// Bug 5.
// Trying to verify this file yields the following error message:
// 
// 		Precondition of call test1(ub) might not hold.
// 		Permission to ub[0] might not suffice.
//
// The reason can be seen in the generated Viper file: The `_termination_proof`
// method for `test0` replaces the permission amount (`write`) with `wildcard`.
// Note: In order for this bug to occur, `test1` needs to return a struct
// (at least not some primitive type), and the functions must be `pure` (and
// thus also annotated with `decreases`).
//
// This problem first occurred while trying to verify `pkg\slayers\path\epic`.

type Struct struct {}

requires acc(&ub[0])
decreases
pure func test1(ub []byte) Struct {
    return Struct{}
}

requires acc(&ub[0])
decreases
pure func test0(ub []byte) bool {
    return Struct{} == test1(ub)
}
```

Reported by @henriman

ArquintL:
Regression (just in case we are not aware of it yet): in contrast to 8facab51220570ff88f72e71e794bcb336208e47, termination measures are no longer transformed resulting in crashes of Silicon

jcp19:
So, the bug related to termination measures has been fixed (gobra no longer immediately crashes with consistency errors for programs that perform termination checking). @henriman 

 There is currently a seemingly absurd bug where ADT members are not translated to the Viper program. I will debug this one next. After this one, I guess it is a matter of cleaning up the code (sorely needed), reviewing everything, and dropping all copied files from the SIF plugin for Viper

src/main/scala/viper/gobra/translator/transformers/hyper/SIFLowGuardTransformer.scala
```diff
@@ -482,11 +482,18 @@ trait SIFLowGuardTransformer {
...
 
-  def removeLow[N <: Node](n: N): N = {
-    n.transform{
-      case n: SIFLowExp => TrueLit()(n.pos, n.info, n.errT)
-      case n: SIFLowEventExp => TrueLit()(n.pos, n.info, n.errT)
+  def removeLow[N <: Node](n: N): Either[Seq[AbstractError], N] = {
+    var errs: Seq[AbstractError] = Seq.empty
+    val transformedN = n.transform{
+      case n: SIFLowExp =>
+        errs = errs :+ ConsistencyError(s"Low expression found: ${n.exp}", n.pos)
```
jcp19, line 489: I'm not sure if this is a good solution because every package that imports `sync` to use a Mutex will use the new spec for this function, which contains low annotations. Thus, this solution forces us to always enable SIF when verifying concurrent code

ArquintL: I've also realized this after looking at the failing testcases ^^
Maybe we should distinguish between imported and non-imported SIF expressions and ignore the former while cause errors for the latter (if SIF is disabled).
What I want to avoid is any confusion as in my program contains SIF expressions but I (e.g. by mistake) configured Gobra not to do a product construction and, thus, I'm thinking I've verified SIF but actually Gobra didn't

ArquintL: I've just changed the implementation to report an error only if the resulting proof obligations will be different. In particular, if an abstract method or function has a low expression (because it's abstract or imported) then we just replace the expression by `true` and do not report an error

ArquintL:
HyperMode `extended` now uses the [SIF plugin](https://github.com/viperproject/silver-sif-extension). For consistency reasons, it would make sense to move `SIFLowGuardTransformer.scala` to the plugin such that we do not need to conditionally run Viper plugins based on which HyperMode is selected

ArquintL:
Does anyone have a clue why the CI does not run `check-license-headers` and `build-test-deploy-container (PR)` for this PR?

jcp19:
I suggest that, from now on, we implement changes to this branch by separate PRs (except to merge with master).

jcp19:
This is now failing to build because the sif extension uses different versions of silver/silicon/carbon -.- I will repair this later

jcp19:
So, I finally updated the submodules. I found the culprit of the failure to be https://github.com/viperproject/silicon/pull/927, and this was fixed in https://github.com/viperproject/silicon/pull/937. Thus, we can now stick to the master branch of the silver-sif-extension. When the CI goes through, we should merge this.

ArquintL:
The CI is currently failing due to `commitment-extended.go` and `termination-extended-simple01.gobra` that crash the SIF translation (due to my "fixes"). While I do have locally a fix, I'm hitting an issue that `goto` statements are by default unsupported by the SIF plugin, which requires enabling `enableExperimentalFeatures`

ArquintL:
This PR is currently blocked by https://github.com/viperproject/silver-sif-extension/pull/9 and https://github.com/viperproject/silver-sif-extension/pull/10

ArquintL:
Note that this branch currently uses the `arquintl-goto-encoding` instead of `master` branch of [silver-sif-extension](https://github.com/viperproject/silver-sif-extension). We could either wait until https://github.com/viperproject/silver-sif-extension/pull/10 is merged or merge now and wait until the `silver-sif-extension` submodule is updated automatically by `update-submodules.yml`

jcp19:
aren't we close to merging https://github.com/viperproject/silver-sif-extension/pull/10 too? From a quick look, it seems that this is also in the final phase of the review, and thus, I would be happy to wait until that is merged.

jcp19:
LGTM

ArquintL (approved):

PR #818, opened 2025-01-10
Reinstate the usage of the SilFrontend API and support for the refute statement

jcp19:
This PR reinstates the changes by Lea 🍔 (@bruggerl) introduced in PR https://github.com/viperproject/gobra/pull/776. To make sure that there are no feature regressions in Gobra, we should only merge this when
- [x] the pretty printer for termination measures in Viper is fixed (https://github.com/viperproject/silver/pull/831)
- [x] the chopper can deal with termination measures ~~(https://github.com/viperproject/silver/pull/832 - might not be the best solution, we shall see)~~

In addition to all the fixes listed in PR https://github.com/viperproject/gobra/pull/776, this PR now also fixes a bug where the option `--respectFunctionPrePermAmounts` (introduced in PRs and ) is not properly communicated to the backend, as prior to this PR, we called the TerminationPlugin directly, and passed a wrong SilFrontendConfig object

Depends on #819.

**Note to the reviewer:** The only non-reviewed changes introduced by this PR are in the `PluginAwareChopper` and its tests, and in the `ChopperUtil` file


src/main/scala/viper/gobra/util/PluginAwareChopper.scala
```diff
@@ -0,0 +1,54 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2011-2025 ETH Zurich.
+
+package viper.gobra.util
+
+import viper.silver.ast.utility.chopper.{ChopperLike, Cut, Edges, SCC, Vertices, ViperGraphs}
+
+// Chopper with computation of dependencies extended to support the termination plugin and other plugins
+// used by Gobra.
```
ArquintL, line 12:
```suggestion
/**
  * Chopper with support for plugins used by Gobra, in particular Viper's termination plugin.
  */
```

src/main/scala/viper/gobra/util/PluginAwareChopper.scala
```diff
@@ -0,0 +1,54 @@
...
+object PluginAwareChopper extends ChopperLike with ViperGraphs with PluginAwareEdges with Vertices with Cut with SCC
+
+trait PluginAwareEdges extends Edges {
```
ArquintL, line 15:
```suggestion
/**
  * Extends the chopper's dependency computation to add artificial dependencies for domains ending in 'WellFoundedOrder',
  * their axioms, and subexpressions.
  * These artificial dependencies ensure that these axioms are retained by the chopper, which is necessary
  * to ensure successful verification of the chopped program because these axioms define well-founded orders
  * via "bounded" and "decreasing" functions. Without these artificial dependencies, the chopper would remove
  * these axioms as the decreases measures do not explicitly depend on, i.e., call these domain functions.
  */
trait PluginAwareEdges extends Edges {
```

src/test/scala/viper/gobra/PluginAwareChopperTests.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import org.scalatest.Inside
+import org.scalatest.funsuite.AnyFunSuite
+import org.scalatest.matchers.should.Matchers
+import viper.gobra.util.PluginAwareChopper
+import viper.silver.ast
+import viper.silver.ast.utility.chopper.Penalty
+
+class PluginAwareChopperTests extends AnyFunSuite with Matchers with Inside {
+  test("Domains ending in WellFoundedOrders are kept in chopped programs") {
```
ArquintL, line 17:
```suggestion
  test("Domains ending in 'WellFoundedOrder' are kept in chopped programs") {
```

src/test/scala/viper/gobra/PluginAwareChopperTests.scala
```diff
@@ -0,0 +1,33 @@
...
+    val intVarDecl = ast.LocalVarDecl("i", ast.Int)()
+    val function = ast.Function("functionA", Seq(intVarDecl), ast.Bool, Seq.empty, Seq.empty, None)()
```
ArquintL, line 19: Small detail: I'd add a simple decreases measure, e.g. `decreases i` such that the axioms are necessary. I know this is not relevant for the current implementation but it would be cool if this test remains correct for a more advanced implementation that actually looks at decreases measures and decides, e.g., based on the expression's type what WellFoundedOrder domains & axioms are necessary

ArquintL (approved):
As suggested, I've looked at the 3 not yet reviewed files and have a minor wish if it does not take too long to implement besides 3 suggestions for a bit more documentation or typo fixes. Otherwise, looks good! Thanks for this PR! =)

PR #822, opened 2025-01-17
Add support for new interpretation of fractional perms in pure function preconditions

jcp19:
Changes done in accordance to the corresponding change in Viper (https://github.com/viperproject/silver/pull/816).

The default is still Viper's old interpretation for fractional permissions because our case studies all use the old semantics, but we should switch in the future because the new interpretation leads to more concise specs.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Config.scala
+++ b/src/main/scala/viper/gobra/frontend/Config.scala
@@ -775,11 +775,12 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
     }
   }
 
-  val respectFunctionPrePermAmounts: ScallopOption[Boolean] = opt[Boolean](
+  val respectFunctionPrePermAmounts: ScallopOption[Boolean] = toggle(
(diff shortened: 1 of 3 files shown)


Dspil (approved):

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -771,6 +778,16 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
     }
   }
 
+  val respectFunctionPrePermAmounts: ScallopOption[Boolean] = toggle(
+    name = "respectFunctionPrePermAmounts",
+    descrYes = s"Respects precise permission amounts in pure function preconditions instead of only checking read access, as done in older versions of Gobra." +
+      "This option should be used for verifying legacy projects written with the old interpretation of fractional permissions." +
+      "New projects are encouraged to not use this flag.",
```
ArquintL, line 785:
```suggestion
      "New projects are encouraged to set this flag to false.",
```

ArquintL (approved):

PR #824, opened 2025-01-21
More robust config creation

ArquintL:
Removes the violation in `getPackageInfo` within `Source` and replaces it with proper error messages. This violation lead to server crashes in Gobra-IDE if a file did not contain a package clause.


src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -33,8 +33,8 @@ import scala.concurrent.Future
 object Parser extends LazyLogging {
 
   type ParseSuccessResult = (Vector[Source], PPackage)
-  type ParseResult = Either[Vector[ParserError], ParseSuccessResult]
-  type ImportToSourceOrErrorMap = Vector[(AbstractPackage, Either[Vector[ParserError], Vector[Source]])]
+  type ParseResult = Either[Vector[VerifierError], ParseSuccessResult]
```
jcp19, line 36: I find the name a bit confusing, in that I didn't expect the parser to return `VerifierError` (as this makes it sound like sth failed during verification). Is there a reason why having `ParseError` here is cumbersome or not possible? Otherwise, is `ParseError` still used anywhere else? Because, if it isn't, then it should be deprecated

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -387,38 +389,47 @@ trait PackageAndRecursiveModeConfig extends RawConfig {
...
+      else for {
+        pkgInfo <- getPackageInfo(sources.head, projectRoot)
+      } yield pkgInfo -> sources
     }.partitionMap(identity)
-    if (errors.length == 1) Left(errors.head)
-    else if (errors.nonEmpty) Left(s"multiple errors have been found while localizing sources: ${errors.mkString(", ")}")
-    else Right(createConfig(mappings.toMap))
+    for {
+      mappings <- if (errors.nonEmpty) Left(errors.flatten) else Right(mappings)
+    } yield createConfig(mappings.toMap)
```
jcp19, line 406: nitpick: I think this piece of code becomes easier to read like this:
```suggestion
    if (errors.nonEmpty) {
        Left(errors.flatten)
    } else {
        Right(createConfig(mappings.toMap))
    }
```

src/main/scala/viper/gobra/frontend/Parser.scala
jcp19: (I will continue the review after this comment is answered)

src/main/scala/viper/gobra/frontend/Config.scala
ArquintL: I agree, thanks!

src/test/scala/viper/gobra/parsing/ParserTestFrontend.scala
```diff
@@ -10,28 +10,28 @@ import org.bitbucket.inkytonik.kiama.util.{Source, StringSource}
...
+  def parseProgram(source : String) : Either[Vector[VerifierError], PProgram] = parse(source, source => Parser.parseProgram(source))
+  def parseExp(source : String) : Either[Vector[VerifierError], PExpression] = parse(source, Parser.parseExpr)
   def parseExpOrFail(source : String) : PExpression = parseOrFail(source, Parser.parseExpr)
-  def parseStmt(source : String) : Either[Vector[ParserError], PStatement] = parse(source, Parser.parseStmt)
+  def parseStmt(source : String) : Either[Vector[VerifierError], PStatement] = parse(source, Parser.parseStmt)
   def parseStmtOrFail(source : String) : PStatement = parseOrFail(source, Parser.parseStmt)
-  def parseType(source : String) : Either[Vector[ParserError], PType] = parse(source, Parser.parseType)
+  def parseType(source : String) : Either[Vector[VerifierError], PType] = parse(source, Parser.parseType)
   def parseTypeOrFail(source : String) : PType = parseOrFail(source, Parser.parseType)
   def parseImportDecl(source: String): Vector[PImport] = parseOrFail(source, Parser.parseImportDecl)
```
jcp19, line 36: can we keep all of these as parser errors too?

src/main/scala/viper/gobra/frontend/Source.scala
```diff
@@ -51,31 +51,39 @@ object Source {
...
+          // Fallback case if the prefix is empty, for example if the directory of a FileSource is in the current directory
+          packageName
+        }
+      }
+    } yield new PackageInfo(packageId, packageName, isBuiltIn)
+  }
+
+  /**
+    * Forcefully tries to create a package info or throws an runtime exception.
+    * Only use for unit tests
```
jcp19, line 81:
```suggestion
    * Only used for unit tests
```

src/main/scala/viper/gobra/frontend/Source.scala
```diff
@@ -51,31 +51,39 @@ object Source {
...
-      if(prefix.nonEmpty) {
-        // The - is enough to unambiguously separate the prefix from the package name, since it can't occur in the package name
-        // per Go's spec (https://go.dev/ref/spec#Package_clause)
-        prefix + " - " + packageName
-      } else {
-        // Fallback case if the prefix is empty, for example if the directory of a FileSource is in the current directory
-        packageName
+  def getPackageInfo(src: Source, projectRoot: Path): Either[Vector[ParserError], PackageInfo] = {
+    for {
+      packageName <- PackageResolver.getPackageClause(src: Source).toRight({
```
jcp19, line 56: do we need the type annotation `src: Source` here?

jcp19, line 56: nitpick: for readability purposes, I would try to make this `for` block smaller, possibly moving the thing on the rhs of the first `<-` to a local var definition outside this black, as well the def of `isBuiltIn`

jcp19 (approved):

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -355,20 +355,22 @@ trait RawConfig {
   * This is for example used when parsing in-file configs.
   */
 case class NoInputModeConfig(baseConfig: BaseConfig) extends RawConfig {
-  override lazy val config: Either[String, Config] = Right(createConfig(Map.empty))
+  override lazy val config: Either[Vector[VerifierError], Config] = Right(createConfig(Map.empty))
```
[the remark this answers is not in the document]
ArquintL: The only error so far is a `ParserError` due to missing package clauses. We could make this a `ConfigError` but I felt like `ParserError` is descriptive for users

src/main/scala/viper/gobra/frontend/Source.scala
ArquintL: I don't think so. No idea how that ended up here ^^

ArquintL: agreed

src/main/scala/viper/gobra/frontend/Config.scala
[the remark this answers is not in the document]
jcp19: Sure, fine by me

PR #825, opened 2025-01-21
Fixes #823

### LINKED ISSUE #823: Issue with backend `VSWITHSILICON`
As of recently, Gobra runs into an exception whenever it verifies a file with the Viper server with silicon. This seems to be a problem caused by the interaction of the viperserver and the `PredicateInstancePlugin`. In the future, we should run the tests against all backends that we want to support, otherwise these errors might fall through the cracks. I guess that PR https://github.com/viperproject/gobra/pull/818 might be the culprit, but we need to do more debugging to figure this out.

Example:
```go
package test

pure
requires acc(x)
decreases _
func getVal(x *int) int {
	return *x
}
```

Error:
```
[info] Writing [level:INFO] logs into (default) journal: /var/folders/4s/59ywvwc566j0cj10rn6bjcdr0000gn/T/viperserver_journal_173714098439911452264655554032242.log
[info] An unknown Exception was thrown.
[info] java.lang.RuntimeException: Unexpected expression decreases ErrorMem_ad177c4e_SY$db8f20c_ad177c4e_(thisItf) cannot be symbolically evaluated
[info] java.util.concurrent.ExecutionException: java.lang.RuntimeException: Unexpected expression decreases ErrorMem_ad177c4e_SY$db8f20c_ad177c4e_(thisItf) cannot be symbolically evaluated
[info] 	at java.base/java.util.concurrent.FutureTask.report(FutureTask.java:122)
[info] 	at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:191)
[info] 	at viper.silicon.Silicon.verify(Silicon.scala:205)
[info] 	at viper.silicon.Silicon.verify(Silicon.scala:163)
[info] 	at viper.server.core.ViperBackend.verification(VerificationWorker.scala:254)
[info] 	at viper.server.core.ViperBackend.$anonfun$execute$3(VerificationWorker.scala:164)
[info] 	at scala.util.Either.flatMap(Either.scala:352)
[info] 	at viper.server.core.ViperBackend.$anonfun$execute$1(VerificationWorker.scala:156)
[info] 	at scala.util.Either.flatMap(Either.scala:352)
[info] 	at viper.server.core.ViperBackend.execute(VerificationWorker.scala:155)
[info] 	at viper.server.core.VerificationWorker.run(VerificationWorker.scala:68)
[info] 	at viper.server.core.VerificationWorker.call(VerificationWorker.scala:109)
[info] 	at viper.server.core.VerificationWorker.call(VerificationWorker.scala:32)
[info] 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
[info] 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
[info] 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
[info] 	at java.base/java.lang.Thread.run(Thread.java:829)
[info] Caused by: java.lang.RuntimeException: Unexpected expression decreases ErrorMem_ad177c4e_SY$db8f20c_ad177c4e_(thisItf) cannot be symbolically evaluated
[info] 	at scala.sys.package$.error(package.scala:27)
[info] 	at viper.silicon.rules.evaluator$.eval2(Evaluator.scala:1287)
[info] 	at viper.silicon.rules.evaluator$.eval3(Evaluator.scala:134)
[info] 	at viper.silicon.rules.evaluator$.eval(Evaluator.scala:95)
[info] 	at viper.silicon.rules.producer$.produceTlc(Producer.scala:549)
[info] 	at viper.silicon.rules.producer$.wrappedProduceTlc(Producer.scala:200)
[info] 	at viper.silicon.rules.producer$.produceTlcs(Producer.scala:149)
[info] 	at viper.silicon.rules.producer$.$anonfun$produceTlcs$1(Producer.scala:161)
[info] 	at viper.silicon.rules.producer$.$anonfun$wrappedProduceTlc$1(Producer.scala:202)
[info] 	at viper.silicon.rules.producer$.$anonfun$produceTlc$1(Producer.scala:217)
[info] 	at viper.silicon.rules.producer$.$anonfun$produceTlc$48(Producer.scala:374)
[info] 	at viper.silicon.rules.chunkSupporter$.produce(ChunkSupporter.scala:221)
[info] 	at viper.silicon.rules.producer$.$anonfun$produceTlc$42(Producer.scala:367)
[info] 	at viper.silicon.rules.permissionSupporter$.assertNotNegative(PermissionSupporter.scala:24)
[info] 	at viper.silicon.rules.producer$.$anonfun$produceTlc$41(Producer.scala:350)
[info] 	at viper.silicon.rules.evaluator$.$anonfun$eval$1(Evaluator.scala:97)
[info] 	at viper.silicon.rules.evaluator$.$anonfun$eval3$3(Evaluator.scala:149)
```

ArquintL:
Removes the `--disablePlugins` flag for ViperServer-based backends, fixes invalid command line options that are passed to Carbon (with and without ViperServer), and adds a simple testcase for each backed.
Adding testcases required changing `Config` and the way we merge in-file configuration options. In particular, I made `backend` optional such that we know while merging configuration options whether a backend has been specified or not. We emit a violation if one attempts to merge differing backends as the merge strategy is unclear in this case.

Closes #823

### DIFF
--- a/src/main/scala/viper/gobra/backend/ViperBackends.scala
+++ b/src/main/scala/viper/gobra/backend/ViperBackends.scala
@@ -119,7 +119,7 @@ object ViperBackends {
     /** returns an existing ViperCoreServer instance or otherwise creates a new one */
     protected def getOrCreateServer(config: Config)(executionContext: GobraExecutionContext): ViperCoreServer = {
       server.getOrElse({
-        var serverConfig = List("--disablePlugins", "--logLevel", config.logLevel.levelStr)
+        var serverConfig = List("--logLevel", config.logLevel.levelStr)
         if(config.cacheFile.isDefined) {
           serverConfig = serverConfig.appendedAll(List("--cacheFile", config.cacheFile.get.toString))
         }
--- a/src/main/scala/viper/gobra/backend/ViperBackends.scala
+++ b/src/main/scala/viper/gobra/backend/ViperBackends.scala
@@ -180,7 +180,7 @@ object ViperBackends {
   case class ViperServerWithCarbon(initialServer: Option[ViperCoreServer] = None) extends ViperServerBackend(initialServer) {
     override def getViperVerifierConfig(exePaths: Vector[String], config: Config): ViperVerifierConfig = {
       var options: Vector[String] = Vector.empty
-      options ++= Vector("--logLevel", "ERROR")
+      // options ++= Vector("--logLevel", "ERROR")
       if (config.respectFunctionPrePermAmounts) {
         options ++= Vector("--respectFunctionPrePermAmounts")
       }
--- a/src/main/scala/viper/gobra/frontend/Config.scala
+++ b/src/main/scala/viper/gobra/frontend/Config.scala
@@ -248,7 +248,7 @@ case class Config(
       TypeBounds(Int = TypeBounds.IntWith64Bit, UInt = TypeBounds.UIntWith64Bit)
     }
 
-  lazy val backendOrDefault: ViperBackend = backend.getOrElse(ConfigDefaults.DefaultBackend)
+  val backendOrDefault: ViperBackend = backend.getOrElse(ConfigDefaults.DefaultBackend)
 }
 
 object Config {
@@ -871,8 +871,8 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
   conflicts(input, List(projectRoot, inclPackages, exclPackages))
   conflicts(directory, List(inclPackages, exclPackages))
 
-  // must be lazy to guarantee that this value is computed only during the CLI options validation and not before.
-  lazy val isSiliconBasedBackend = backend.toOption.getOrElse(ConfigDefaults.DefaultBackend) match {
+  // must be a function or lazy to guarantee that this value is computed only during the CLI options validation and not before.
+  private def isSiliconBasedBackend = backend.toOption.getOrElse(ConfigDefaults.DefaultBackend) match {
     case ViperBackends.SiliconBackend | _: ViperBackends.ViperServerWithSilicon => true
     case _ => false
   }
--- a/src/main/scala/viper/gobra/frontend/Config.scala
+++ b/src/main/scala/viper/gobra/frontend/Config.scala
@@ -129,7 +129,8 @@ case class Config(
                    includeDirs: Vector[Path] = ConfigDefaults.DefaultIncludeDirs.map(_.toPath).toVector,
                    projectRoot: Path = ConfigDefaults.DefaultProjectRoot.toPath,
                    reporter: GobraReporter = ConfigDefaults.DefaultReporter,
-                   backend: ViperBackend = ConfigDefaults.DefaultBackend,
+                   // `None` indicates that no backend has been specified and instructs Gobra to use the default backend
+                   backend: Option[ViperBackend] = None,
                    isolate: Option[Vector[SourcePosition]] = None,
                    choppingUpperBound: Int = ConfigDefaults.DefaultChoppingUpperBound,
                    packageTimeout: Duration = ConfigDefaults.DefaultPackageTimeout,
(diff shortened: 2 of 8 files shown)


viperserver
jcp19, line 1: Did you change the files directly in the viperserver? If not, ignore this comment (github's UI is not very clear here). Otherwise, maybe we should update the submodules instead

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -241,6 +247,8 @@ case class Config(
     } else {
       TypeBounds(Int = TypeBounds.IntWith64Bit, UInt = TypeBounds.UIntWith64Bit)
     }
+
+  lazy val backendOrDefault: ViperBackend = backend.getOrElse(ConfigDefaults.DefaultBackend)
```
jcp19, line 251: does this need to be lazy? If so, we should probably add a comment. Otherwise, I would drop the "lazy", given that this seems to always be read anyway.

In addition, if this needs to be lazy and given how "easy" the computations are and how infrequently we read this, I would probably make this a `def` instead, but this is a minor point

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -864,8 +872,8 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
   conflicts(directory, List(inclPackages, exclPackages))
 
   // must be lazy to guarantee that this value is computed only during the CLI options validation and not before.
-  lazy val isSiliconBasedBackend = backend.toOption match {
-    case Some(ViperBackends.SiliconBackend | _: ViperBackends.ViperServerWithSilicon) => true
+  lazy val isSiliconBasedBackend = backend.toOption.getOrElse(ConfigDefaults.DefaultBackend) match {
+    case ViperBackends.SiliconBackend | _: ViperBackends.ViperServerWithSilicon => true
```
jcp19, line 876: could we use `backendOrDefault` here instead of `backend.toOption.getOrElse(ConfigDefaults.DefaultBackend)`

jcp19: might be worth making this a def too

src/main/scala/viper/gobra/backend/ViperBackends.scala
```diff
@@ -119,7 +119,7 @@ object ViperBackends {
     /** returns an existing ViperCoreServer instance or otherwise creates a new one */
     protected def getOrCreateServer(config: Config)(executionContext: GobraExecutionContext): ViperCoreServer = {
       server.getOrElse({
-        var serverConfig = List("--disablePlugins", "--logLevel", config.logLevel.levelStr)
```
jcp19, line 122: interesting, I wasn't aware we were explicitly disabling plugins in ViperServer

jcp19 (approved):
LGTM, I just have a few minor comments

src/main/scala/viper/gobra/backend/ViperBackends.scala
ArquintL: Since we ran the plugins (or at least some of them) before passing the AST to the backends, there was no need to run them again

src/main/scala/viper/gobra/frontend/Config.scala
ArquintL: I don't see any necessity for making it lazy. I'm changing it to be a regular `val`

src/main/scala/viper/gobra/frontend/Config.scala
ArquintL: `backend` in this context does not refer to `Config`'s field but the Scallop CLI option. Thus, calling `backendOrDefault` is not an option. However, I'm changing this to be a (private) `def`

PR #827, opened 2025-01-24
Relative Paths in Generated Parser

ArquintL:
Adapts `genparser.sh` to turn absolute paths into relative ones in the generated parser files


jcp19 (approved):
Thank you, this is much better now!

PR #833, opened 2025-01-31
Fix issue 828

### LINKED ISSUE #828: unexpected Viper error: PredicateInstanceNoAccess
Gobra reports an unexpected Viper error for the following Gobra program:
``` go
package issue

type List struct {
	next *List
	Value int
}

pred elems(l *List) {
    l != nil ==> (acc(l) && elems(l.next))
}

requires elems(l)
decreases
func foo(l *List) {
    // invariant elems(l)
    decreases elems(l)
    for l != nil {
        unfold elems(l)
        l = l.next
    }
}

```
Error:
``` text
13:14:39.481 [ForkJoinPool-3-worker-1] ERROR viper.gobra.reporting.FileWriterReporter - Error at: <pred_term.gobra:16:15> 
Encountered an unexpected Viper error. This is a bug. The following information is for debugging purposes:
  Accessing predicate instance might fail. There might be insufficient permission to access elems_92443cc4_F(l_V0_CN0) (pred_term.gobra@16.15--16.23)

    error is PredicateInstanceNoAccess
    error offending node = PI_elems_92443cc4_F(l_V0_CN0)
    error offending node src = Some(Info(decreases  elems(l) ,elems_92443cc4_F(l_V0_CN0),Origin(pred_term.gobra@16.15--16.23,elems(l)),Vector()))
    reason is InsufficientPermission
    reason offending node = elems_92443cc4_F(l_V0_CN0)
    reason offending node src = None
        . 
```
        .

ArquintL:
This PR adds backtranslation of the (one and only) error that Viper's PredicateInstance plugin creates.

The reason is currently translated to `Permission to unknown might not suffice.` as positional information for the offending node is missing. However, this is fixed as soon [Silver PR #836](https://github.com/viperproject/silver/pull/836) got merged. I intentionally did not update the submodules and the fix introduced by this Silver PR can be propagated independently of this PR through our periodic submodule update mechansim.

Fixes #828

### DIFF
--- a/src/main/scala/viper/gobra/reporting/VerifierError.scala
+++ b/src/main/scala/viper/gobra/reporting/VerifierError.scala
@@ -355,6 +355,11 @@ case class ChannelSendError(info: Source.Verifier.Info) extends VerificationErro
   override def localMessage: String = s"The send expression ${info.trySrc[PSendStmt](" ")}might fail"
 }
 
+case class PredicateInstanceNoAccessError(info: Source.Verifier.Info) extends VerificationError {
+  override def localId: String = "predicate_instance_no_access_error"
+  override def localMessage: String = "Accessing predicate instance might fail"
+}
+
 case class FunctionTerminationError(info: Source.Verifier.Info) extends VerificationError {
   override def localId: String = "pure_function_termination_error"
   override def localMessage: String = s"Pure function might not terminate"
--- a/src/main/scala/viper/gobra/reporting/DefaultErrorBackTranslator.scala
+++ b/src/main/scala/viper/gobra/reporting/DefaultErrorBackTranslator.scala
@@ -11,6 +11,7 @@ import viper.gobra.reporting.Source.Verifier./
 import viper.silver
 import viper.silver.ast.Not
 import viper.silver.verifier.{AbstractVerificationError, errors => vprerr, reasons => vprrea}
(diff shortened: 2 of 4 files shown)


src/main/scala/viper/gobra/reporting/VerifierError.scala
```diff
@@ -355,6 +355,11 @@ case class ChannelSendError(info: Source.Verifier.Info) extends VerificationErro
   override def localMessage: String = s"The send expression ${info.trySrc[PSendStmt](" ")}might fail"
 }
 
+case class PredicateInstanceNoAccessError(info: Source.Verifier.Info) extends VerificationError {
+  override def localId: String = "predicate_instance_no_access_error"
+  override def localMessage: String = "Accessing predicate instance might fail"
```
jcp19, line 360: I would probably replace this with "Permission to predicate instance might not be available.", because it may be unclear what we mean by "accessing a predicate instance", but I leave this choice up to you.

jcp19 (approved):
LGTM, I just have a small comment about the error message.

src/main/scala/viper/gobra/reporting/VerifierError.scala
ArquintL: I was actually thinking about this as well but decided against it as this is also Viper's error message and the reason for this issue's test case specifies that the reason are insufficient permissions

PR #847, opened 2025-02-11
Expands tutorial by total correctness

### LINKED ISSUE #844: tutorial: predicate examples don't verify
I'm stepping through the Gobra [tutorial](https://github.com/viperproject/gobra/blob/77cbab0ff4330d41f48205bf991ce1777e2cd780/docs/tutorial.md) and am having trouble getting even corrected versions of the predicate examples to verify with the latest Gobra IDE extension in VSCodium:

- [x] Verifies: https://github.com/viperproject/gobra/blob/77cbab0ff4330d41f48205bf991ce1777e2cd780/docs/tutorial.md?plain=1#L295-L306

- [x] Verifies: https://github.com/viperproject/gobra/blob/77cbab0ff4330d41f48205bf991ce1777e2cd780/docs/tutorial.md?plain=1#L295-L306 + https://github.com/viperproject/gobra/blob/77cbab0ff4330d41f48205bf991ce1777e2cd780/docs/tutorial.md?plain=1#L319-L330

- [ ] Does not verify: https://github.com/viperproject/gobra/blob/77cbab0ff4330d41f48205bf991ce1777e2cd780/docs/tutorial.md?plain=1#L295-L306 + https://github.com/viperproject/gobra/blob/77cbab0ff4330d41f48205bf991ce1777e2cd780/docs/tutorial.md?plain=1#L334-L339

Verifies with:

```patch
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -333,6 +333,7 @@ In cases where statements are not allowed (e.g., specifications and bodies of pu
 
 ```go
 requires list(ptr)
+decreases
 pure func head(ptr *node) int {
   return unfolding list(ptr) in ptr.value
 }
```

- [ ] Does not verify: https://github.com/viperproject/gobra/blob/77cbab0ff4330d41f48205bf991ce1777e2cd780/docs/tutorial.md?plain=1#L295-L306 + https://github.com/viperproject/gobra/blob/77cbab0ff4330d41f48205bf991ce1777e2cd780/docs/tutorial.md?plain=1#L343-L349

Does not verify with:

```patch
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -343,6 +343,7 @@ In the previous examples of this section, the precondition requires write permis
 ```go
 requires p > 0
 requires acc(list(ptr), p)
+decreases
 pure func (ptr *node) contains(value int, ghost p perm) bool {
   return unfolding list(ptr) in ptr.value == value || (ptr.next != nil && ptr.next.contains(value, p))
 }
```

```
The pure function is not well-formed. 
Permission to list(ptr) might not suffice.
```

- [ ] Does not verify: https://github.com/viperproject/gobra/blob/77cbab0ff4330d41f48205bf991ce1777e2cd780/docs/tutorial.md?plain=1#L295-L306 + https://github.com/viperproject/gobra/blob/77cbab0ff4330d41f48205bf991ce1777e2cd780/docs/tutorial.md?plain=1#L353-L358

Does not verify with:

```patch
--- a/docs/tutorial.md
+++ b/docs/tutorial.md
@@ -352,6 +352,7 @@ Often, in particular for pure functions, we do not care about the exact permissi
 
 ```go
 requires acc(list(ptr), _)
+decreases
 pure func (ptr *node) contains(value int) bool {
   return unfolding list(ptr) in ptr.value == value || (ptr.next != nil && ptr.next.contains(value))
 }
```

```
The pure function is not well-formed. 
Permission to list(ptr) might not suffice.
```

---

**CC:** @jhoyla

ArquintL:
Adds a short section on proving total correctness to the tutorial, adapts all tutorial examples to prove termination whenever mandated by Gobra, and adds unit tests checking that the tutorial examples verify with the (mostly) default Gobra settings.

Fixes #844


src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -4526,10 +4526,16 @@ object Desugar extends LazyLogging {
 
       def goE(expr: PExpression): Writer[in.Node] = expr match {
         case p: PInvoke => info.resolve(p) match {
-          case Some(x: ap.PredicateCall) => predicateCallAccD(ctx, info)(x)(src)
+          case Some(x: ap.PredicateCall) =>
+            // we turn a `PredicateAccess` into an `Access` to unify this case
+            // with desugaring a PAccess:
+            for {
+              pa <- predicateCallAccD(ctx, info)(x)(src)
+            } yield in.Access(in.Accessible.Predicate(pa), in.FullPerm(pa.info))(pa.info)
```
jcp19, line 4534: Is there a reason why this is preferable to what we had before?

src/test/resources/regressions/examples/tutorial-examples/basic-annotations.gobra
```diff
@@ -4,34 +4,31 @@
...
+ensures res == (n % 2 == 0)
 pure func isEven(n int) (res bool) {
-	return n % 2 == 0
+  return n % 2 == 0
 }
 
 ensures isEven(n) ==> res == n / 2
 ensures !isEven(n) ==> res == n / 2 + 1
-decreases
 func halfRoundedUp(n int) (res int) {
```
jcp19, line 27: is there a reason why we are changing the indentation here? I'd rather keep tabs over spaces, as this is what gofmt enforces in .go files (same applies to all other files)

src/test/resources/regressions/examples/tutorial-examples/external-interface/counterImpl/counterImpl.gobra
```diff
@@ -6,17 +6,18 @@ package counterImpl
 // implementation
 type counter struct{ f, max int }
 
+decreases
```
jcp19, line 9: in some examples, we put the `decreases` clause before the pre/posts, in others we put it after. I would rather use a single style that we consider idiomatic throughout the examples (with a preference for putting termination measures after pre/posts)

src/test/scala/viper/gobra/GobraTutorialTests.scala
```diff
@@ -0,0 +1,27 @@
...
+
+  protected override def getConfig(source: Source): Config =
+    Config(
+      logLevel = Level.INFO,
+      reporter = StringifyReporter,
+      packageInfoInputMap = Map(Source.getPackageInfoOrCrash(source, Path.of("")) -> Vector(source)),
+      checkConsistency = true,
+      z3Exe = z3Exe,
+    )
+}
```
jcp19, line 27: is the reason to have a separate runner for the tutorial tests to guarantee that we use this simplified config?

docs/tutorial.md
```diff
@@ -187,20 +188,58 @@ func halfRoundedUp(n int) (res int) {
 ```
 
 Currently, the `pure` modifier has to be written after the pre and postcondition and before the `func` keyword. Furthermore, pure functions and methods have strict syntax constrains, the body must be a single return statement that returns a single deterministic side-effect free expression.
+Furthermore, `pure` functions must terminate. Thus, we add a `decreases` keyword instructing Gobra to prove termination as explained next.
+
+
+## Total Correctness
+
+By default, Gobra proves partial correctness, i.e., if a function terminates then its postcondition provably holds.
```
jcp19, line 196:
```suggestion
By default, Gobra proves partial correctness, i.e., if a function terminates then its postcondition holds.
```
This word doesn't add much, and might induce confusion in people from different backgrounds.

docs/tutorial.md
```diff
@@ -187,20 +188,58 @@ func halfRoundedUp(n int) (res int) {
...
+To prove total correctness, a so-called termination measure must be provided.
+A termination measure is an assertion that is bounded from below and decreases in each iteration or recursive call, allowing Gobra to prove that a function or loop provably termination.
```
jcp19, line 198: hmm, I wouldn't call it an assertion (many assertions are not valid termination measures, like `A --* B` and `acc(x.f)`) and termination measures may not be assertions (e.g., constant `0`).
I'd rather say an expression that evaluates to a type for which there is a well-founded order

jcp19 (approved):
I have a few minor remarks about the style/consistency of the text

src/main/scala/viper/gobra/frontend/Desugar.scala
ArquintL: So far, we only allowed predicate calls but no access as termination measure. Thus, it did not matter whether we store the desugared predicate call in the internal AST and defer forming the access to the encoding phase or whether we form the access already in the desugarer. However, the latter has the advantage we treat a predicate call and an access uniformly as an access in the internal AST.

src/test/resources/regressions/examples/tutorial-examples/basic-annotations.gobra
ArquintL: I changed it to spaces to allow seamless copy-pasting from the `tutorial-examples` to `tutorial.md` but will revert this change. Is it fine to indent using tabs in `tutorial.md`?

src/test/resources/regressions/examples/tutorial-examples/external-interface/counterImpl/counterImpl.gobra
ArquintL: I'm used to put it first as it's usually shorter but I'm fine with settling for putting it last

src/test/resources/regressions/examples/tutorial-examples/basic-annotations.gobra
jcp19: yup, fine by me!

PR #854, opened 2025-02-18
Fix error obtained when importing `sync` with flag --noassumeInjectivityOnInhale

jcp19:
Currently, we get an error when we import `sync` in a package that we try to verify with --noassumeInjectivityOnInhale:
```
Error at: <stubs/sync/waitgroup.gobra:24:10> Method contract is not well-formed. 
Quantified resource g might not be injective.
```

The fix is to split the array in the definition of WaitGroup to avoid generating the ill-formed qtfier at the encoding level. The actual choice of private fields does not matter and ideally, we would have been able to make this type completely opaque, but that is currently not possible in gobra


ArquintL (approved):
LGTM. However, I'm wondering whether we could address the underlying problem in the encoding as an array with 3 integers should be equivalent to 3 integers ^^

PR #855, opened 2025-02-19
Avoids type-checking if imports contain errors

ArquintL:
This PR makes type-checking non-lazy and avoids type-checking a package (and, thus, causing many consequent errors) if an imported package contains type errors.


jcp19 (commented):
I am wondering about the role of `--lazyImports` now; is this feature now deprecated?.

ArquintL:
> I am wondering about the role of --lazyImports now; is this feature now deprecated?.

I guess so. Since this PR makes the type-checker non-lazy, there is no way we could retain what this flag used to provide. I'm going to remove it

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GhostAssignability.scala
```diff
@@ -122,14 +122,32 @@ trait GhostAssignability {
...
+  private def dfltGhostAssignableMsg[L <: PNode](pre: WellDefinedness[L])(ghost: L => Boolean): (Boolean, L) => Messages = {
+    case (g, l) =>
+      val lhsMsgs = pre(l).out
+      if (lhsMsgs.isEmpty) {
+        // `l` is well-defined
+        error(l, "ghost error: ghost cannot be assigned to non-ghost", g && !ghost(l))
+      } else {
+        // We return the well-definedness messages of 'l' even though these messages should already
+        // get reported when visiting 'l' in the AST.
+        // This ensures that we do not miss them and deduplication will remove duplicate messages
```
jcp19, line 148: I'm not sure I understand why this is necessary if we are just reporting messages that have been reported before. When could we "miss" these messages if we returned `noMessages` here instead?

jcp19 (approved):
Just one question left

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GhostAssignability.scala
ArquintL: My rational is that if we ever change ghost separation to no longer iterate over all AST nodes (for whatever reason) and, thus, these messages do not get reported anymore when visiting the correspond `PParameter` node then returning here `noMessages` (aka everything looks good) despite not performing the actual checks seemed dangerous to me. If we are convinced that we would catch such a buggy code change in the future via our test suite then I'm happy to slightly simplify the code, check validity of `l` only (as opposed to collecting its messages), and return `noMessages` if it's invalid

src/main/scala/viper/gobra/frontend/info/Info.scala
```diff
@@ -248,48 +252,95 @@ object Info extends LazyLogging {
...
+        // should be reported at the import node instead of two.
+        // however, the duplicate removal should happen after translation so that the error position is correctly
+        // taken into account for the equality check.
+        val typeErrors = pkg.positions.translate(messages, TypeError).distinct
+        config.reporter report TypeCheckFailureMessage(sourceNames, pkg.packageClause.id.name, () => pkg, typeErrors)
+        Left(typeErrors)
+      },
+      info => {
+        config.reporter report TypeCheckSuccessMessage(sourceNames, config.taskName, () => info, () => pkg, () => getErasedGhostCode(pkg, info), () => getGoifiedGhostCode(pkg, info))
+        Right(info)
```
jcp19, line 324: Isn't this comment outdated now, given that we stop at the first package where we find errors?

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/separation/GhostAssignability.scala
ArquintL: I've created issue #857 for this

src/main/scala/viper/gobra/frontend/info/Info.scala
ArquintL: I've updated this comment

PR #860, opened 2025-02-24
Domains under interface

ArquintL:
This PR allows domains to implement an interface (see `Implements.scala`).
A domain is identity preserving only if all types that show up as input or output parameters of domain functions of this domain are identity preserving.

PS: the motivation for this PR is that I want to implement a ghost interface for various types, one of them being a domain.

### DIFF
--- a/src/test/resources/regressions/features/domains/implements-interface-simple02.gobra
+++ b/src/test/resources/regressions/features/domains/implements-interface-simple02.gobra
@@ -1,0 +1,53 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+// this testcase checks whether a domain can implement a non-empty interface
+
+package ImplementsInterfaceSimple01
+
+ghost type intPair domain {
+    func fst(intPair) int
+    func snd(intPair) int
+    func pair(int, int) intPair
+
+    axiom {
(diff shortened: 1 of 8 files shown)


ArquintL:
CI currently fails as it requires the changes from #862

ArquintL:
@jcp19 I made all pointers identity preserving independently of the type they are pointing to as I don't see why their identity preservation should depend on the pointee's type

src/test/resources/regressions/features/domains/implements-interface-simple02.gobra
```diff
@@ -0,0 +1,53 @@
...
+    }
+}
+
+type Equality interface {
+    ghost
+    pure hash() int
+
+    ghost
+    isEqual(other Equality) bool
+}
```
jcp19, line 28: do you think it would be worth adding a test where the interface has a non-ghost method? (my guess is no, as this should be already tested somewhere else)

jcp19 (approved):

src/test/resources/regressions/features/domains/implements-interface-simple02.gobra
ArquintL: I've added two more test cases with an `Equality` interface specifying that both functions are actual. One test case checks that you cannot implement this interface by implementing ghost functions (ghostness mismatch for the function) and the second test case checks that implementing actual functions for a receiver of domain fails (actual function cannot have a ghost receiver)

PR #869, opened 2025-02-28
Fix reserved parser keyword error reporting

ArquintL:
This PR addresses a bug discovered by @felixlinker where our parser just reports "no viable alternative at input '(proof'
func test(proof ProofType)" for the following input, thus, making it hard to spot that `proof` is a reserved word. While our parse error listener already supports pointing out reserved words, the list of reserved words was outdated. I hope to prevent similar bugs in the future by corresponding notes in `GobraLexer.g4`.

```
package main

func test(proof ProofType)
```

### DIFF
--- a/src/main/scala/viper/gobra/frontend/InformativeErrorListener.scala
+++ b/src/main/scala/viper/gobra/frontend/InformativeErrorListener.scala
@@ -21,6 +21,19 @@ import scala.collection.mutable.ListBuffer
 class InformativeErrorListener(val messages: ListBuffer[ParserError], val source: Source) extends BaseErrorListener {
 
   /**
+    * First token that Gobra introduces that is not present in Go.
+    * First refers to the order of token declarations in GobraLexer.g4
+    * as this order is preserved by ANTLR4 when generating GobraParser.java
+    */
+  private val FIRST_GOBRA_TOKEN = GobraParser.TRUE
+  /**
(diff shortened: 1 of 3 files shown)


src/main/scala/viper/gobra/frontend/InformativeErrorListener.scala
```diff
@@ -20,6 +20,19 @@ import scala.collection.mutable.ListBuffer
...
+    * as this order is preserved by ANTLR4 when generating GobraParser.java
+    */
+  private val FIRST_GOBRA_TOKEN = GobraParser.TRUE
+  /**
+    * Last token that Gobra introduces that is not present in Go.
+    * Last refers to the order of token declarations in GobraLexer.g4
+    * this order is preserved by ANTLR4 when generating GobraParser.java
+    */
+  private val LAST_GOBRA_TOKEN = GobraParser.BACKEND
+
```
jcp19, line 35: Is it guaranteed that the ids of the tokens are monotonic with respect to the position in which the tokens are declared?

jcp19 (approved):
LGTM, I just have a question about an assumption that the code makes

src/main/scala/viper/gobra/frontend/InformativeErrorListener.scala
ArquintL: As discussed offline, we leave it as is

PR #895, opened 2025-03-19
Configure Gobra via JSON

ArquintL:
This PR pushes @jcp19's idea further to use a JSON file to configure Gobra by integrating the parsing of the JSON file into Gobra. The accepted JSON structure follows [gobrago](https://github.com/jcp19/gobrago) with the following differences:
- the module-level and package-level JSON files are called `gobra-mod.json` and `gobra.json`, respectively
- the module-level JSON file has two fields, `installation_cfg` and `default_job_cfg`
- if a field is present in the package-level JSON (a so-called job configuration) then it takes precedence over the module-level field (the default job configuration). This also applies to options expressed in a `other` field. In particular, these options are parsed and merged with the other non-other fields, constituting a configuration after merging. I.e., this merging is happening before applying the precedence rule that overwrites module-level options.

In addition, the following design choices have been made:
- `--config <path>` is used to configure Gobra
- `<path>` must exist and can either be one of the following:
  - a path to a package-level JSON file in which case the Gobra files in the same folder as the JSON file are considered a package and are verified
  - a path to a directory in which case the Gobra files in this folder are considered a package and are verified. A package-level JSON file is optional
- In this configuration mode, only `--printConfig` is additionally permitted to display the actually used configuration. All other options result in an error. If `--printConfig` is not provided, Gobra prints a short note stating with configuration files have been considered.
- In this configuration mode, the module-level JSON file is mandatory, which is located by traversing the filesystem starting from `<path>` towards the filesystem's root directory


ArquintL:
@jcp19 This PR is in draft mode as the design choices are more like suggestions. Code-wise, I do not have any additional changes planned so I'd appreciate a code review. We currently do not have a JSON field for every existing CLI option but if necessary, we can always add more (in separate PRs)

jcp19:
High-level questions about the design:
- is the module-level json always required, even when we do not use the `--config` option? My expectation is that it is not, otherwise it may make it harder for new users to try out Gobra. In addition, what should be the behaviour if a user does not pass the `--config` flag but still has a module level json, or what happens if they use the `-p` flag and have a package json there?
- I wonder if the distinction between the fields of the module level and package level still make sense (why can't we have the same set of fields in both?)

jcp19:
I think that the following design could also work:
- there are two kinds of files which have the same format. Both are optional. One is called `gobra-project.json` and must be at the project root, the other is called `gobra.json` and must be in the package folder.
- by default, we always determine options based on the following hierarchy: default config (determined by the default options for all flags) < `gobra-project.json` (if present) < `gobra.json` (if present) < CLI options provided at the call site < file-level options **(*)** (the order of the last two is up for debate)
- This chain of options would always be used by default, in all modes (`-i`, `-p`, and `-r`). A short message saying which config files were found and used in the environment should be shown so that they are aware of it and are not confused by the behaviour. If no json files are found, no message is displayed (this keeps the UI light in the likely case they are not using it in a project) 
- we might consider having a flag to ignore the json files in the surrounding environment, but I don't see when this would be useful

**(*)** While thinking of how all options relate to each other, and especially, how the configs of each package relate to each other, I was confronted again with the fact that right now, we read all configs of all files, regardless of their packages. I think this is a bad idea, which leads to unexpected results

ArquintL:
> is the module-level json always required, even when we do not use the --config option? My expectation is that it is not, otherwise it may make it harder for new users to try out Gobra. In addition, what should be the behaviour if a user does not pass the --config flag but still has a module level json, or what happens if they use the -p flag and have a package json there?

Sorry for the imprecision, it's only mandatory if `--config` is used. Otherwise, it's simply ignored (same holds for `gobra.json` files)


> I wonder if the distinction between the fields of the module level and package level still make sense (why can't we have the same set of fields in both?)

One can argue whether we should allow `installation_cfg` on the module-level as well. Would you then rename `default_job_cfg` to something else such that it's uniform for module- and package-level config files (i.e., omit the "default" part as this becomes implicitly the case via our overwriting semantics for configs)?

> there are two kinds of files which have the same format. Both are optional. One is called gobra-project.json and must be at the project root or at the module root, the other is called gobra.json and must be in the package folder.

If both files are optional? How would we identify which folder forms a module's root? By detecting `go.mod`?

> This chain of options would always be used by default, in all modes (-i, -p, and -r)

Not sure whether this is less confusing but I think that this should work. The only weird case affects `-r` where the behavior is not that clear. I.e., if you invoke Gobra on a package but specify `-r`, will it iterate over all packages in the module or only over packages in subfolders contained in the current working directory?

jcp19:
> If both files are optional? How would we identify which folder forms a module's root? By detecting go.mod?

Either that or by traversing the directory tree, starting in the current package, and moving from parent to parent until a `gobra-project.json` is found.

> Not sure whether this is less confusing but I think that this should work. The only weird case affects -r where the behavior is not that clear. I.e., if you invoke Gobra on a package but specify -r, will it iterate over all packages in the module or only over packages in subfolders contained in the current working directory?

IIRC, you do not pass a package in the recursive mode, but rather, the project root. So, I guess we would iterate through all packages in that directory. I think we need to think about what should be the roles of the project root and module root (should they be unified somehow?).

jcp19:
> @ArquintL this is marked as a draft PR but you asked for the review. Should I provide feedback on the implementation already? regarding the cli options, I would opt for one of the following:
> 
> 1. make it an error to provide additional flags when `--config <path>` is passed
> 2. whatever flags passed on the CI would override those in the config (this is useful if you want to mostly reuse the config, but try out different options quickly), and I would produce a message saying that this is the case

After additional consideration, I am happy with @ArquintL general proposal, but I would rather choose one of the behaviours I proposed above for the case when we provide more CLI flags in addition to `--config`. I will proceed with the code review. In addition, we should probably show to the user which packages are being verified when the --config option is used, to avoid confusing the user (for example, because they passed the wrong file to a json file)

jcp19 (commented):
So far, I reviewed up to `src/main/scala/viper/gobra/frontend/Config.scala`. I will continue the review later

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -447,6 +502,182 @@ case class RecursiveModeConfig(projectRoot: Path = ConfigDefaults.DefaultProject
...
+            inputDirectories = Vector(pathToDirectory),
+            baseConfig = baseConfig,
+          ).config
+      }
+      otherArgs = mergedConfig.job_cfg.other.getOrElse(List.empty)
+      otherConfig <- if (otherArgs.isEmpty) Right(None)
+        else {
+          // at this point, we do not know anymore from which config file these arguments come from.
+          // Thus, we do not resolve potential paths.
+          Config.parseCliArgs(otherArgs, None).map(Some(_))
```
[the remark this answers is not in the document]
ArquintL: I've changed the implementation accordingly

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -80,29 +81,54 @@ object ConfigDefaults {
...
+    def value: String
+  }
+
+  /** trait for an enum value to configure parsing and serialization of this enum */
+  trait CliEnum[E <: EnumCase] {
+    def values: List[E]
+    def convert(s: String): E = values.find(_.value == s).getOrElse(Violation.violation(s"Unexpected value: $s"))
+  }
+}
+
```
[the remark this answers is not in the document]
ArquintL: Good point. However, `ViperBackend` and `MoreJoins` carry additional fields which are useful but do not seem to be supported by `Enumeration` (at least according to Claude)

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -297,11 +364,26 @@ object Config {
   def sourceHasHeader(s: Source): Boolean = header.findFirstIn(s.content).nonEmpty
 
   val enableExperimentalHyperFeaturesOptionName = "enableExperimentalHyperFeatures"
-  val enableLazyImportOptionName = "enableLazyImport"
-  val enableLazyImportOptionPrettyPrinted = s"--$enableLazyImportOptionName"
+
+  /** Options that must not appear in the `other` field of a JSON config because their paths
+    * cannot be correctly resolved (they are turned into Source objects before path resolution). */
+  private val disallowedInOther: Set[String] = Set("--input", "-i", "--directory", "-p")
```
jcp19: what about include paths (`-I`) and `--projectRoot`? Is there a criteria for things commands whose paths are resolved? And for those commands where we resolve the paths, are they resolved against the current path?

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -357,6 +439,591 @@ case class BaseConfig(gobraDirectory: Option[Path] = ConfigDefaults.DefaultGobra
...
+object InputConfigOption {
+  /** Implicit class to extend ScallopOption with toInputConfigOption conversion */
+  implicit class ScallopOptionExtension[T](opt: ScallopOption[T]) {
+    /** Converts a ScallopOption to an InputConfigOption using the option's name and value.
+      * Only includes the value if the option was explicitly supplied by the user,
+      * not just set to a default value. */
+    def toInputConfigOption: InputConfigOption[T] =
+      InputConfigOption(opt.name, if (opt.isSupplied) opt.toOption else None)
+  }
+}
```
jcp19: I feel like this use of `implicit` affords us very little, and I would rather have the more predictable behaviour and maintainability of having `toInputConfigOption` as a regular (static) method, rather than an extension method.

src/main/scala/viper/gobra/frontend/JsonConfig.scala
```diff
@@ -0,0 +1,156 @@
...
+                            jar_path: Option[String] = None,
+                            jvm_options: Option[List[String]] = None,
+                            z3_path: Option[String] = None,
+                          ) extends Resolvable {
+  type R = GobraInstallCfg
+
+  /** resolves all relative paths in relation to `basePath` */
+  override def resolvePaths(basePath: Path): GobraInstallCfg =
+    copy(
+      jar_path = resolveOptPath(basePath, jar_path),
```
jcp19: hmm, what's the point of passing the jar here? given that a user has to call gobra (and select the jar there), this feels redundant

src/main/scala/viper/gobra/frontend/JsonConfig.scala
```diff
@@ -0,0 +1,156 @@
...
+                               check_consistency: Option[Boolean] = None,
+                               overflow: Option[Boolean] = None,
+                               conditionalize_permissions: Option[Boolean] = None,
+                               only_files_with_header: Option[Boolean] = None,
+                               includes: Option[List[String]] = None,
+                               input_files: Option[List[String]] = None,
+                               mce_mode: Option[MCE.Mode] = None,
+                               module: Option[String] = None,
+                               more_joins: Option[MoreJoins.Mode] = None,
+                               pkg_path: Option[String] = None, // TODO: what is this?
```
jcp19: I guess this is the input to `-p`

src/main/scala/viper/gobra/frontend/JsonConfig.scala
```diff
@@ -0,0 +1,156 @@
...
+                               parallelize_branches: Option[Boolean] = None,
+                               print_vpr: Option[Boolean] = None,
+                               project_root: Option[String] = None,
+                               recursive: Option[Boolean] = None, // TODO: why would a package specify recursion?
```
jcp19: I don't oppose dropping this for json configs.

jcp19 (approved):
In general, it looks ok to me. I have a few minor comments / remarks

src/main/scala/viper/gobra/frontend/Config.scala
ArquintL: Good catch! I think I wanted to disallow all config options that allow you to provide a path to basically force users to use structured JSON fields instead of resorting to `other`

src/main/scala/viper/gobra/frontend/JsonConfig.scala
ArquintL: Hmm I think I took the initial definition of these structures from your separate Go tool building the command. I'm happy to drop this option but I'm wondering whether there's a use for it in Gobra-IDE. But I guess for Gobra-IDE / within VSCode, we already use `settings.json` to configure Gobra-IDE so this option would be more natural there

jcp19: yeah, in gobrago, we had to select the jar, but that implicitly happens now when we call gobra with --config

src/main/scala/viper/gobra/frontend/Config.scala
ArquintL: The resolution logic takes the `includes`, `input_files`, `pkg_path`, and `project_root` fields of `VerificationJobCfg` and resolves the paths specified therein based on the directory in which the package-specific config file sits

src/main/scala/viper/gobra/frontend/JsonConfig.scala
ArquintL: I leave this option for the moment but I've added a long comment mentioning that one might want to drop or move this option. In particular, it could make sense to allow users to specify `recursive` on the module-level such that verifying the module results in recursively verifying all its packages. However, we would then have to revisit the meaning of `recursive` as currently `recursive`, `inputs`, and `directory` are mutually exclusive. It would however make sense that each package could specify the files that constitute the package

PR #896, opened 2025-03-20
Update Submodules

viper-admin:
* Updates ViperServer from `182a2436152dd8eff07c512671b69707f009b0ad` to `711507c13d169342382ccdf356dc37ea8bfc5ed7`.


ArquintL:
Viper added `PluginAwareReporter` (see https://github.com/viperproject/silver/pull/854). While ViperServer also has been updated to use this new reporter trait, we should check before merging this PR whether we can also simplify Gobra by implementing this trait or whether this is unnecessary as ViperServer already does so. Hint: look for calls to `mapEntityVerificationResult` and `mapVerificationResult`

jcp19 (approved):

PR #897, opened 2025-03-20
Fix type-checker crash when encountering an invalid ADT

ArquintL:


src/test/resources/regressions/features/adts/match-error1.gobra
```diff
@@ -0,0 +1,29 @@
...
+    RootEntry{}
+    SendEntry{
+        //:: ExpectedOutput(type_error)
+        msg InexistentType // unknown identifier
+        prev Trace
+    }
+}
+
+ghost
+decreases _ // assume termination instead of using `len(t)`, which would cause another type error
```
jcp19, line 20: the type error we would get here would be essentially the same we have below, right?

src/test/resources/regressions/features/adts/match-error1.gobra
```diff
@@ -0,0 +1,29 @@
...
+ensures   0 <= res
+pure func (t Trace) length() (res int) {
+    return match t {
+        case RootEntry{}: 0
+        // we need a recursive call to trigger the error in a previous version of Gobra
+        //:: ExpectedOutput(type_error)
+        case SendEntry{_, ?p}: 1 + p.length() // `length` is an unknown identifier as `Trace` is malformed
```
jcp19, line 27: Maybe in the future (on separate PRs) we should look into improve the error reporting, and in particular, we could look into not reporting type errors on methods whose parameters are of malformed types. This error here is completely redundant (given that it occurs when we have a type error in the definition of the type), distracting and potentially confusing (iirc that the error that is reported here if that p.length() is not known).

PR #900, opened 2025-03-31
Fix bug 805

### LINKED ISSUE #805: Array Pointer Shorthand
> For a of pointer to array type: `a[x]` is shorthand for `(*a)[x]`   ([go spec](https://go.dev/ref/spec#Index_expressions))

This is currently not handled by Gobra and one must explicitly dereference the pointers.
The specs can get hard to read: 
```go
// @ preserves acc(&((*p)[0]))
// @ ensures (*p)[0] == 0
func dereferencing(p *[10]int) {
	(*p)[0] = 0
}
```
We should be able to write:
```go
// @ preserves acc(&p[0])
// @ ensures p[0] == 0
func shorthand(p *[10]int) {
	p[0] = 0
}
```

```text
15:18:02.095 [main] ERROR viper.gobra.GobraRunner$ - An assumption was violated during execution.
15:18:02.096 [main] ERROR viper.gobra.GobraRunner$ - Logic error: Expected slice, array, map, or sequence, but got *[10]int
viper.gobra.util.Violation$LogicException: Logic error: Expected slice, array, map, or sequence, but got *[10]int
```

For pointers to structs the indirection works:
```go
func structs() {
	p := new(struct{ value int })
	p.value = 1
	//@ assert p.value == 1
}
```

jcp19:
This PR fixes #805 by adding support for operations over pointers to arrays. In addition, it simplifies some code in neighboring parts of code that was unnecessarily complicated

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Desugar.scala
+++ b/src/main/scala/viper/gobra/frontend/Desugar.scala
@@ -2762,20 +2762,9 @@ object Desugar extends LazyLogging {
             case t => Violation.violation(s"desugaring of slice expressions of base type $t is currently not supported")
           }
 
-          case PLength(op) => for {
-            dop <- go(op)
-          } yield dop match {
-            case dop : in.ArrayLit => in.IntLit(dop.length)(src)
-            case dop : in.SequenceLit => in.IntLit(dop.length)(src)
-            case _ => in.Length(dop)(src)
-          }
+          case PLength(op) => go(op).map(in.Length(_)(src))
 
-          case PCapacity(op) => for {
-            dop <- go(op)
-          } yield dop match {
-            case dop : in.ArrayLit => in.IntLit(dop.length)(src)
-            case _ => in.Capacity(dop)(src)
-          }
+          case PCapacity(op) => go(op).map(in.Capacity(_)(src))
 
           case g: PGhostExpression => ghostExprD(ctx, info)(g)
 
--- a/src/main/scala/viper/gobra/frontend/Desugar.scala
+++ b/src/main/scala/viper/gobra/frontend/Desugar.scala
@@ -2744,12 +2744,13 @@ object Desugar extends LazyLogging {
                 val drop = in.SequenceDrop(dbase, lo)(src)
                 in.SequenceTake(drop, sub)(src)
             }
-            case baseT @ (_: in.ArrayT | _: in.SliceT) => (dlow, dhigh) match {
-              case (None, None) => in.Slice(dbase, in.IntLit(0)(src), in.Length(dbase)(src), dcap, baseT)(src)
-              case (Some(lo), None) => in.Slice(dbase, lo, in.Length(dbase)(src), dcap, baseT)(src)
-              case (None, Some(hi)) => in.Slice(dbase, in.IntLit(0)(src), hi, dcap, baseT)(src)
-              case (Some(lo), Some(hi)) => in.Slice(dbase, lo, hi, dcap, baseT)(src)
-            }
+            case baseT @ (_: in.ArrayT | _: in.SliceT | in.PointerT(_: in.ArrayT, _)) =>
+              (dlow, dhigh) match {
+                case (None, None) => in.Slice(dbase, in.IntLit(0)(src), in.Length(dbase)(src), dcap, baseT)(src)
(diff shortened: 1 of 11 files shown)


src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -521,6 +525,10 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
     case PLength(op) => isExpr(op).out ++ {
       underlyingType(exprType(op)) match {
         case _: ArrayT | _: SliceT | _: GhostSliceT | StringT | _: VariadicT | _: MapT | _: MathMapT => noMessages
+        case ActualPointerT(_: ArrayT)  =>
+          // Go allows getting the length of a pointer to an array, but it does not allow obtaining the
+          // length of a pointer of a type T whose underlying slice is an array.
```
ArquintL, line 530:
```suggestion
          // Go allows getting the length of a pointer to an array, but it does not allow obtaining the
          // length of a pointer of type T, whose underlying type is an array.
```

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -529,6 +537,10 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
     case PCapacity(op) => isExpr(op).out ++ {
       underlyingType(exprType(op)) match {
         case _: ArrayT | _: SliceT | _: GhostSliceT => noMessages
+        case ActualPointerT(_: ArrayT)  =>
+          // Go allows getting the capacity of a pointer to an array, but it does not allow obtaining the
+          // capacity of a pointer of a type T whose underlying slice is an array.
```
ArquintL, line 542:
```suggestion
          // Go allows getting the capacity of a pointer to an array, but it does not allow obtaining the
          // capacity of a pointer of a type T, whose underlying type is an array.
```

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -529,6 +537,10 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
+        noMessages
```
ArquintL, line 543:
```suggestion
          noMessages
```

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -2773,11 +2774,7 @@ object Desugar extends LazyLogging {
             dop <- go(op)
           } yield dop match {
             case dop : in.ArrayLit => in.IntLit(dop.length)(src)
```
ArquintL, line 2776: Is there a reason for specialcasing `ArrayLit` here? Given that you've extended the array encoding to support `cap`, it seems to me we could simply convert `PCapacity` to `Capacity`.
Furthermore, the same comment also applies to the lines above for the `PLength` case

ArquintL (approved):
I've some minor suggestions and a question

src/main/scala/viper/gobra/frontend/Desugar.scala
jcp19: good point, I don't see immediately why you would do that, except for wanting to avoid a call to the length function, which may be useful in some cases to avoid introducing triggering terms (for the same reason, it may lead to confusing triggering behaviour, as a pattern that is expected to occur does not occur in the generated viper program). I have simplified this

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
jcp19: I agree with this suggestion, except for the comma before "whose", given that the clause introduced by whose is supposed to be restrictive. Putting the comma thus changes the meaning of the sentence

src/main/scala/viper/gobra/frontend/Desugar.scala
jcp19: ah, this special case was prior to my changes. I removed the special case

ArquintL: Yes, I could have been clearer on this ^^

PR #904, opened 2025-03-31
Fix issue 878

jcp19:


ArquintL (approved):
Can you please adapt the documentation on lines 120-138?
Furthermore, this PR introduces `low` assumptions and invariants unconditionally on whether we enabled a product construction. I would find it cleaner to do so only if `hyperMode` is set to `on` or `extended` such that we can check for the existence of `low` and `lowEvent` assertions instead of blindly removing them when `hyperMode` is disabled

ArquintL:
This btw is exactly the reason for the failing CI

ArquintL:
> Hmm, doing so may require extending our encoding API to pass a Config object to this point in the code. Do you see a better way of doing this?

While looking into it, I realized that it's easier than it sounds and quickly did it :P

PR #906, opened 2025-03-31
Fix issue 881

### LINKED ISSUE #881: hyperGobra doesn't encode packaging magic wands correctly
I am working on verifying SIF for the VerifiedSCION router (using hyperGobra) with @jcp19.

The following code verifies successfully using "normal" Gobra (or hyperGobra with `--hyperMode=off`).

```go
ghost
requires Pred(x)
ensures Pred(1)
decreases
func Step(x int)

pred Pred(x int)

func test() {
	package Pred(0) --* Pred(1) {
		Step(0)
	}
}
```

However, when using hyperGobra with `--hyperMode=on`, we get an error (`java.util.NoSuchElementException: key not found: x_V01`). The problem seems to be that the operations in the `package` block are not correctly translated during the product-program construction, as the Viper encoding calls the encoding of `Step` with just one argument (using `--checkConsistency` suggests the same):

```
method test_3d39276c_F() {
    ...
    // package acc(Pred_3d39276c_F(0)) --* acc(Pred_3d39276c_F(1))
    package acc(Pred_3d39276c_F0(0), write) --*
    acc(Pred_3d39276c_F0(1), write) {
      // Step_3d39276c_F(0)
      Step_3d39276c_F(0)
    }
    package acc(Pred_3d39276c_F1(0), write) --*
    acc(Pred_3d39276c_F1(1), write) {
      // Step_3d39276c_F(0)
      Step_3d39276c_F(0)
    }
}
```

jcp19:
Fixes #881

### DIFF
--- a/src/main/scala/viper/gobra/translator/transformers/hyper/SIFLowGuardTransformer.scala
+++ b/src/main/scala/viper/gobra/translator/transformers/hyper/SIFLowGuardTransformer.scala
@@ -475,6 +475,11 @@ trait SIFLowGuardTransformer {
       case v: LocalVarDecl => v.copy(name = ctx.rename(v.name))(v.pos, v.info, v.errT)
       case f: Field => runField(f, ctx)
       case e: Exp => runExp(e, ctx)
+      case p: Package =>
+        Package(
+          wand = runExp(p.wand, ctx).asInstanceOf[MagicWand],
+          proofScript = toSeqn(statement(p.proofScript, ctx)),
+        )(p.pos, p.info, p.errT)
     }
   }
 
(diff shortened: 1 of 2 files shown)


jcp19:
Yeah, you raise a good point. For the time being, it seems that we are stuck maintaining two similar-ish plugins. I think in the long term, we should unify most of the implementations of these plugins. In particular, we should:
- unify the code that establishes the calling conventions (including order of the arguments), so that we can in principle use both encodings in the same program
- identify the shared functionality (e.g., the transformations over the expressions) and have it defined once and forall

Unfortunately, due to time constraints of our students, I don't think there is a way around merging a few patches that fix pressing issues.

My comments also partially address the growing changeset of `hyperMode`. In addition, I think it is unavoidable that this will continue growing, as the original PR supports only a very small and well-behaved fragment of Gobra which is not enough for larger projects like scion

ArquintL:
Given that hyper Gobra is widely used (relative to the adoption of Gobra itself), I think we should rather push for a quick-ish review of `hyperGobra` and merge it into `master` as otherwise we end up with a monster PR and re-review some changes

PR #909, opened 2025-04-02
errors.New

ArquintL:
This PR extends the `errors` package to (finally!) provide the `New` function.

### DIFF
--- a/src/main/resources/stubs/errors/errors.gobra
+++ b/src/main/resources/stubs/errors/errors.gobra
@@ -11,8 +11,8 @@
 //
 // An error e wraps another error if e's type has one of the methods
 //
-//	Unwrap() error
-//	Unwrap() []error
+//  Unwrap() error
+//  Unwrap() []error
 //
 // If e.Unwrap() returns a non-nil error w or a slice containing w,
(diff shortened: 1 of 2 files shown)


ArquintL:
@jcp19 Was there a particular reason to add `wrap.go` but not `errors.go`? It seems strange to me that we provide (some advanced) parts of the `errors` package but not the most basic one

jcp19:
> @jcp19 Was there a particular reason to add `wrap.go` but not `errors.go`? It seems strange to me that we provide (some advanced) parts of the `errors` package but not the most basic one

the reason is that I never needed errors.go 😄

src/main/resources/stubs/errors/errors.gobra
```diff
@@ -0,0 +1,122 @@
...
+
+// New returns an error that formats as the given text.
+// Each call to New returns a distinct error value even if the text is identical.
+ensures err != nil && err.ErrorMem()
+decreases
+func New(text string) (err error) {
+	err = &errorString{text}
+    fold err.ErrorMem()
+    return
+}
```
jcp19, line 73: the indentation is inconsistent

jcp19 (approved):
LGTM

PR #923, opened 2025-05-09
Add support for converting maps to dicts

jcp19:
In a recent conversation with @ThomasMayerl and @rayman2000, we realized that introducing ghost pure functions that convert from maps to dictionaries and from slices to sequences is still an unfortunate rite of passage for any user of Gobra. This PR addresses the former, which is impossible to fully implement as a pure function using the current syntax of Gobra, as we do not have a good way of iterating through maps without using loops. In addition, this conversion can be implemented super concisely and efficiently in Gobra already, due to the way maps are encoded.

I considered also implementing the conversion from slices to ghost collections, including seqs, sets and msets, but I don't have time atm

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
@@ -222,6 +222,10 @@ trait GhostExprTyping extends BaseTyping { this: TypeInfoImpl =>
           case Single(_: MathMapT | _: MapT) => isExpr(exp).out
           case t => error(expr, s"expected a map, but got $t")
         }
+        case PMathMapConversion(op) => underlyingType(exprType(op)) match {
+          case Single(_: MapT) => isExpr(op).out
+          case t => error(expr, s"expected a map, but got a $t")
+        }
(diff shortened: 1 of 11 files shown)


ArquintL (approved):
Very cool, thanks so much! I was actually quite shocked by the size of this PR as I'd have expected another built-in pure function with some pre- and postconditions but after looking at the changeset, this makes much more sense to actually make use of the underlying math. map

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
```diff
@@ -333,6 +337,10 @@ trait GhostExprTyping extends BaseTyping { this: TypeInfoImpl =>
           case Single(t: MapT) => SetT(t.elem)
           case t => violation(s"expected a map, but got $t")
         }
+        case PMathMapConversion(op) => underlyingType(exprType(op)) match {
+          case MapT(k, v) => MathMapT(k, v)
+          case t => violation(s"expected a sequence, set, multiset or option type, but got $t")
```
jcp19, line 342:
```suggestion
          case t => violation(s"expected a map but got $t")
```

PR #927, opened 2025-06-02
Generate cleaner output

jcp19:
Currently, the Gobra output contains noise that is totally unnecessary:
1. the option `--gobraDirectory` is almost never used. More often than that, people do not inspect the results of the file. To make matters worse, the default tends to lead to errors related to permissions to create/modify pre-existing directories. Thus, one usually finds the following two unnecessary lines in every execution of Gobra:
```
[info] Writing report to .gobra/stats.json
[info] Could not write to the file .gobra/stats.json. Check whether the permissions to the file allow writing to it.
```

Solution: unless explicitly instructed, Gobra does not generate such a file.

2. Gobra reports a warnings, but not the contents of the warnings themselves. This is rather confusing to users. In addition, these warnings are imprecise, and often report dependencies on unverified packages even though the packages **are** verified.

Solution: for now, I changed the debug level of the message. Nonetheless, I'm keen on completely dropping these kinds of messages from the system. What do you think, @ArquintL?

3. The message "Closures in Gobra are experimental and ..." is just annoying. Gobra is an experimental tool, and emphasizing that only closures are experimental might give the wrong impression.

Solution: I finally dropped it 🎉

### DIFF
--- a/src/main/scala/viper/gobra/Gobra.scala
+++ b/src/main/scala/viper/gobra/Gobra.scala
@@ -133,7 +133,7 @@ trait GoVerifier extends StrictLogging {
 
     // Print warnings
     if(warningCount > 0) {
-      logger.info(s"$name has found $warningCount warning(s)")
+      logger.debug(s"$name has found $warningCount warning(s)")
     }
 
     // Print errors
--- a/src/main/scala/viper/gobra/Gobra.scala
+++ b/src/main/scala/viper/gobra/Gobra.scala
@@ -70,13 +70,16 @@ trait GoVerifier extends StrictLogging {
     // by some signal.
     Runtime.getRuntime.addShutdownHook(new Thread() {
       override def run(): Unit = {
-        val statsFile = config.gobraDirectory.resolve("stats.json").toFile
-        logger.info("Writing report to " + statsFile.getPath)
-        val wroteFile = statsCollector.writeJsonReportToFile(statsFile)
-        if (!wroteFile) {
-          logger.error(s"Could not write to the file $statsFile. Check whether the permissions to the file allow writing to it.")
+        config.gobraDirectory match {
+          case Some(path) =>
+            val statsFile = path.resolve("stats.json").toFile
+            logger.info("Writing report to " + statsFile.getPath)
+            val wroteFile = statsCollector.writeJsonReportToFile(statsFile)
+            if (!wroteFile) {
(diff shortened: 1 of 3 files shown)


ArquintL:
> For reference, the kinds of warnings generated by Gobra look like the following (exemplified on a package from Gobra):

>> [info] Depending on imported package that is not verified: ../VerifiedSCION/verification/utils/slices - slices
>> [info] Depending on imported package that is not verified: ../VerifiedSCION/pkg/private/serrors - serrors
>> [info] Depending on imported package that is not verified: ../VerifiedSCION/verification/utils/monoset - monoset
>> [info] Depending on imported package that is not verified: ../VerifiedSCION/pkg/slayers/path - path

> Well, all of these packages are verified 😅

Is this because you verify them individually, i.e., using multiple invocations of Gobra? The idea is that you verify all packages you want to verify in a module and Gobra will then tell you what the remaining assumptions are

ArquintL:
> do you then pass the entire module and pass --includePackages to select the package you want to verify? What if two packages are named the same?

Yes, I do, such that changing the CLI command remains easy ^^
Uniqueness of package names is a requirement though, which, fortunately, is under my control. In the long run, it might be better to switch from package names to import paths to avoid this problem

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -119,24 +126,32 @@ trait GoVerifier extends StrictLogging {
...
-    logger.debug(s"${statsCollector.getNumberOfSpecifiedMembersWithAssumptions} are assumed to be satisfied")
+    logger.debug(s"Gobra found ${statsCollector.getNumberOfVerifiableMembers} methods and functions." )
+    logger.debug{
+      val specMem = statsCollector.getNumberOfSpecifiedMembers
+      s"$specMem member${addPlural(specMem)} are explicitly specified."
+    }
+    logger.debug{
+      val memAssum = statsCollector.getNumberOfSpecifiedMembersWithAssumptions
+      s"$memAssum member${addPlural(memAssum)} contain explicit assumptions."
+    }
```
ArquintL, line 142: Out of curiosity, what are explicit members and explicit assumptions? Does this refer to members and assumptions in the Gobra program as opposed to in the encoding?

ArquintL (approved):
LGTM, one question out of curiosity

PR #933, opened 2025-06-06
Fix using for range over map without using visited

### LINKED ISSUE #808: Map Range mandatory with
For range loops over maps, one has to add `with visited` even if `visited` is not used. 
Otherwise, we get an error "Loop invariant might not be established".
```go
// @ requires acc(m)
func mapRange(m map[int]int) {
	//@ invariant acc(m)
	for k, v := range m {}
}
```
This is inconsistent with slices where it is not mandatory.
```go
// @ requires acc(s)
func sliceRange(s []int) {
	//@ invariant acc(s)
	for i, e := range s {}
}
```

ThomasMayerl:
This PR fixes #808 by only generating invariants for the visited set when visited is used


jcp19 (approved):
LGTM

PR #940, opened 2025-06-29
Improve let-in syntax

### LINKED ISSUE #899: Gobra reports let-in parser error if there is a parser error in another function
The following code verifies because the syntax error is commented out:

```go
var mapstrstr = map[string][]string{}

preserves acc(mapstrstr)
preserves let val, ok := mapstrstr[str] in acc(val)
func test(str string) 

//preserves  */thismakesnosense
func test_2()
```

However, in the following snippet, Gobra will report an error for the let-in clause even though the syntax error can be found in another function:
```go
var mapstrstr = map[string][]string{}

preserves acc(mapstrstr)
preserves let val, ok := mapstrstr[str] in acc(val)
func test(str string) 

preserves  */thismakesnosense
func test_2()
```
The error reported is `unexpected end of line in expression, expecting 'in'` for the line with the let-in clause.

ThomasMayerl:
This PR fixes #899. Essentially, the issue was that if we use a let-in clause, if we - at a later point - have a syntax error, Gobra would mark the let-in cause as being wrong (even if the let-in clause is correct and the error occurs later). Basically, the issue is that ANTLR generates an LL Parser and that it first tries to put as much as possible into shortVarDecl. Because IN can also be part of expression, meaning it can also be part of shortVarDecl, if there is a syntax error, the parser will put this (including the IN) into the shortVarDecl. The syntax error reported is then that the IN clause is missing. Just due to the way an LL Parser works this issue cannot really be fixed without adapting the syntax (e.g. by putting parentheses around the shortVarDecl part). However, this would break existing programs. Thus, this PR basically allows both syntaxes: For existing programs, the version without parentheses still works. For new programs or existing programs with new syntax errors, we can "opt-in" by just placing parentheses around the shortVarDecl and thus resolving the ambiguity.

### DIFF
--- a/src/main/antlr4/GobraParser.g4
+++ b/src/main/antlr4/GobraParser.g4
@@ -334,10 +334,14 @@ expression:
   |<assoc=right> expression IMPLIES expression #implication
   |<assoc=right> expression QMARK expression COLON expression #ternaryExpr
   | UNFOLDING predicateAccess IN expression #unfolding
-  | LET shortVarDecl IN expression #let
+  | letExpression #let
   | (FORALL | EXISTS) boundVariables COLON COLON triggers expression #quantification
   ;
 
+letExpression
+  : LET L_PAREN shortVarDecl R_PAREN IN expression
+  | LET shortVarDecl IN expression
+  ;
 
 // Added ghost statements
 statement:
--- a/src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
+++ b/src/main/scala/viper/gobra/frontend/ParseTreeTranslator.scala
@@ -1715,8 +1715,8 @@ class ParseTreeTranslator(pom: PositionManager, source: Source, specOnly : Boole
   }
 
   override def visitLet(ctx: LetContext): PLet = {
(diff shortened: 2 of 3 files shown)


ThomasMayerl:
It seems that requiring parentheses around in-clauses regarding sets does indeed solve the problem. However, shortVarDecl uses expression which accepts
```
expression p41_op = (
    IN
    | MULTI
    | SUBSET
  ) expression #p41Expr
```
Thus, if we changed it there to require parentheses, all expressions using in (even outside of let) would require them, too. Alternatively, we could change shortVarDecl (or let) to use its own kind of expression where we require parentheses around in, which, in my opinion, is not a nice solution.

ArquintL:
I quickly consulted ChatGPT and it seems that there are two options for achieving the context-based enforcement of parentheses:
- use a global flag (in `@parser::members`) indicating that we are in the context of parsing a let expression. The downside is that we would disallow `in` in any subexpression unless we carefully reset the flag after processing the top-level expression
- we have to special-case the `expr` rule and add a rule like `exprWithParenthesizedIn`, which is used only by the let expression. We could avoid some of the duplication by introducing rules for the separate cases, like unary expressions.

PR #944, opened 2025-07-17
Use `elem` to denote ghost collection containment

### LINKED ISSUE #899: Gobra reports let-in parser error if there is a parser error in another function

ArquintL:
This PR resolves parsing ambiguities (that manifest, e.g., as a parse error at a let expression despite the actual error being much later in the program text) by renaming `in` to `elem`.
@reviewers: I've separated the changes in Gobra and the tests by placing them in different commits, which might help the reviewing process.

Verifying this program now results in the following error message:
```
func foo(ghost xs seq[int]) {
  assert 42 in xs
}
```
```
Error at: <file, line, column of 'in'> Unexpected 'in' encountered. Did you mean to use 'elem' denoting ghost collection containment?
  assert true in xs
              ^^
```

Supersedes #940
Closes #899

cc @ThomasMayerl


jcp19 (approved):
LGTM

PR #947, opened 2025-07-18
Fix issue 943

jcp19:
Rather surprisingly, checking whether a quantified assertion is strongly pure was never properly implemented.

### DIFF
--- a/src/test/resources/regressions/issues/000943.gobra
+++ b/src/test/resources/regressions/issues/000943.gobra
@@ -1,0 +1,14 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package issue000943
+
+//:: ExpectedOutput(type_error)
+preserves forall i int :: 0 <= i && i < len(b) ==> acc(&b[i])
+decreases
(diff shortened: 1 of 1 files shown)


ArquintL (approved):
LGTM. I'm however a bit surprised that this PR does not change the behavior of preserves for pure functions. Is the issue in this case limited to just quantified permissions?

jcp19:
> I'm however a bit surprised that this PR does not change the behavior of preserves for pure functions

I'm not sure what you mean. We can still have pure `preserves` in preconditions, and we checked before that they were pure (it just so happened the check was wrong for QPs). Sure, we could forbid `preserves` in pure functions altogether, but I don't find this strictly necessary, in the same way we do not prevent `requires A; ensures A` in the specs of pure functions.

ArquintL:
I was wondering about how we treat `preserves` in the context of pure functions as I don't see any use case. It's rejected if it the assertion is not strictly pure and if it is then no additional knowledge is gained (i.e., `preserves A` could be replaced by `requires A` without hurting any caller).
Based on the description of #942, it seemed that strange things happen with `preserves` as the generated Viper function did neither have pre- nor postconditions.

Issue #831, opened 2025-01-30
Low sensitivity doesn't transfer when converting string to byte slice


henriman:
I am working on verifying SIF for the VerifiedSCION router using hyperGobra with @jcp19.

jcp19:
Fixed in #961

Issue #832, opened 2025-01-30
Low sensitivity doesn't transfer when converting `[]byte` to `string`


henriman:
I am working on verifying SIF for the VerifiedSCION router (using hyperGobra) with @jcp19.

When converting a low byte slice `data []byte` to a string (`string(data)`), the resulting string is not considered low.

Consider the following two examples. For both `test10` and `test11`, the postcondition cannot be established.

```go
ensures low(res)
func test10() (res string) {
    res := string([]byte{0, 1, 2})
}

requires acc(data)
requires low(data)
requires forall i int :: { data[i] } 0 <= i && i < len(data) ==> low(data[i])
ensures low(res)
func test11(data []byte) (res string) {
    res := string(data)
}
```

It seems that in the Viper encoding, the two versions of the variables introduced by the product program construction use two different (abstract) functions to "perform the conversion" (`byteSliceToStrFunc0` and `byteSliceToStrFunc1` resp.), and thus we can't infer that if the variables were equal before, their conversions will be equal as well.

jcp19:
From my perspective, this issue is fixed, but a related issue is outstanding (https://github.com/viperproject/gobra/issues/966).

PR #961, opened 2025-08-17
Add support for indexing strings

jcp19:
This PR:
- Adds support for indexing strings (e.g., `"hello"[1]`)
- Relates the contents of a string `s` with the contents of a slice obtained with `[]byte(s)`. This fixes issue #831.
- Relates the contents of a slice `s` with the contents of the string `string(s)`. This **partially** addresses issue #832 (see my comment there).

At the moment, violating the precondition of an indexing operation (i.e., passing an invalid index to a string) leads to a bad error message, similar to `precondition of unknown may not hold`. This is similar to other parts of our encoding, where we are not able to generate good error messages when the precondition of a function internal to our encoding may not hold. PR https://github.com/viperproject/silver/pull/815 will hopefully allow us to fix that by attaching an error message to the precondition directly in the encoding.

### DIFF
--- a/src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
+++ b/src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
@@ -81,6 +81,12 @@ class StringEncoding extends LeafTypeEncoding {
       case conv@in.Conversion(in.StringT(_), expr :: ctx.Slice(in.IntT(_, TypeBounds.Byte))) =>
         val (pos, info, errT) = conv.vprMeta
         for { e <- goE(expr) } yield byteSliceToStr(e)(ctx)(pos, info, errT)
+      case e@in.IndexedExp(base, index, _: in.StringT) =>
+        val (pos, info, errT) = e.vprMeta
+        for {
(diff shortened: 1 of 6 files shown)


src/main/scala/viper/gobra/translator/encodings/StringEncoding.scala
```diff
@@ -331,4 +378,47 @@ class StringEncoding extends LeafTypeEncoding {
 
   private def byteSliceToStr(slice: vpr.Exp)(ctx: Context)(pos: vpr.Position, info: vpr.Info, errT: vpr.ErrorTrafo): vpr.FuncApp =
     byteSliceToStrFuncGenerator(Vector(slice), ())(pos, info, errT)(ctx)
+
+  /**
+   * Generates the function
+   *    requires 0 <= i && i < len(s)
+   *    decreases _
+   *    pure func stringIndexFunc(s []byte, i int) (res byte)
```
jcp19, line 386:
```suggestion
   *    pure func stringIndexFunc(s string, i int) (res byte)
```

PR #968, opened 2025-09-05
Update Submodules to '25.08-RC'

ArquintL:


marcoeilers:
I think this is the RC that basically doesn't work on Windows because of a name clash between ``defaultHeapSupporter`` and ``DefaultHeapSupporter``. So you should move to the actual release version.

PR #970, opened 2025-09-17
Derive ... by ...

ArquintL:
Introduces a new statement `derive <cond> by <block>` to derive `cond` by performing a proof by contradiction. This is particularly useful in connection with opaque functions as their body is revealed within `block` without spilling to the rest of the method as the branch evaluating `block` results in a contradiction and is, thus, killed.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Desugar.scala
+++ b/src/main/scala/viper/gobra/frontend/Desugar.scala
@@ -17,7 +17,7 @@ import viper.gobra.frontend.info.base.{BuiltInMemberTag, Type, SymbolTable => st
 import viper.gobra.frontend.info.implementation.resolution.MemberPath
 import viper.gobra.frontend.info.{ExternalTypeInfo, TypeInfo}
 import viper.gobra.reporting.Source.{AutoImplProofAnnotation, ImportPreNotEstablished, MainPreNotEstablished}
-import viper.gobra.reporting.{DesugaredMessage, Source}
+import viper.gobra.reporting.{AssertError, DeriveError, DesugaredMessage, Source}
 import viper.gobra.theory.Addressability
 import viper.gobra.translator.Names
 import viper.gobra.util.Violation.violation
@@ -4179,6 +4179,28 @@ object Desugar extends LazyLogging {
             } yield in.PredExprUnfold(predExpInstance.base.asInstanceOf[in.PredicateConstructor], args, access.p)(src)
             case _ => for {e <- goA(exp)} yield in.Unfold(e.asInstanceOf[in.Access])(src)
           }
+        case PDeriveStmt(exp, block) =>
+          /**
+            * `derive cond by block`
+            * is encoded to
+            * ```
+            * if !cond {
+            *   block
+            *   assert false
+            * }
+            * ```
+            */
+          for {
+            cond <- exprD(ctx, info)(exp)
+            b <- stmtD(ctx, info)(block)
+            // transform error of `assert false` to indicate that the derive statement failed
+            annotatedSrc = Source.Parser.Single(stmt, Source.AnnotatedOrigin(src.origin.get, new Source.OverwriteErrorAnnotation(_ match {
+              case e: AssertError => DeriveError(e.info)
+              case e => e
+            }, false)))
+            contradiction = in.Assert(in.ExprAssertion(in.BoolLit(false)(src))(src))(annotatedSrc)
+            thenStmt: in.Stmt = in.Seqn(Vector(b, contradiction))(src)
+          } yield in.If(in.Negation(cond)(cond.info), thenStmt, in.Seqn(Vector.empty)(src))(src)
         case POpenDupPkgInv() =>
           // open the current package's invariant.
           val currPkg = info.tree.originalRoot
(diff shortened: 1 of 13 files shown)


src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -4179,6 +4179,28 @@ object Desugar extends LazyLogging {
...
+            * ```
+            */
+          for {
+            cond <- exprD(ctx, info)(exp)
+            b <- stmtD(ctx, info)(block)
+            // transform error of `assert false` to indicate that the derive statement failed
+            annotatedSrc = Source.Parser.Single(stmt, Source.AnnotatedOrigin(src.origin.get, new Source.OverwriteErrorAnnotation(_ match {
+              case e: AssertError => DeriveError(e.info)
+              case e => e
+            }, false)))
```
jcp19, line 4200: This line is too long. I would probably move part of this (e.g., the case distinction on the error) to another line to make it easier to read.

src/main/scala/viper/gobra/reporting/VerifierError.scala
```diff
@@ -181,6 +181,11 @@ case class AssertError(info: Source.Verifier.Info) extends VerificationError {
   override def localMessage: String = "Assert might fail"
 }
 
+case class DeriveError(info: Source.Verifier.Info) extends VerificationError {
+  override def localId: String = "derive_error"
+  override def localMessage: String = "Derive statement failed. The body might not derive a contradiction after assuming the negated expression"
```
jcp19, line 186: I think this may lead to noisy error messages, where the "title" message is too long. It may be more consistent with the rest of Gobra to have the `localMessage` here be "Derive statement failed" and then we can introduce an ErrorReason whose message contains the rest of the message above, similarly to how it is done for assertion errors

jcp19 (commented):
In general, I am happy to introduce more high-level proof "strategies" in Gobra to make proofs clearer. However, I think we should perhaps consider how such a feature may evolve and think of a syntax that is consistent with those ideas.

First, we are introducing a new keyword "derive", which may be undesirable, and it is not clear to me why that is the case (couldn't we have an `assert ... by ...` like in Dafny or Verus). Maybe a reason for that is that this is not a "full assert", in that we cannot assert impure assertions, but I don't think that is a good enough reason to introduce an entirely new concept ("deriving"). Maybe, it would be enough to produce an error message like "Proofs by contradiction cannot be used to prove impure assertions".

Second, nothing in the syntax suggests that this proof is done by contradiction, and I would imagine that this may confuse some people. I think we could, instead, provide syntactic hints that indicate which proof strategy we want to use, like `assert P by contra { ... }`. If no strategy is selected, than by default we could do the same transformation as Dafny's "assert by". This syntax is also extensible in case we want to introduce additional strategies (possible ideas for these strategies can be found in section 30 of the Verus docs: https://verus-lang.github.io/verus/guide/reference-assert-by.html)

ArquintL:
`assert P by { L }` (the default mode) could be encoded similar to Dafny as `if(*) { L; assert P; assume false }; assume P`

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostStmtTyping.scala
```diff
@@ -23,6 +24,10 @@ trait GhostStmtTyping extends BaseTyping { this: TypeInfoImpl =>
     case PInhale(exp) => assignableToSpec(exp)
     case PFold(acc) => wellDefFoldable(acc)
     case PUnfold(acc) => wellDefFoldable(acc)
+    case n: PAssertBy =>
+      isExpr(n.exp).out ++
+        isPureExpr(n.exp) ++
```
jcp19: Checking the condition is pure makes total sense for proofs by contradiction, but I wonder if this is too strict for regular `assert-by`. At any rate, the current design is already usable and I am happy with keeping it and extending it later if needed.

src/main/scala/viper/gobra/translator/encodings/typeless/AssertionEncoding.scala
```diff
@@ -100,6 +106,60 @@ class AssertionEncoding extends Encoding {
...
 
+    case n: in.AssertByProof =>
+      // Dafny-style
+      // assert P by { L }
+      //    ~~>
+      // if(*) { L; assert P; assume false }; assume P
+      val (pos, info, errT) = n.vprMeta
+      val src = n.info
+      val nonDetChoice = in.LocalVar(ctx.freshNames.next(), in.BoolT(Addressability.exclusiveVariable))(src)
+      val errInfo = n.vprMeta._2.asInstanceOf[Source.Verifier.Info]
```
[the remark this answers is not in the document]
ArquintL: It unfortunately is

src/main/scala/viper/gobra/translator/encodings/typeless/AssertionEncoding.scala
```diff
@@ -100,6 +106,60 @@ class AssertionEncoding extends Encoding {
...
+      val (pos, info, errT) = n.vprMeta
+      val assertFalse = vpr.Assert(vpr.FalseLit()(pos, info, errT))(pos, info, errT)
+      val errInfo = n.vprMeta._2.asInstanceOf[Source.Verifier.Info]
+      for {
+        cond <- ctx.expression(n.ass.asInstanceOf[in.ExprAssertion].exp)
+        proof <- ctx.statement(n.proof)
+        thenBranch = vu.seqn(Vector(proof, assertFalse))(pos, info, errT)
+        elseBranch = vu.nop(pos, info, errT)
+        _ <- cl.errorT({
+          case e@errors.AssertFailed(_, reason, _) if e causedBy assertFalse =>
```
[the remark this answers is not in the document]
ArquintL: No I thought this is specific to the `assertFalse` assertion. But in any case, I've restructured the code quite a bit, also addressing your other comments

PR #986, opened 2026-01-18
Support for `rel` Expressions

ArquintL:
This PR adds support for `rel(e, i)` expressions, where `e` is an arbitrary expression and `i` is an int literal (either `0` or `1`), which denotes one of the two executions when reasoning hyper properties in which `e` should be evaluated.
Since Viper's `rel` expression is hidden behind a feature flag, Gobra also requires `--enableExperimentalHyperFeatures` as well as the `extended` hyper mode.

@jcp19 I've started implementing `rel` for the non-extended hyper mode as supporting this expressions in that mode should technically be possible. However, it wasn't immediately clear to me how to make it fit into the current implementation as the implementation requires, e.g., expressing a transformation to turn a hyper assertion into a unary assertion. For now, support in the `extended` hyper mode seems sufficient but please let me know if you have thoughts on supporting the non-extended hyper mode.

This PR now also removes the deprecated `silver-sif-extension` dependency and instead uses the SIF plugin that is now part of `silver`. The changes to the Dockerfile also resolve the warnings we have observed in the GitHub CI.

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,4 +1,4 @@
 # Any copyright is dedicated to the Public Domain.
 # http://creativecommons.org/publicdomain/zero/1.0/
 
-sbt.version = 1.7.2
+sbt.version = 1.12.4
--- a/src/main/antlr4/GobraLexer.g4
+++ b/src/main/antlr4/GobraLexer.g4
@@ -25,6 +25,8 @@ REFUTE      : 'refute';
 ASSUME      : 'assume';
 INHALE      : 'inhale';
 EXHALE      : 'exhale';
+CONTRA      : 'contra';
+BY          : 'by';
 PRE         : 'requires';
 PRESERVES   : 'preserves';
 POST        : 'ensures';
--- a/src/main/antlr4/GobraParser.g4
+++ b/src/main/antlr4/GobraParser.g4
@@ -63,7 +63,8 @@ ghostMember: implementationProof
 ghostStatement:
   GHOST statement  #explicitGhostStatement
   | fold_stmt=(FOLD | UNFOLD) predicateAccess #foldStatement
-  | kind=(ASSUME | ASSERT | REFUTE | INHALE | EXHALE) expression #proofStatement
+  | kind=(ASSUME | REFUTE | INHALE | EXHALE) expression #proofStatement
+  | ASSERT expression (BY CONTRA? block)? #assertStatement
   | matchStmt #matchStmt_
   | OPEN_DUP_SINV #pkgInvStatement
   ;
--- a/src/main/scala/viper/gobra/ast/frontend/Ast.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/Ast.scala
@@ -971,6 +971,15 @@ case class PExplicitGhostStatement(actual: PStatement) extends PGhostStatement w
 
 case class PAssert(exp: PExpression) extends PGhostStatement
 
+sealed trait PAssertBy extends PGhostStatement {
+  def exp: PExpression
+  def block: PBlock
+}
+
+case class PAssertByProof(exp: PExpression, block: PBlock) extends PAssertBy
+
+case class PAssertByContra(exp: PExpression, block: PBlock) extends PAssertBy
+
 case class PRefute(exp: PExpression) extends PGhostStatement
 
 case class PAssume(exp: PExpression) extends PGhostStatement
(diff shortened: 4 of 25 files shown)


src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostExprTyping.scala
```diff
@@ -127,6 +128,9 @@ trait GhostExprTyping extends BaseTyping { this: TypeInfoImpl =>
 
     case PLow(e) => isExpr(e).out
     case _: PLowContext => noMessages
+    case n: PRel =>
+      error(n, s"rel expressions are currently only supported if hyperMode '${Hyper.EnabledExtended.name}' and the '${Config.enableExperimentalHyperFeaturesOptionName}' flag are used.", config.hyperModeOrDefault != Hyper.EnabledExtended || !config.enableExperimentalHyperFeatures) ++
+      error(n, s"expected 0 or 1 to denote the major or minor execution, but got ${n.lit.lit}", n.lit.lit < 0 || 1 < n.lit.lit)
```
jcp19: I would avoid technical jargon, I think the message below would be clearer but I leave it up to you to decide:
```suggestion
      error(n, s"$Currently, only 0 and 1 are valid execution identifiers, but {n.lit.lit} was provided. ", n.lit.lit < 0 || 1 < n.lit.lit)
```

jcp19:
> However, it wasn't immediately clear to me how to make it fit into the current implementation as the implementation requires, e.g., expressing a transformation to turn a hyper assertion into a unary assertion. For now, support in the extended hyper mode seems sufficient but please let me know if you have thoughts on supporting the non-extended hyper mode.

Can you provide more details on the problem here?

ArquintL:
> Can you provide more details on the problem here?

There are several problems, such as:
- If the hyperMode is set to `off`, then we remove all hyper assertions in members that do not have proof obligations `removeLow`. `rel(e,i)` could simply be turned into `e` to preserve well-typedness of the expression but this might introduce an unsoundness, as illustrated by the following example:
  ```
  ensures rel(b, 0) && !rel(b, 1)
  func unsoundTransformation() (b bool)
  ```
- Similarly, `runExp` removes the minor execution if `onlyMajor` is desired
- And lastly, `assertion` performs the actual encoding of an expression `e`. However, my intuition is not good enough about what a "positive" or "negative unary" is supposed to be to extend the encoding. In particular, it seems that I'd have to add additional cases to support, e.g., `rel(i_V0, 0) < rel(i_V0, 1)`, which currently results in the error "Encountered unexpected node during product-program transformation of assertions" (which btw also occurs if a program contains the expression `!low(e)`)

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -834,16 +861,18 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
...
-    case "noMajor" => Hyper.NoMajor
-  }
+  ).map { Hyper.withName(_).get }
+
+  val enableExperimentalHyperFeatures: ScallopOption[Boolean] = opt[Boolean](
+    name = Config.enableExperimentalHyperFeaturesOptionName,
+    descr = s"Enables certain hyperproperty-related experimental features such as rel expressions and extended support for gotos.",
+    default = Some(ConfigDefaults.DefaultEnableExperimentalHyperFeatures),
+    noshort = true,
+  )
```
jcp19: Given that there is no more user-level documentation, I think it would be better to be precise about this:

```suggestion
  val enableExperimentalHyperFeatures: ScallopOption[Boolean] = opt[Boolean](
    name = Config.enableExperimentalHyperFeaturesOptionName,
    descr = s"Enables hyperproperty-related experimental features: rel expressions and improved support for gotos.",
    default = Some(ConfigDefaults.DefaultEnableExperimentalHyperFeatures),
    noshort = true,
  )
```

jcp19 (commented):
I am a bit confused about the "extended support for gotos" that is mentioned in the Config. Can you please clarify what it does and if the current treatment of gotos (those in the surface language of Gobra and those generated by Gobra's encoding) is sound?

ArquintL:
> I am a bit confused about the "extended support for gotos" that is mentioned in the Config. Can you please clarify what it does and if the current treatment of gotos (those in the surface language of Gobra and those generated by Gobra's encoding) is sound?

For future reference and as discussed, `--enableExperimentalHyperFeatures` does not change the SIF plugin's behavior with respect to `goto` statements as long as we enable `enableGotoLowEventEncoding`, which Gobra always does. I'll update the config description accordingly

PR #1006, opened 2026-03-03
Fix error position when output parameters are used in the postcondition of pure functions and methods

jcp19:
Currently, verifying the following function leads to a verification error with an unhelpful error message:
```
ensures res
decreases
pure func P1() (res bool) {
    return false
}
```

Message:
```
Error at: <:0:0> Postcondition might not hold.
[info] Assertion unknown might not hold.
```

This PR fixes that. I am sure most users of Gobra have seen this at some point, but it was never clear how it was triggered. @AndreaKe's experiments with Gobra helped me to finally identify the root cause: occurrences of the output variable in postconditions of pure functions are translated to vpr.Result, but the position information is not attached to this new node.

### DIFF
--- a/src/test/resources/regressions/features/purefuncs/result-var.gobra
+++ b/src/test/resources/regressions/features/purefuncs/result-var.gobra
@@ -18,3 +18,10 @@ decreases
 pure func (i I) P2() (res bool) {
     return false
 }
+
+//:: ExpectedOutput(postcondition_error:assertion_error)
+ensures b
+decreases
+pure func P3(b bool) (res bool) {
+    return false
+}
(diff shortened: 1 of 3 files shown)


ArquintL (approved):

src/test/resources/regressions/features/purefuncs/result-var.gobra
```diff
@@ -0,0 +1,20 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package debug
+
+//:: ExpectedOutput(postcondition_error:assertion_error)
+ensures res
+decreases
+pure func P1() (res bool) {
```
ArquintL, line 9: Looking at the failing unit tests, the error currently pops up on line 9 instead of 7, while this is still an improvement over line 0 ^^ I’m wondering whether it would be easy to change that

jcp19: Yeah, obviously this is not acceptable, that's why the tests do not pass. I thought I had fixed that in my tests, but turns out this relied on another bug that added the wrong position information to variables sometimes

ArquintL:
Thanks so much for both fixes!

PR #1011, opened 2026-04-01
Update Submodules

viper-admin:
* Updates ViperServer from `d97606725c3b882164ec4d86ddc5bce12fd5eb36` to `25e30449d677da1f1240f551d119b50fb8a044d7`.


jcp19:
let's put off merging this for now, I noticed some regressions in verified scion (where the verification of a few methods causes an exception now)

PR #1013, opened 2026-04-10
Add assert timeout configuration option for Silicon backend

jcp19:
## Summary
This PR adds support for configuring a timeout (in milliseconds) for assert operations performed by the Silicon backend. The timeout can be specified via the `--assertTimeout` command-line flag and is only applicable when using Silicon or ViperServer with Silicon as the backend.

## Key Changes
- Added `--assertTimeout` command-line option via Scallop configuration
- Implemented merge logic for `assertTimeout` that takes the minimum value when both configs specify a timeout (most restrictive)
- Added validation to ensure:
  - The flag is only used with Silicon-based backends
  - The timeout value is a positive integer (milliseconds)


jcp19:
Human here:

I edited the PR description, I reviewed the code, and vouch for both.

I find this option particularly useful in combination with a proposed change I discussed yesterday with Nick (@rayman2000) and Marco (@marcoeilers): to introduce a new type of error reason for proof obligations that time out. This is useful not only for humans debugging performance issues (e.g., understanding which precondition is taking forever to verify on a method-call) and to LLMs, as they would get more info on what the cause of verification error is on those cases. Ofc, we need to open a new PR to support that kind of error-reason if and when it is introduced in Viper.

EDIT: if that error reason is eventually introduced in Viper, I think it is worth having a permissive (but finite) default value for the assert timeout in Gobra (10-15s). This would make Gobra more beginner friendly, as the new users would have a clear indication that something is not ideal about their queries

src/test/resources/regressions/features/assert_timeout/dangerous_triggers.gobra
```diff
@@ -0,0 +1,21 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+// ##(--assertTimeout 1000)
+
+package asserttimeout
+
+// magic is an abstract ghost pure function (uninterpreted) used only in specification
+ghost pure func magic(i int) int
```
jcp19:
```suggestion

// magic is an abstract ghost pure function (uninterpreted) used only in specification
ghost
decreases
pure func magic(i int) int
```

jcp19:
Funnily enough, the added test does not terminate because the long verification times occur due to a push, not an assert. We could in principle configure the pushTimeout as well (silicon already supports this), but the SMT solver seems to ignore this parameter

ArquintL:
> Funnily enough, the added test does not terminate because the long verification times occur due to a push, not an assert. We could in principle configure the pushTimeout as well (silicon already supports this), but the SMT solver seems to ignore this parameter

What happens on push? Is the SMT solver automatically figuring out whether there are any inconsistencies?

ArquintL:
Is this PR now ready for review? (the description still says no)

jcp19:
It is. Unfortunately, there is no non-flaky way of testing that the flag is correctly propagated (the previous test I had led to a long push statement). Nonetheless, it would be easy to pipe this flag through Gobra.

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -1726,6 +1752,13 @@ class ScallopGobraConfig(arguments: Seq[String], isInputOptional: Boolean = fals
     noshort = true,
   )
 
+  val assertTimeout: ScallopOption[Int] = opt[Int](
+    name = "assertTimeout",
+    descr = "Sets a timeout (in milliseconds) for assert operations performed by Silicon. Does not have any effect on other backends.",
```
ArquintL:
```suggestion
    descr = "Sets a timeout (in milliseconds) for assert operations performed by Silicon. Does not have any effect on non-Silicon-based backends.",
```

ArquintL (approved):
I'd rephrase the CLI option slightly to make it clear that it _does_ have an effect for ViperServer + Silicon

PR #1018, opened 2026-04-24
Reject occurrences of output parameters in preconditions

### LINKED ISSUE #1017: Referencing return value in pre-condition does not raise error
Yesterday, I encountered an error where I referenced a return value slice in a function's precondition, and that lead to a null pointer exception. I was expected a normal error. I tried reproducing that issue, but instead found a different one! Maybe they are related, though.

The function below verifies using Gobra nightly in VS Code, but I believe it should verify as I require something about the return value:
```go
// @ requires len(r) == 1
// @ ensures acc(r)
func returnInPre() (r []int) {
	return []int{}
}
```

jcp19:
Disable output parameters from being used in preconditions, regardless of whether they are declared with `requires` or `preserves`

### DIFF
--- a/src/test/resources/regressions/issues/001017.gobra
+++ b/src/test/resources/regressions/issues/001017.gobra
@@ -1,0 +1,40 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package issue001017
+
+// Referencing a named return value in a precondition must be a type error,
+// because output parameters have no meaningful value at function entry.
+
+//:: ExpectedOutput(type_error)
+requires len(r) == 1
+ensures acc(r)
+func returnInPre() (r []int) {
+	return []int{}
+}
+
+// preserves is equivalent to requires + ensures, so it must also be rejected.
+//:: ExpectedOutput(type_error)
+preserves len(r) == 1
+func preservesOut() (r []int) {
+	return []int{}
+}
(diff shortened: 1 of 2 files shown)


src/test/resources/regressions/issues/001017.gobra
```diff
@@ -0,0 +1,40 @@
...
+// Valid: the output parameter may be referenced in the postcondition.
+ensures len(r) == 0
+func okPost() (r []int) {
+	return []int{}
+}
+
+// Valid: a precondition may reference an input parameter of the same name.
+requires len(r) == 1
+func okInParam(r []int) {
+}
```
jcp19:
```suggestion
```

src/test/resources/regressions/issues/001017.gobra
```diff
@@ -0,0 +1,40 @@
...
+	return []int{}
+}
+
+type T struct{}
+
+//:: ExpectedOutput(type_error)
+requires len(r) == 1
+func (t T) returnInPreMethod() (r []int) {
+	return []int{}
+}
```
jcp19, line 30:
```suggestion
// the check is performed on methods, as well as on functions
//:: ExpectedOutput(type_error)
requires len(r) == 1
func (t T) returnInPreMethod() (r []int) {
	return []int{}
}
```

ArquintL (approved):

PR #1019, opened 2026-04-27
Disallow conditional termination measures in ghost and pure members

jcp19:
Conditional termination measures (decreases <expr> if <cond>) are not meaningful for ghost or pure functions, methods, and interface method signatures. Reject them at type-check time with a dedicated error.


ArquintL (approved):
LGTM. I don't think I've ever used conditional termination measures ^^ but I'm wondering whether there's a case where one ever wants a different termination measure under different conditions. I'm fine with merging this PR until we have found such an example :P

jcp19:
> LGTM. I don't think I've ever used conditional termination measures ^^ but I'm wondering whether there's a case where one ever wants a different termination measure under different conditions. I'm fine with merging this PR until we have found such an example :P

Same for me. I'd rather merge this to prevent misuse, and in the cases where one might want to do that, then there are ways around it.

PR #1021, opened 2026-05-02
Add support for existential elimination with `assign such that`

jcp19:
Currently, it is very annoying to work with existential quantifiers in Gobra because there is no way to name the witness for an existentially quantified assertion and reuse it throughout the proof. This PR adds support for a new kind of statement, inspired by [Dafny's such-that assignment](https://dafny.org/latest/DafnyRef/DafnyRef#855-such-that-assignment).

With this syntax, one may write `var x T |= P(x)` to store in `x` a value of type `T` that satisfies property `P(x)`, if such value can be found by gobra. Otherwise, a verification error is thrown. For this statement to be well formed, `P` is required to be pure, i.e., to not contain any resources. This limitation comes fundamentally from Viper, which disallows having existentially quantified impure assertions.


src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -450,6 +450,17 @@ sealed trait AssertBy extends Stmt {
...
 
+/**
+  * `var x T |= { triggers } P`
+  *
+  * Introduces a fresh ghost local `v`, asserts that some value exists satisfying `cond`,
+  * and then assumes `cond` holds for the chosen witness. Encoded as
+  *   declare v; assert exists v :: cond; inhale cond
+  * where the existential is built by substituting `v` with a fresh bound variable
+  * in `cond`. `cond` is a pure boolean expression.
+  */
```
jcp19:
```suggestion
```

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -4211,6 +4212,16 @@ object Desugar extends LazyLogging {
             case w: in.MagicWand => in.ApplyWand(w)(src)
             case e => Violation.violation(s"Expected a magic wand, but got $e")
           }
+        case PAssignSuchThat(left, typ, cond) =>
+          val t = typeD(info.symbType(typ), Addressability.exclusiveVariable)(src)
+          for {
+            // Declare the fresh ghost local using the canonical name of the PIdnDef,
+            // so subsequent references to it (including inside `cond`) resolve to the
+            // same LocalVar.
```
jcp19:
```suggestion
```

src/main/scala/viper/gobra/translator/encodings/typeless/AssertionEncoding.scala
```diff
@@ -107,6 +107,30 @@ class AssertionEncoding extends Encoding {
...
     case n@ in.Exhale(ass) => for {v <- ctx.assertion(ass)} yield withSrc(vpr.Exhale(v), n)
 
+    case n@ in.AssignSuchThat(v, cond) =>
+      // `var x T |= P` is encoded as
+      //   assert exists x' : T :: P[x -> x']
+      //   inhale P
+      // The local `v` is already registered as a block-level Viper decl by the
+      // desugarer (via `declare`), so it is in scope after the statement.
+      // The existential carries `cond`'s source info so error messages show
+      // just `P` rather than the whole `var x T |= P` statement.
```
ArquintL:
```suggestion
      // The local variable `x` is already registered as a block-level Viper decl by the
      // desugarer (via `declare`), so it is in scope after the statement.
      // The existential carries `cond`'s source info so error messages show
      // just `P` rather than the whole `var x T |= P` statement.
```

src/main/scala/viper/gobra/translator/encodings/typeless/AssertionEncoding.scala
```diff
@@ -107,6 +107,30 @@ class AssertionEncoding extends Encoding {
...
+      val (condPos, condInfo, condErrT) = cond.vprMeta
+      val boundVar = in.BoundVar(v.id + "_B", v.typ.withAddressability(Addressability.boundVariable))(v.info)
```
ArquintL, line 119: Is this safe wrt. variables that potentially have the same name? What if the user writes two assign such that stmts, one introducing `x` and the other `x_B`? Doesn't this result in a clash where the existential shadows the variable declaration (not sure Viper allows this or not)

src/test/resources/regressions/features/assign-such-that/impure-rhs.gobra
```diff
@@ -0,0 +1,19 @@
+// Any copyright is dedicated to the Public Domain.
```
ArquintL, line 1: another testcase worth adding:

package assignsuchthatimpurerhs2

// The RHS must be a pure expression. A (potentially) impure assertion
// is not pure, so it must be rejected.

func f() {
	p := new(int)
	var x *int |= acc(x)
	_ = x
}

src/test/resources/regressions/features/assign-such-that/not-pure.gobra
```diff
@@ -0,0 +1,13 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package assignsuchthatnotpure
+
+pred P(x *int)
+
```
ArquintL:
```suggestion
```

src/test/resources/regressions/features/assign-such-that/witness-fail.gobra
```diff
@@ -0,0 +1,9 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package assignsuchthatwitnessfail
+
+func f() {
+	//:: ExpectedOutput(assign_such_that_error:assign_such_that_no_witness_error)
+	var x int |= x > 0 && x < 0
+}
```
ArquintL, line 9: for consistency

```suggestion
	var x int |= x > 0 && x < 0
	_ = x
}
```

src/test/resources/regressions/features/assign-such-that/redeclare.gobra
```diff
@@ -0,0 +1,15 @@
+// Any copyright is dedicated to the Public Domain.
```
ArquintL, line 1: another testcase worth adding:
```

package assignsuchthatredeclare

// Declaring a variable whose name already exists in the enclosing scope must
// be rejected by name resolution (just like an ordinary `var x int`).

func f() {
	var x int |= x > 0
	var x int |= x > 0
	_ = x
}
```

src/test/resources/regressions/features/assign-such-that/witness-fail.gobra
```diff
@@ -0,0 +1,9 @@
+// Any copyright is dedicated to the Public Domain.
```
ArquintL, line 1: another testcase worth adding to make sure that the encoding does not clash with default values!
```

package assignsuchthatwitnessnonzero

// this test makes sure that the encoding does not use Go's default values for initializing `x`
// before inhaling the expression provided by "assign such that".

func f() {
	var x int |= x != 0
    assert false
}
```

ArquintL (approved):
Thanks for implementing this long overdue feature!
LGTM. I have one question regarding potential name clashes and a few suggestions for additional testcases

src/main/scala/viper/gobra/translator/encodings/typeless/AssertionEncoding.scala
jcp19: This would not cause a clash, as all the names of variables provided by users are translated to other names at the viper level to avoid these kinds of clashes

jcp19: In particular, a variable named `x` in viper is translated to sth like `x_V1`, and thus, the binder for the existential quantifier gets the name `x_V1_B`. If a user writes the name `x_B`, then the viper name would be `x_B_V1` and the binder would be `x_B_V1_B`

src/test/resources/regressions/features/assign-such-that/witness-fail.gobra
jcp19: added test inspired on this one to `witness-nonzero.gobra`

src/test/resources/regressions/features/assign-such-that/impure-rhs.gobra
jcp19: added this to `impure-rhs2.gobra`

PR #1022, opened 2026-05-02
Represent array type lengths as BigInt

ArquintL:
This is a subset of PR #531 in an attempt to make that PR's diff eventually smaller

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/implementation/property/ConstantEvaluation.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/property/ConstantEvaluation.scala
@@ -15,19 +15,19 @@ import viper.gobra.util.Violation.violation
 
 trait ConstantEvaluation { this: TypeInfoImpl =>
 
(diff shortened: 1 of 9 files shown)


src/main/scala/viper/gobra/frontend/info/implementation/property/ConstantEvaluation.scala
```diff
@@ -15,6 +15,22 @@ import viper.gobra.util.Violation.violation
...
+  }
+
+  def evalString(exp: PExpression): String = {
+    stringConstantEval(exp).getOrElse(violation(s"expected constant string expression, but got $exp"))
+  }
+
+  def evalPerm(exp: PExpression): (BigInt, BigInt) = {
+    permConstantEval(exp).getOrElse(violation(s"expected constant permission expression, but got $exp"))
+  }
+
```
jcp19: I would just add the suffix `OrFail` to these functions to make it clear on the caller side that they lead to exceptions

jcp19 (approved):
LGTM

PR #1023, opened 2026-05-02
Encoding changes for zero-sized data

ArquintL:
This is a subset of PR #531 in an attempt to make that PR's diff eventually smaller.

This PR not only unifies the handling of zero-sized arrays and structs but also fixes the definition of what zero-sized data is. So far, we treated only `*struct{}` and `*[0]int` as pointing to zero-sized data. I.e., [`structPointersMayDiffer`](https://github.com/viperproject/gobra/blob/bd84e149fe044f9c6d863dcbf550ac755f62ab21/src/test/resources/regressions/features/zero_sized_equality/equality.gobra#L6-L10) and [`emptyArrayPointersMayDiffer `](https://github.com/viperproject/gobra/blob/bd84e149fe044f9c6d863dcbf550ac755f62ab21/src/test/resources/regressions/features/zero_sized_equality/equality.gobra#L12-L16) would fail as expected, i.e., the assertion fails as we define equality as returning unknown. All other equality checks in "src/test/resources/regressions/features/zero_sized_equality/equality.gobra" do however not correctly take the fact into account that these are comparisons on pointers to zero-sized data, as manifested in [this CI run](https://github.com/viperproject/gobra/actions/runs/25274696892/job/74102646806).
```
func structPointersMayDiffer(x, y *struct{}) {
        assert x == x && y == y // syntactical equality works
	assert x == y || x != y
}

func emptyArrayPointersMayDiffer(x, y *[0]int) {
        assert x == x && y == y // syntactical equality works
	assert x == y || x != y
}
```

This PR fixes this by recursively considering whether a type is zero-sized. I.e., types like `*[1]struct{}`, `type Z struct { a [1]struct{} }`, and `type Z2 struct { a [0]int }` are correctly considered being zero-sized too.


src/test/resources/regressions/features/zero_sized_equality/equality.gobra
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func namedZeroSizedPointersMayDiffer(x, y *Z) {
+	//:: ExpectedOutput(assert_error:assertion_error)
+	assert x == y
+}
+
+type Z struct {
+	a [1]struct{}
+}
```
jcp19, line 38: I'm a bit confused about these changes. IIRC, all of these tests already have the expected results without the changes introduced here. So, what is it that we are changing in this PR exactly?

ArquintL: I don't think I have a definite answer ^^
However, it seems to me that this PR unifies the treatment of zero-sized data (instead of special casing zero-sized arrays and zero-sized structs).
I'll have to try this but a change to me seems to be that the definition of zero-sized data is recursive (which seems to not have been the case before). Thus, we might be able to construct a testcase exhibiting differing behavior

ArquintL: Yes that was it. I've updated the PR body correspondingly to make this clearer

jcp19: Awesome, thanks for the clarification!

jcp19 (approved):
Ok, now it is clear what is going on. Thanks for this! If possible, it would be good to adapt  PR #531 to add forward links to these PRs to keep track of the currently unmerged changes

PR #1024, opened 2026-05-02
Replace `!<...!>` predicate-constructor delimiters with `{...}`

jcp19:
This PR deprecates the delimiters `!<...!>` predicate constructors. This is probably the ugliest piece of Gobra syntax.

The new proposed delimiters `{...}` clash with the struct literal syntax, but we can disambiguate them based on the symbol that is used. To avoid deep changes to the AST, this change is implemented as a parser postprocessor, which allows us to keep the kiama AST (and thus, the type-checker) completely unchanged. Alternatively, we can adapt AST too.

Besides the adapted tests, I have tried this change on [VerifiedSCION](https://github.com/viperproject/VerifiedSCION/pull/416/changes).

### DIFF
--- a/src/main/antlr4/GobraParser.g4
+++ b/src/main/antlr4/GobraParser.g4
@@ -424,7 +424,7 @@ functionLit: specification closureDecl[$specification.trusted, $specification.pu
 
 closureDecl[boolean trusted, boolean pure]:  FUNC IDENTIFIER? (signature blockWithBodyParameterInfo?);
 
-predConstructArgs: L_PRED expressionList? COMMA? R_PRED;
+predConstructArgs: L_CURLY expressionList? COMMA? R_CURLY;
 
 // Added predicate spec and method specifications
 interfaceType:
--- a/src/main/scala/viper/gobra/frontend/Parser.scala
+++ b/src/main/scala/viper/gobra/frontend/Parser.scala
@@ -538,7 +538,8 @@ object Parser extends LazyLogging {
       def run(
         pkg: PPackage,
         localPredicateNames: Set[String],
-        importedPredicateNames: String => Set[String],
+        importQualifiers: Set[String],
+        importedFPredicateNames: String => Set[String],
       ): PPackage = {
         def hasKey(lit: PLiteralValue): Boolean = lit.elems.exists(_.key.isDefined)
         def hasBlank(lit: PLiteralValue): Boolean = lit.elems.exists {
@@ -548,21 +549,46 @@ object Parser extends LazyLogging {
         def isLocalOrBuiltInPredName(name: String): Boolean =
           localPredicateNames.contains(name) || builtInPredicateNames.contains(name)
 
+        // Determines whether a `PCompositeLit` should be reinterpreted as a `PPredConstructor`.
+        //
+        // The disambiguation hinges on what `typ` could mean. Go's package-level namespace is
+        // unique, so for a single name we have at most one declaration; for a dotted name
+        // `qual.id` the meaning of `id` is determined by what `qual` is.
         def shouldRewrite(typ: PLiteralType, lit: PLiteralValue): Boolean = {
+          // Composite literals with keyed elements (`T{x: 1}`) are never predicate constructors.
           if (hasKey(lit)) false
           else typ match {
-            // Plain `P{...}`: a function predicate of the current package or a built-in fpredicate.
+            // `IDENT{...}`: by Go's name uniqueness, IDENT is at most one of {fpredicate,
+            // type, ...}. If it's a (function or built-in) predicate name, rewrite.
             case PNamedOperand(id) => hasBlank(lit) || isLocalOrBuiltInPredName(id.name)
-            // `qual.id{...}`: three flavours sharing this surface form. The `id.name` check
-            // catches local method-predicate references (`recv.isZero{...}` or
-            // `Mutex.isZero{...}`) where the qualifier names a value/type of the current
-            // package; the imported-names check catches imported function-predicate references
-            // (`pkg.P{...}`).
+
+            // `qual.id{...}` with `qual` a single identifier. Two cases.
             case PDot(qual: PNamedOperand, id) =>
-              hasBlank(lit) || isLocalOrBuiltInPredName(id.name) || importedPredicateNames(qual.id.name).contains(id.name)
-            // Other dotted bases shouldn't actually occur for `PCompositeLit` since the grammar
-            // restricts the type of a composite literal to a `typeName` (single ident or
-            // qualifiedIdent), but we keep this defensive branch for symmetry.
+              if (importQualifiers.contains(qual.id.name)) {
+                // `qual` is an import alias. `qual.id` is therefore a top-level name in the
+                // imported package, which means it is *either* a type (struct literal) *or* a
+                // top-level function predicate (predicate constructor). Imported method
+                // predicates are not in scope as `qual.id` -- they are reached via
+                // `qual.Type.id`, which doesn't fit `qualifiedIdent`. So we look only at
+                // imported fpredicate names. Importantly, we do *not* fall back to local
+                // predicate names: the local namespace is irrelevant when the qualifier is an
+                // import.
+                hasBlank(lit) || importedFPredicateNames(qual.id.name).contains(id.name)
+              } else {
+                // `qual` is a (local) value, type, or undeclared identifier. Either way
+                // `qual.id` is not a Go type expression -- Go has no nested types, struct
+                // fields are not types, and method values aren't types either -- so the
+                // composite-literal interpretation is structurally impossible. The only valid
+                // reading is a method-predicate reference (`recv.isZero{...}` or the
+                // predicate-expression form `Mutex.isZero{...}`). Rewrite unconditionally and
+                // let the type checker emit a precise error if `id` isn't actually a predicate.
+                true
+              }
+
+            // Defensive: composite-literal `typ` is grammatically restricted to
+            // `typeName: IDENT | qualifiedIdent`, so a `PDot` whose base isn't a
+            // `PNamedOperand` shouldn't reach this point. Treat like the `PNamedOperand` case
+            // above.
             case PDot(_, id) => hasBlank(lit) || isLocalOrBuiltInPredName(id.name)
             case _ => false
           }
@@ -595,16 +621,26 @@ object Parser extends LazyLogging {
       }
     }
 
-    /** Runs the rewrite over `pkg`. `localPredicateNames` is the set of top-level predicate
-      * names (function and method) declared in `pkg`. `importedPredicateNames(qual)` returns the
-      * set of predicate names exposed by the package imported under qualifier `qual`, or an
-      * empty set if `qual` is not an import qualifier or its imports aren't yet known. */
+    /** Runs the rewrite over `pkg`.
+      *
+      * @param localPredicateNames     names of every top-level predicate (function or method)
+      *                                declared in `pkg`. Used to disambiguate the unqualified
+      *                                `IDENT{...}` form.
+      * @param importQualifiers        the set of import qualifier names visible in `pkg`. Used
+      *                                to decide whether the qualifier in `qual.id{...}` refers
+      *                                to an imported package or to a local value/type.
+      * @param importedFPredicateNames `importedFPredicateNames(q)` is the set of top-level
+      *                                function-predicate names in the package imported under
+      *                                qualifier `q`. Method predicates are intentionally
+      *                                omitted: they aren't reachable as `q.id`.
+      */
     def rewrite(
       pkg: PPackage,
       localPredicateNames: Set[String],
-      importedPredicateNames: String => Set[String],
+      importQualifiers: Set[String],
+      importedFPredicateNames: String => Set[String],
     )(positions: Positions): PPackage =
-      new Impl(positions).run(pkg, localPredicateNames, importedPredicateNames)
+      new Impl(positions).run(pkg, localPredicateNames, importQualifiers, importedFPredicateNames)
   }
 
   private class SyntaxAnalyzer[Rule <: ParserRuleContext, Node <: AnyRef](tokens: CommonTokenStream, source: Source, errors: ListBuffer[ParserError], pom: PositionManager, specOnly: Boolean = false) extends GobraParser(tokens){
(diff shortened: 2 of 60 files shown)


src/test/resources/regressions/features/defunc/mutex1.gobra
```diff
@@ -10,14 +10,14 @@ pred mutexInvariant(x *int) {
 	acc(x)
 }
 
-// For now, requires parentheses surrounding exp!<...!>
-requires acc(pmutex.LockP(), _) && pmutex.LockInv() == mutexInvariant!<x!>
-ensures acc(pmutex.LockP(), _) && pmutex.LockInv() == mutexInvariant!<x!>
+// For now, requires parentheses surrounding exp{...}
```
jcp19:
```suggestion
```

src/main/antlr4/GobraParser.g4
```diff
@@ -424,7 +424,7 @@ functionLit: specification closureDecl[$specification.trusted, $specification.pu
 
 closureDecl[boolean trusted, boolean pure]:  FUNC IDENTIFIER? (signature blockWithBodyParameterInfo?);
 
-predConstructArgs: L_PRED expressionList? COMMA? R_PRED;
```
ArquintL, line 427: `L_PRED` and `R_PRED` could be removed now

src/test/scala/viper/gobra/parsing/ParserUnitTests.scala
```diff
@@ -2494,19 +2494,19 @@ class ParserUnitTests extends AnyFunSuite with Matchers with Inside {
   }
 
   test("Parser: should be able to parse a fpredicate constructor") {
-    frontend.parseExpOrFail("mutexInvariant!<x!>") should matchPattern {
+    frontend.parseExpOrFail("(mutexInvariant){x}") should matchPattern {
```
ArquintL, line 2505: please add a note that we wrap the predicate name in parenthesis as the postprocessor would otherwise treat it as a named struct as the postprocessing is based on the presence of a predicate definition with the same name (which is not the case here) and parenthesis enforce this behavior without there being a predicate definition. In addition, it would be nice to have an additional testcase where you have the same expression but without parenthesis _and_ the corresponding predicate definition unless it gets too complicated to express such a unit test

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -515,6 +516,108 @@ object Parser extends LazyLogging {
...
+        }
+        def isLocalOrBuiltInPredName(name: String): Boolean =
+          localPredicateNames.contains(name) || builtInPredicateNames.contains(name)
+
+        def shouldRewrite(typ: PLiteralType, lit: PLiteralValue): Boolean = {
+          if (hasKey(lit)) false
+          else typ match {
+            case PNamedOperand(id) => hasBlank(lit) || isLocalOrBuiltInPredName(id.name)
+            case PDot(qual: PNamedOperand, id) =>
+              hasBlank(lit) || isLocalOrBuiltInPredName(id.name) || importedPredicateNames(qual.id.name).contains(id.name)
```
ArquintL: Why does `isLocalOrBuiltInPredName(id.name)` show up here? I'm a bit confused that there's a qualifier `qual` that seems to be completely ignored

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -515,6 +516,108 @@ object Parser extends LazyLogging {
...
+            case PDot(_, id) => hasBlank(lit) || isLocalOrBuiltInPredName(id.name)
```
ArquintL: same here

src/main/scala/viper/gobra/frontend/Parser.scala
ArquintL: Oh is this for mpredicates?

src/main/scala/viper/gobra/frontend/info/Info.scala
```diff
@@ -221,6 +221,60 @@ object Info extends LazyLogging {
...
+    deps: Map[AbstractImport, ExternalTypeInfo],
+  ): PPackage = {
+    import viper.gobra.ast.frontend.PExplicitQualifiedImport
+
+    // Local predicate names (function and method): used to skip work for cases the parse-time
+    // postprocessor already handled.
+    val localPredicateNames: Set[String] = pkg.programs.flatMap(_.declarations.collect {
+      case d: viper.gobra.ast.frontend.PFPredicateDecl => d.id.name
+      case d: viper.gobra.ast.frontend.PMPredicateDecl => d.id.name
+    }).toSet
```
ArquintL, line 254: I don't understand the comment & implementation. Why do we compute `localPredicateNames` and why don't we pass the empty set as `localPredicateNames` to `PredicateConstructorRewriter`?
Second, I'm wondering what would go wrong if we would do the entire work here, i.e., drop the first pass?

ArquintL (commented):
Thanks so much for implementing this feature! I do have some comments / questions

src/main/antlr4/GobraParser.g4
jcp19: Done in 36abfba — removed `L_PRED` and `R_PRED` from `GobraLexer.g4` and the dead `R_PRED` reference from `GobraParserBase.closingBracket`.

---

src/test/scala/viper/gobra/parsing/ParserUnitTests.scala
jcp19: Added in 36abfba:

- A short comment in `ParserUnitTests` explains why these three tests parenthesise the base. The `parseExp*` helpers stop at parsing and so don't run `Info.rewritePredConstructors`; wrapping in parens forces the parser to take the `primaryExpr predConstructArgs` route and produce a `PPredConstructor` directly, so the test pins down only the parsing rule.
- New test `should parse \`name{x}\` as a composite literal at the syntax level` that asserts the *unparenthesised* `mutexInvariant{x}` parses as a `PCompositeLit` — i.e. the rewrite to `PPredConstructor` is genuinely deferred to the type-checker.

For the variant where the predicate definition is in scope and we expect the rewrite itself to fire, the natural place is the regression suite (e.g. `regressions/issues/000176.gobra`), since exercising `Info.rewritePredConstructors` requires going through `Info.checkSources` rather than the `parseExp*` helpers used here.

---

src/main/scala/viper/gobra/frontend/Parser.scala
jcp19: Yes — added a comment in 36abfba making this explicit: the `id.name` lookup catches local **method-predicate** references on the dotted form. The qualifier in `recv.isZero{...}` (or `Mutex.isZero{...}`, the predicate-expression form) is some receiver expression / type whose own meaning we can't determine syntactically, but if `isZero` matches a top-level mpredicate name in the current package it's almost certainly a predicate constructor. The `importedPredicateNames(qual.id.name)` branch handles the orthogonal case where the qualifier *is* an import alias.

---

src/main/scala/viper/gobra/frontend/Parser.scala
jcp19: Same answer — covered by the comment added in 36abfba on the `PNamedOperand` and `PDot` branches of `shouldRewrite`. The defensive `case PDot(_, id)` would only matter if the parser were ever extended to allow non-`PNamedOperand` qualifiers in a composite literal's type; today the grammar restricts it to `typeName: IDENT | qualifiedIdent`, so this branch is unreachable in practice.

---

src/main/scala/viper/gobra/frontend/info/Info.scala
jcp19: You're right on both counts — done in 36abfba.

I dropped the parse-time pass entirely and now do the whole rewrite from `Info.checkSources` (renamed `rewriteCrossPackagePredConstructors` → `rewritePredConstructors`). The two-pass design was a leftover from when I started without cross-package support: the parse-time pass handled local predicates and the type-check-time pass added imports on top. Once both kinds of names are available together via `dependentTypeInfo`, there's no reason to split them — `localPredicateNames` is collected from `pkg.programs` directly in the new single function, so the duplication you flagged is gone.

The reasoning for keeping them split was weak: the parse-time pass meant the cached `PPackage` already contained `PPredConstructor` for local cases, but every consumer of a parsed package goes through `Info.checkSources` anyway, so caching the rewrite at parse time wasn't buying anything. Single pass is simpler and more uniform.

---

ArquintL:
Don't we have a problem in the following case:
`qual.NamedStructType{ 42 }` if NamedStructType is a struct type in pkg qual while also an mpredicate in the local package?

jcp19:
> Don't we have a problem in the following case: `qual.NamedStructType{ 42 }` if NamedStructType is a struct type in pkg qual while also an mpredicate in the local package?

Can you elaborate on this? I guess in these situations, the entity resolver is responsible for picking the best candidate. The same problem exists in `qual.M()`, where `qual` may either be a local variable and `M` a method, or a qualified function with qualifier `qual`

src/main/scala/viper/gobra/frontend/Parser.scala
ArquintL: Wouldn't it be more defensive to throw an exception?

ArquintL:
I was wondering whether the Go spec makes ambiguities impossible but I guess that's not the case. The Go spec only makes top-level names (like named struct types) unique, which I guess includes fpredicates.
The cases I was going through in my head where the following (for the `PDot` case):
- pkg-local struct type: impossible as it's a dot expression
- imported struct type could look the same as
  - imported fpredicate
  - (imported or pkg-local) mpredicate

Since we check whether the `id` matches a fpredicate or mpredicate, I think we run into a problem in the following cases for `qual.iden{ 42 }`:
- if `iden` is an imported struct type but also a local fpredicate or mpredicate identifier (as we only test for `isLocalOrBuiltInPredName(id.name)`)
- if `iden` is an imported struct type but also an imported mpredicate identifier (fpredicate is imho not possible as it would be in the same package as the struct type and, thus, rejected by the type-checker due to non-uniqueness)

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -548,21 +549,46 @@ object Parser extends LazyLogging {
...
+                // reading is a method-predicate reference (`recv.isZero{...}` or the
+                // predicate-expression form `Mutex.isZero{...}`). Rewrite unconditionally and
+                // let the type checker emit a precise error if `id` isn't actually a predicate.
+                true
+              }
+
+            // Defensive: composite-literal `typ` is grammatically restricted to
+            // `typeName: IDENT | qualifiedIdent`, so a `PDot` whose base isn't a
+            // `PNamedOperand` shouldn't reach this point. Treat like the `PNamedOperand` case
+            // above.
```
ArquintL: I don't fully understand this comment. Does this comment say that a named type cannot reach this case? As a predicate constructor can (e.g., `importQualifier.NamedType.MyPred`)

ArquintL: If this interpretation is correct, shouldn't this case return `true`?

src/main/scala/viper/gobra/frontend/Parser.scala
jcp19: done

src/main/scala/viper/gobra/frontend/info/Info.scala
```diff
@@ -221,6 +221,80 @@ object Info extends LazyLogging {
...
+    // Local ADT type -> clause names, so the rewriter keeps `X.A{...}` an ADT literal even when
+    // clause `A` collides with an mpredicate name (legal: clauses and mpredicates live in
+    // different namespaces). ADTs are ghost, so unwrap `PExplicitGhostMember` too.
+    val localAdtClauses: Map[String, Set[String]] = pkg.programs.flatMap(_.declarations.collect {
+      case PTypeDef(adt: PAdtType, id) => id.name -> adt.clauses.map(_.id.name).toSet
+      case PExplicitGhostMember(PTypeDef(adt: PAdtType, id)) => id.name -> adt.clauses.map(_.id.name).toSet
+    }).toMap
+
+    // Set of import qualifiers visible in `pkg` (across all programs).
+    val importQualifiers: Set[String] = pkg.programs.flatMap(_.imports.collect {
```
ArquintL: I guess this only affects precision but wouldn't it be better to consider qualifiers per file / PProgram? Imho you cannot use a qualifier in file even if there's an import for this qualifier in a _different_ file in the same package

src/main/scala/viper/gobra/frontend/info/Info.scala
```diff
@@ -221,6 +221,80 @@ object Info extends LazyLogging {
...
+      case pi: PExplicitQualifiedImport => pi.qualifier.name
+    }).toSet
+
+    // Map: import-qualifier -> set of *function*-predicate names declared at the top level of
+    // the imported package. Method predicates are intentionally omitted: they are attached to
+    // types and so are not reachable as `qualifier.id`.
+    val importedFPredicatesByQualifier: Map[String, Set[String]] = pkg.programs.flatMap { prog =>
+      prog.imports.collect {
+        case pi: PExplicitQualifiedImport =>
```
ArquintL, line 270: what about implicitly qualified imports?

src/main/scala/viper/gobra/frontend/Parser.scala
```diff
@@ -515,6 +515,132 @@ object Parser extends LazyLogging {
     }
   }
 
+  /**
+    * Rewrites `PCompositeLit` nodes whose name resolves (syntactically) to a predicate into the
+    * equivalent `PPredConstructor`. Predicate constructors (`P{x, _}`) and composite literals
+    * (`T{x, y}`) share the same surface syntax, so the parser uniformly builds `PCompositeLit`
+    * for `IDENT { ... }` and `IDENT.IDENT { ... }` shapes; we resolve the ambiguity here, once
+    * the type checker has enough information.
```
ArquintL: This is a high-level idea but would be curious about your inputs:
Would it make sense to let the parser produce `PCompositLitOrPredConstructor` nodes, which we then turn here into either `PCompositLit` or `PPredConstructor`. I'd hope this would be a fairly easy change with the added benefit that forgetting this step or missing any node here would simply result in a type checker error as `PCompositLitOrPredConstructor` would trivially fail type checking

ArquintL (approved):
3 minor questions / remarks but I guess we're almost good with merging this 🙌

src/main/scala/viper/gobra/frontend/Parser.scala
jcp19: done

PR #1033, opened 2026-06-02
Adds Unbounded Integer Type

ArquintL:
Adds `integer` as an explicit unbounded integer type while leaving `int` and `uint` unchanged so far

Contains some changes from #1014 and #1015

@copilot summary:
This PR introduces `integer` as an explicit unbounded integer ghost type, while keeping `int` and `uint` behavior unchanged. In addition, it rejects bitwise and shift operators on values of type `integer` and adds regression tests for the new type behavior.


ArquintL:
I've reverted the last two commits as they result in several CI failures that would all require a proper type inference mechanism, i.e., the one that #1014 will introduce. In the interest of keeping this PR small, I'm avoiding this change, which means that assigning an `integer` to an `int` in ghost code won't result in a type-error for now as `int` uses unbounded semantics.

jcp19 (approved):
LGTM, but I am a bit surprised about the new syntax for `perm` being included here. That is fine by me, but we should at least add more tests (e.g., show that it produces the expected values at verification time)

PR #1036, opened 2026-06-13
Lift `asserting` expressions from Viper

jcp19:
This PR introduces `asserting a in e` expressions, where `a` is an assertion and `e` is a pure expression.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
@@ -355,7 +355,7 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
       case invoke: PInvoke => failedProp(s"The call must be $expectedCall", !isExpectedCall(invoke))
       case f: PUnfolding => validExpression(f.op)
       case f: PAsserting => validExpression(f.op)
-      case _ => failedProp(s"only unfolding/asserting expressions and the call $expectedCall is allowed")
+      case _ => failedProp(s"only unfolding/asserting expressions and the call $expectedCall are allowed")
     }
 
     validExpression(retExpr)
--- a/src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
@@ -354,7 +354,8 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
     def validExpression(expr: PExpression): PropertyResult = expr match {
       case invoke: PInvoke => failedProp(s"The call must be $expectedCall", !isExpectedCall(invoke))
       case f: PUnfolding => validExpression(f.op)
-      case _ => failedProp(s"only unfolding expressions and the call $expectedCall is allowed")
+      case f: PAsserting => validExpression(f.op)
+      case _ => failedProp(s"only unfolding/asserting expressions and the call $expectedCall is allowed")
     }
 
     validExpression(retExpr)
--- a/src/main/scala/viper/gobra/translator/encodings/typeless/AssertionEncoding.scala
+++ b/src/main/scala/viper/gobra/translator/encodings/typeless/AssertionEncoding.scala
@@ -70,6 +70,11 @@ class AssertionEncoding extends Encoding {
     case as: in.Asserting =>
       for {
         a <- ctx.assertion(as.assertion)
+        // `pure` folds any auxiliary (pure) code emitted while encoding the body
+        // into the body expression itself, keeping it inside the `vpr.Asserting`
+        // node so that `e2` is sequenced *after* `e1`. In practice, this is
+        // a defensive strategy (mirroring `unfolding`'s encoding) that is
+        // not strictly necessary at the moment.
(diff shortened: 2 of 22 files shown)


src/main/scala/viper/gobra/frontend/info/implementation/typing/StmtTyping.scala
```diff
@@ -354,7 +354,8 @@ trait StmtTyping extends BaseTyping { this: TypeInfoImpl =>
     def validExpression(expr: PExpression): PropertyResult = expr match {
       case invoke: PInvoke => failedProp(s"The call must be $expectedCall", !isExpectedCall(invoke))
       case f: PUnfolding => validExpression(f.op)
-      case _ => failedProp(s"only unfolding expressions and the call $expectedCall is allowed")
+      case f: PAsserting => validExpression(f.op)
+      case _ => failedProp(s"only unfolding/asserting expressions and the call $expectedCall is allowed")
```
ArquintL:
```suggestion
      case _ => failedProp(s"only unfolding/asserting expressions and the call $expectedCall are allowed")
```

src/main/scala/viper/gobra/translator/encodings/typeless/AssertionEncoding.scala
```diff
@@ -67,6 +67,12 @@ class AssertionEncoding extends Encoding {
         r <- ctx.expression(let.right)
       } yield withSrc(vpr.Let(l, r, exp), let)
 
+    case as: in.Asserting =>
+      for {
+        a <- ctx.assertion(as.assertion)
+        e <- pure(ctx.expression(as.in))(ctx)
```
ArquintL, line 78: Why is `pure` necessary here?

src/test/resources/regressions/features/asserting/asserting-with-acc.gobra
```diff
@@ -0,0 +1,24 @@
...
+pure func readUnfolding(x *int) int {
+	return unfolding mem(x) in *x
+}
+
+requires mem(x)
+ensures mem(x)
+func useAcc(x *int) {
+	// asserting that we hold `mem(x)`, then read through it via a pure function
+	ghost v := asserting mem(x) in readUnfolding(x)
+}
```
ArquintL, line 24: I'd add another test case checking that `asserting` actually asserts an impure assertion:

```suggestion
}

func assertingPermissions(x *int) bool {
    // asserting that we hold `mem(x)` even though we don't hold it
    //:: ExpectedOutput(assert_error:assertion_error)
    return asserting mem(x) in true
}
```

ArquintL (approved):
LGTM. I do have one tiny question out of curiosity and two small suggestions.
In addition, could you please mention in the PR what the killer use case is as I'm wondering myself ^^

jcp19:
As for the use case, I find it extremely these expressions useful to understand which facts may fail to be established in the body of a pure function (for debugging purposes), or if you need to establish intermediate facts in the body of a pure function to guide the solver.

So far, our best approach was to introduce a pure function like
```
ghost
pure
requires b
decreases
func asserting(b bool) Unit {
    return Unit{}
}
```
and then use it like the following in a pure function:
```
let _ := asserting(...) in
...
```

You can find one example in my refinement library: 
- https://github.com/viperproject/gobra-libs/blob/00adb8d55f29c1474defdd462fdce791ed866296/dicts/dicts.gobra#L81
- https://github.com/viperproject/gobra-libs/blob/00adb8d55f29c1474defdd462fdce791ed866296/refinementguard/lii.gobra#L64

ArquintL:
> As for the use case, I find it extremely these expressions useful to understand which facts may fail to be established in the body of a pure function (for debugging purposes), or if you need to establish intermediate facts in the body of a pure function to guide the solver.
> 
> So far, our best approach was to introduce a pure function like
> 
> ```
> ghost
> pure
> requires b
> decreases
> func asserting(b bool) Unit {
>     return Unit{}
> }
> ```
> 
> and then use it like the following in a pure function:
> 
> ```
> let _ := asserting(...) in
> ...
> ```
> 
> You can find one example in my refinement library:
> 
> * https://github.com/viperproject/gobra-libs/blob/00adb8d55f29c1474defdd462fdce791ed866296/dicts/dicts.gobra#L81
> * https://github.com/viperproject/gobra-libs/blob/00adb8d55f29c1474defdd462fdce791ed866296/refinementguard/lii.gobra#L64

I see, so this is bringing a form of `assert` to pure functions. This makes a lot of sense!

ArquintL:
The additional test case I've suggested fails due to a mismatch of the error: expected `assert_error:assertion_error` but got `assignment_error:permission_error`. Making sure that this is reported as an assert or asserting error would be cleaner

ArquintL:
> The additional test case I've suggested fails due to a mismatch of the error: expected `assert_error:assertion_error` but got `assignment_error:permission_error`. Making sure that this is reported as an assert or asserting error would be cleaner

@claude can you take care of this?

jcp19:
> The additional test case I've suggested fails due to a mismatch of the error: expected `assert_error:assertion_error` but got `assignment_error:permission_error`. Making sure that this is reported as an assert or asserting error would be cleaner

I wonder if this is the right place to fix it. We are backtranslating the error produced by Viper, so I guess the root cause of the wrong error kind is Viper itself

PR #1044, opened 2026-07-14
Fixes parsing of magic wands

### LINKED ISSUE #1043: Predicate-constructor braces consume magic-wand `package` proof blocks
After [PR #1024](https://github.com/viperproject/gobra/pull/1024) changed predicate-constructor syntax from `!<...!>` to `{...}`, the parser can consume a magic-wand `package` statement's proof block as though it were predicate-constructor arguments on the wand's right-hand expression.

Valid code such as `package P() --* P() {}` consequently fails during parse-tree translation with `Wrong base type for predicate constructor`.

I reproduced this on Gobra `master` at `a30ee0c3` (the first-parent history contains #1024 as `7bf9b288`).

## Minimal reproducer

```go

package issue001043

pred P()

requires P()
func test() {
	package P() --* P() {}
}
```

## Expected behavior

The final `{}` is parsed as the empty proof block of the `package` statement, and the program verifies successfully.

## Actual behavior

The regression test fails with:

```text
[parser_error] .../issues/001043.gobra:10:21: Translation of primaryExpr P() failed: Wrong base type for predicate constructor. viper.gobra.frontend.ParseTreeTranslator.visitPredConstrPrimaryExpr(ParseTreeTranslator.scala:1319)
```

## Root cause

PR #1024 made `{...}` serve as predicate-constructor arguments, but `primaryExpr` still contains this unrestricted postfix alternative:

```antlr
| primaryExpr predConstructArgs #predConstrPrimaryExpr
```

When parsing:

```go
package P() --* P() {}
```

ANTLR attaches `{}` to the immediately preceding `P()` and constructs the equivalent of `P(){}`. `ParseTreeTranslator.visitPredConstrPrimaryExpr` then rejects the base because predicate constructors accept a named predicate base (`PNamedOperand` or `PDot`), not an invocation (`PInvoke`). Parenthesized wand operands fail in the same way because the grammar accepts any `primaryExpr` before `predConstructArgs`.

Before #1024, the dedicated `!<...!>` delimiters could not be mistaken for a surrounding block, so the unrestricted postfix rule did not create this ambiguity.

## Suggested fix

Restrict the postfix predicate-constructor alternative to bases that explicitly select a predicate identifier:

```antlr
| primaryExpr DOT IDENTIFIER predConstructArgs #predConstrPrimaryExpr
```

The receiver can still be an arbitrary primary expression, preserving forms such as `(&s).P{x}` and `(T{42}).P{}`. Unparenthesized one- and two-component forms (`P{x}` and `pkg.P{x}`) already go through the composite-literal/predicate-constructor disambiguation pass.

The documented escape syntax for ambiguous imported predicates must also remain supported while retaining the same final-selector restriction for method predicates:

```antlr
| L_PAREN (primaryExpr DOT IDENTIFIER | operandName) R_PAREN predConstructArgs #parenthesizedPredConstrPrimaryExpr
```

This preserves `(P){x}`, `(pkg.P){x}`, and parenthesized method-predicate bases such as `(pkg.T.P){x}` without reintroducing the ability to attach `{...}` directly to calls, wands, or arbitrary parenthesized expressions.

## Regression coverage

The fix should check all of the following:

- `package P() --* P() {}` parses and verifies.
- `package (P() --* P()) {}` parses and verifies.
- Parenthesized predicate bases `(P){x}`, `(pkg.P){x}`, `(P){x, _, y}`, and `(pkg.T.P){x}` still parse.
- Selector-ended method-predicate constructors such as `pkg.T.P{x}`, `(&s).P{x}`, and `(T{42}).P{}` still parse.

ArquintL:
Fixes #1043

### DIFF
--- a/src/main/antlr4/GobraParser.g4
+++ b/src/main/antlr4/GobraParser.g4
@@ -412,7 +412,16 @@ primaryExpr:
   | primaryExpr arguments #invokePrimaryExpr
   | REVEAL primaryExpr arguments #revealInvokePrimaryExpr
   | primaryExpr arguments AS closureSpecInstance #invokePrimaryExprWithSpec
-  | primaryExpr predConstructArgs #predConstrPrimaryExpr
+  // This postfix form must select the predicate identifier explicitly. Requiring a final selector
+  // prevents the `{ ... }` of an enclosing construct (for example, a magic-wand package proof block
+  // after `P() --* P()`) from being consumed as predicate-constructor arguments after an arbitrary
+  // primary expression. The receiver before the final selector can still be any primary expression.
+  // Single- and two-component unparenthesized names (`P{...}` and `pkg.P{...}`) are parsed as
+  // composite literals and disambiguated later. Keep the documented parenthesized escape syntax
+  // (`(P){...}` and `(pkg.P){...}`) with the same restriction: either a bare predicate name or a
+  // primary expression with a final selected predicate identifier.
+  | primaryExpr DOT IDENTIFIER predConstructArgs #predConstrPrimaryExpr
+  | L_PAREN (primaryExpr DOT IDENTIFIER | operandName) R_PAREN predConstructArgs #parenthesizedPredConstrPrimaryExpr
   | call_op=(
   LEN
     | CAP
--- a/src/main/antlr4/GobraParser.g4
+++ b/src/main/antlr4/GobraParser.g4
@@ -412,14 +412,14 @@ primaryExpr:
   | primaryExpr arguments #invokePrimaryExpr
   | REVEAL primaryExpr arguments #revealInvokePrimaryExpr
   | primaryExpr arguments AS closureSpecInstance #invokePrimaryExprWithSpec
-  // This postfix form must select the predicate identifier explicitly. Requiring a final selector
-  // prevents the `{ ... }` of an enclosing construct (for example, a magic-wand package proof block
-  // after `P() --* P()`) from being consumed as predicate-constructor arguments after an arbitrary
-  // primary expression. The receiver before the final selector can still be any primary expression.
-  // Single- and two-component unparenthesized names (`P{...}` and `pkg.P{...}`) are parsed as
-  // composite literals and disambiguated later. Keep the documented parenthesized escape syntax
-  // (`(P){...}` and `(pkg.P){...}`) with the same restriction: either a bare predicate name or a
-  // primary expression with a final selected predicate identifier.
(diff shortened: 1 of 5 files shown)


src/test/resources/regressions/issues/001043.gobra
```diff
@@ -0,0 +1,31 @@
...
+package issue001043
+
+// tests that `package P() --* P() {}` is not parsed as a predicate constructor but as a package
+// statement with empty body.
+// in addition, this file tests that one can use predicate constructors within magic wands.
+
+pred P()
+
+requires P()
+ensures  acc(P() --* P(), 2/1)
```
ArquintL:
```suggestion
ensures  (P() --* P()) && (P() --* P())
```

src/main/antlr4/GobraParser.g4
```diff
@@ -412,7 +412,16 @@ primaryExpr:
...
   | primaryExpr arguments AS closureSpecInstance #invokePrimaryExprWithSpec
-  | primaryExpr predConstructArgs #predConstrPrimaryExpr
+  // This postfix form must select the predicate identifier explicitly. Requiring a final selector
+  // prevents the `{ ... }` of an enclosing construct (for example, a magic-wand package proof block
+  // after `P() --* P()`) from being consumed as predicate-constructor arguments after an arbitrary
+  // primary expression. The receiver before the final selector can still be any primary expression.
+  // Single- and two-component unparenthesized names (`P{...}` and `pkg.P{...}`) are parsed as
+  // composite literals and disambiguated later. Keep the documented parenthesized escape syntax
+  // (`(P){...}` and `(pkg.P){...}`) with the same restriction: either a bare predicate name or a
+  // primary expression with a final selected predicate identifier.
```
jcp19: this is super verbose and not very clear. Does this impose restrictions on the kinds of predicate instance expressions that one can construct? If so, which kinds of expressions does it disallow? Orienting the comment around that and explaining why those restrictions are in place would make it clear to me (ideally, the comment would also be shorter and more to the point)

jcp19 (approved):
LGTM, just have one question

src/main/antlr4/GobraParser.g4
ArquintL: Not sure I excelled at making it more concise but the comment clarifies now how we parse predicate constructors

jcp19: looks much better now, thanks!

jcp19:
LGTM

PR #1046, opened 2026-07-15
Fix position of result assignments

AndreaKe:
Previously, result assignments in method encodings received the meta of the method itself. I changed this such that the meta is derived from the result variable instead. This is more precise and required for the dependency analysis to work correctly.


jcp19 (approved):

jcp19:
Thank you! @AndreaKe

PR #1054, opened 2026-07-20
Resolve embedded interfaces through their owning type context's cached

### LINKED ISSUE #1053: Verifying implementations of an interface defined elsewhere in the package goes 1500x slower than putting all in the same file
## Jargon-Heavy Summary

Cross-file Go interface embedding of pred-bearing base interface causes type-checker timeout (>300s).

## Description

When an interface declared in one file embeds a `pred`-bearing base interface declared in a **different** file of the same Go package, Gobra's type-checker becomes ~1000× slower, routinely timing out at 300s+. The exact same content collapsed into a single file verifies in ~5s. This timeout occurs when specific files are being verified, and the file that defines the interfaces is not included.

This was discovered while building a large AST with dozens of node types in the file `spec.go` implementing a PNode interface. defined in the file `pnode.go`.

## Steps to Reproduce

Two files, same package `frontend`:

**`pnode.go`** — declares the base interface `PNode` with `pred mem()`:
```go
package frontend

import "go/token"

type PNode interface {
    //@ pred mem()
    //@ requires mem()
    //@ ensures  mem()
    Pos() token.Pos
    //@ requires mem()
    //@ ensures  mem()
    End() token.Pos
}

type PDecl interface {
    PNode
    //@ pred mem()
    pDecl()
}

type PExpression interface {
    PNode
    //@ pred mem()
    pExpression()
}
```

**`spec.go`** — declares 5 interfaces that each embed `PNode` across the file boundary, plus concrete types implementing them (PReceiver, PMethodRecvType, PModifier, PFunctionSpecClause, PDecreases with ~10 concrete types each having `pred mem()` and `Pos()`/`End()` methods). The full repro is ~690 lines across both files.

Run:
```bash
gobra --noVerify -I <pkgroot> -m <module> -i spec.go@<any line>
```

**Observed**: Times out at 300s+ (even with `--noVerify`, which skips the SMT solver).

**Workaround**: Merge the content of both files into a single file — verifies in ~5s with `--noVerify`, producing a genuine permission error (not a timeout).

Full reproduction files are available on the `experiment/frontend-family-split` branch of the investigation worktree, and a minimal standalone pair is at:
- `pnode.go`: 45 lines — PNode, PDecl, PExpression
- `spec.go`: 656 lines — 5 interfaces embedding PNode + 10 concrete types

These are committed in the minimal reproduction commit `d927a055`.

### Where the time is spent

The timeout is in **type-checking**, specifically `Errors.info.errors` — the lazy Kiama attribution evaluation of well-definedness rules for all AST nodes. Instrumentation confirms:

| Phase | Single-file (fast) | Cross-file (slow) |
|---|---|---|
| Parsing | 140ms | 182ms |
| `info.errors` (type-check) | **111ms** | **>300s** |
| Desugaring | 409ms | never reached |
| Viper encoding (`predicateFamilyTuple`) | 4ms | never reached |

The `--noVerify` flag skips only the backend verifier, not type-checking — which is why `--noVerify` didn't help.

### Why cross-file is slower

In the cross-file case with `-i spec.go`, only `spec.go` is processed. `PNode`, `PDecl`, and `PExpression` (declared in `pnode.go`) are **absent from the Kiama `GoTree`**. Instrumentation confirms:

| | Single-file | Cross-file |
|---|---|---|
| Interface defs in tree | **8** (PNode, PDecl, PExpression + 5 others) | **5** (only the 5 from spec.go) |
| Total type defs | 28 | 25 |

Yet all 5 interfaces embed `PNode`. Every type-checking operation that needs `PNode` must resolve it through the package-level symbol table rather than through Kiama attribution on a tree-resident node.

### The InterfaceT caching problem

`InterfaceT` (in `viper.gobra.frontend.info.base.Type`) is a case class wrapping a `PInterfaceType` and an `ExternalTypeInfo` context. It is constructed at multiple call sites:

- `TypeTyping.scala:192` — canonical creation during type inference (Kiama `attr` memoized)
- `MemberResolution.scala:214` — during `interfaceMethodSet` computation
- `NameResolution.scala:426` — during entity lookup (level 2)

Kiama's `attr` uses `ConcurrentHashMap` for memoization, which relies on **structural equality** (`equals`/`hashCode`). However, `InterfaceT`'s auto-generated `equals` compares both `decl` *and* `context`. When `PNode` is tree-resident, the Kiama `attr` on `typeSymbType` returns the **same object reference** each time — cache hits are guaranteed. When `PNode` is NOT tree-resident, each construction site creates a structurally distinct `InterfaceT` (different context, or simply different object), defeating the Kiama attr cache.

Instrumentation with creation counters confirms:
- **Single-file (fast)**: 54 `InterfaceT` instances created (all unique references), but Kiama memoization keeps `interfaceMethodSet` calls to 21
- **Cross-file (slow)**: never completes; per-node type-check is ~55ms vs 0.05ms

### Partial fix tested

Adding a `ConcurrentHashMap`-based interning cache (`mkInterfaceT`) that returns the same `InterfaceT` instance for the same `(PInterfaceType, ExternalTypeInfo)` pair, and routing the 6 construction sites through it, reduces `InterfaceT` creations from 54→12 (4.5×) and `interfaceMethodSet` calls from 21→8 in the single-file case. However, this alone does not resolve the cross-file timeout — the ~1000× per-node slowdown remains, suggesting a deeper issue with entity resolution when referenced types are not tree-resident.

The interning fix diff:
```diff
+  private val ifaceTCache: ConcurrentHashMap[PInterfaceType, InterfaceT] = new ConcurrentHashMap()
+  def mkInterfaceT(decl: PInterfaceType, context: ExternalTypeInfo): InterfaceT = { ... }

-  val res = InterfaceT(t, this)
+  val res = mkInterfaceT(t, this)     // TypeTyping.scala:192

-  InterfaceT(t, ctxt)
+  mkInterfaceT(t, ctxt)               // MemberResolution.scala:214

-  memberSet(InterfaceT(int, this))
+  memberSet(mkInterfaceT(int, this))  // NameResolution.scala:426

-  Type.InterfaceT(itfDef, context)
+  Type.mkInterfaceT(itfDef, context)  // SymbolTable.scala (2 sites)
```

## Suggested Fix

1. **Short-term**: Intern `InterfaceT` instances (diff above — correct and beneficial, just not sufficient alone).
2. **Deeper fix**: Investigate why entity resolution for non-tree-resident types is ~1000× more expensive per node. The `symbTableLookup` function in `NameResolution.scala` (4-level lookup) and the `memberSet`/`interfaceMethodSet` chain may be doing redundant work when the target type is not in the current `GoTree`. Consider making cross-file type resolution cache more aggressively, or ensuring that all package types are included in the tree even in file mode.

## Environment

- Gobra version: built from `main` (commit `6d667460` + investigation patches)
- Scala 2.13, JVM 21
- macOS arm64

ArquintL:
This PR fixes the interface method-set performance issue reported in #1053.

When Gobra computes the method set of an interface, it recursively computes the method sets of all embedded interfaces. Previously, the embedded interface type was reconstructed using the context of the interface currently being processed:

`InterfaceT(t, ctxt)`

However, resolving the embedded interface already produces a `NamedType` containing the `ExternalTypeInfo` that owns its declaration. Ignoring that context is problematic when the embedded interface is defined in another file or package.

## Why this caused the slowdown

Kiama attributes are memoized based on their input. For `interfaceMethodSet`, that input is an `InterfaceT`, which contains both:

- the `PInterfaceType` declaration;
- its `ExternalTypeInfo` context.

For interfaces inside the current syntax tree, repeated type resolution generally returns the same cached symbolic type. This allows subsequent `interfaceMethodSet` evaluations to hit Kiama's cache.

For an interface resolved through another type-checking context, the old code instead constructed a new `InterfaceT` using the caller's context. This type did not correspond to the canonical symbolic type maintained by the context that owns the declaration. Consequently, computations involving interfaces outside the current tree could use inconsistent attribute keys and repeatedly evaluate interface member resolution instead of reusing the owning type checker's cached result.

This repeated work becomes particularly expensive for interfaces containing predicates and for programs with many implementations, producing the large type-checking slowdown described in #1053.

## The fix

The embedded interface is now obtained through the context carried by the resolved `NamedType`:

`context.symbType(t).asInstanceOf[InterfaceT]`

`context.symbType(t)` is an existing Kiama attribute. It therefore returns the symbolic interface type cached by the type checker that owns the declaration.

This has two important effects:

1. The embedded interface is associated with the correct package/type-checking context.
2. Repeated method-set computations receive the same cached symbolic type and can hit Kiama's attribute cache.

The change does not depend on whether the interface is declared in the current file, another file of the package, or an imported package. Interface resolution consistently delegates to the context that owns the declaration.

## Regression test

The PR adds a regression in `regressions/issues/001053` covering an interface imported from another package. The imported interface embeds a base interface containing both a method and a predicate. The test checks that those inherited members can be resolved from file mode through the derived interface.

Generated by CodeX

Fixes #1053

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
@@ -209,8 +209,8 @@ trait MemberResolution { this: TypeInfoImpl =>
         AdvancedMemberSet.union {
           topLevel +: es.map(e => interfaceMethodSet(
             entity(e.typ.id) match {
-              // TODO: might break for imported interfaces
-              case NamedType(PTypeDef(t: PInterfaceType, _), _, _) => InterfaceT(t, ctxt)
+              case NamedType(PTypeDef(t: PInterfaceType, _), _, context) =>
+                context.symbType(t).asInstanceOf[InterfaceT]
               case _ => ???
             }
           ).promoteItf(e.typ.name))
--- a/src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
@@ -209,9 +209,11 @@ trait MemberResolution { this: TypeInfoImpl =>
         AdvancedMemberSet.union {
           topLevel +: es.map(e => interfaceMethodSet(
             entity(e.typ.id) match {
-              case NamedType(PTypeDef(t: PInterfaceType, _), _, context) =>
-                context.symbType(t).asInstanceOf[InterfaceT]
-              case _ => ???
+              case NamedType(PTypeDef(t: PInterfaceType, _), _, ownerCtxt) =>
+                ownerCtxt.symbType(t) match {
+                  case itf: InterfaceT => itf
+                  case other => Violation.violation(s"expected interface type, but got $other")
+                }
             }
           ).promoteItf(e.typ.name))
         }
--- a/src/test/resources/regressions/issues/001053.gobra
+++ b/src/test/resources/regressions/issues/001053.gobra
@@ -1,0 +1,12 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package issue001053
+
+// ##(-I ./)
+import interfaces "001053"
+
+requires x != nil && x.Mem()
+func use(x interfaces.Derived) {
+	x.M()
+}
--- a/src/test/resources/regressions/issues/001053/interfaces.gobra
+++ b/src/test/resources/regressions/issues/001053/interfaces.gobra
@@ -1,0 +1,15 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package interfaces
+
+type Base interface {
+	pred Mem()
+
+	requires Mem()
+	M()
+}
+
+type Derived interface {
+	Base
+}


ArquintL:
@budney could you please check & report whether this PR fixes the performance issue that you've reported?

jcp19 (approved):
LGTM, happy to merge it if @budney can confirm it works.

I guess it would also be cool to check whether similar problems exist elsewhere where we resolve a type against the wrong context, preventing effective caching of metadata computed when type-checking

ArquintL:
> I guess it would also be cool to check whether similar problems exist elsewhere where we resolve a type against the wrong context, preventing effective caching of metadata computed when type-checking

Good call — I audited the type checker for places that resolve a type against the wrong context or that rebuild `Type` values instead of using the owning context's cached ones. Three main takeaways, each addressed in its own follow-up PR:

1. **The caches are identity-keyed.** Kiama's `attr` memoizes on *reference identity* (weak identity map), so any freshly allocated `Type` — even one structurally equal to a cached one and carrying the right context — is a guaranteed cache miss. The canonical instance is the one returned by the owning context's memoized `symbType`. That's the deeper reason this PR's fix works: it doesn't just pick the right context, it returns the *cached object* of the owning checker.

2. **One residual correctness issue in the same code path** (predates this PR): `interfaceMethodSet` resolves embedded-interface names with the *local* `entity` attribute even when the interface declaration belongs to another context. A local struct embedding a locally-defined type whose underlying type is an imported interface (which itself embeds another interface) drives the local promotion chain into `interfaceMethodSet` on the foreign `InterfaceT` and crashes with Kiama's `NodeNotInTreeException`. Fixed in #1071 (stacked on this PR) by delegating the method-set computation to the owning context, mirroring the `memberSet` → `localMemberSet` dispatch — which also lets all importers share the owner's cache.

3. **`AdtT`/`AdtClauseT` carry a `context` but were not `ContextualType`**, so member lookups on imported ADTs ran in the importer's context and `Assignability`'s imported/local classification treated imported ADTs as local. Fixed in #1073 for consistency with structs/interfaces/domains/declared types.

Beyond these, the audit found a family of *cache-defeating but semantically benign* sites that rebuild types by hand — the same class as the slowdown fixed here: the well-definedness checks of every interface, struct, and ADT computed full member sets on throwaway `InterfaceT`/`StructT`/`AdtT` instances; the level-2 symbol-table lookup built a fresh `InterfaceT` per identifier occurrence inside interface bodies; `idSymType`/`actualTypeSymbType` allocated a fresh `DeclaredT` per occurrence of a named type (including the qualified `pkg.T` path); four sites hand-rolled `AdtClauseT`; `MethodSpec`/`MPredicateSpec.itfType` minted a second identity per interface; and `pastPromotions`/`fieldSuffix`/`adtSuffix` re-created their inner Kiama attributes on every call, so their memoization never produced a hit. All of these are addressed in #1075, which introduces canonical, memoized type factories (`declaredSymbType`, `interfaceSymbType`, `adtClauseSymbType` on `ExternalTypeInfo`, plus memoized `structSymbType`/`adtSymbType`) that bypass the well-definedness gate (calling `symbType` inside `wellDef*` of the same node would cycle) and switches all the above sites to them.

Not addressed for now: memoizing the `Property` wrappers (`identicalTypes`, `assignableTo`, … recompute on every call) — `implements` has side effects on `_requiredImplements`, so blanket memoization would change observable behavior and needs its own design plus profiling.

budney:
> LGTM, happy to merge it if @budney can confirm it works.

I ran the PR version of Gobra on my code and it worked perfectly. It looks good to me.

PR #1055, opened 2026-07-20
Fixes bound checking for slice expressions and treats string literals as byte arrays

ArquintL:
The existing bound checks rejected the following program, which is however totally fine by the Go spec, which states
> For arrays or strings, the indices are in range if 0 <= low <= high <= len(a), otherwise they are out of range. For slices, the upper index bound is the slice capacity cap(a) rather than the length. A constant index must be non-negative and representable by a value of type int; for arrays or constant strings, constant indices must also be in range. If both indices are constant, they must satisfy low <= high.

```
underlyingArray := new([0]int)
x1 := underlyingArray[0:0]
```


src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -433,35 +430,14 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
-          val lowError = error(low, s"index $low is out of bounds", !lowOpt.forall(i => i >= 0 && lenOpt.forall(i < _)))
-          val highError = error(high, s"index $high is out of bounds", !highOpt.forall(i => i >= 0 && lenOpt.forall(i < _)))
-          val lowLessHighError = (lowOpt, highOpt) match {
-            case (Some(l), Some(h)) =>
-              // this error message is the same shown by the go compiler
-              error(n, s"invalid slice index: $l > $h", l > h)
-            case _ => noMessages
-          }
-          return lowError ++ highError ++ lowLessHighError
+          arraySliceBounds(n, low, high, cap, length = stringConstantEval(base) map (_.length))
```
jcp19, line 440: I'm not sure if `_.length` is correct here. IIRC, the len of a string in go returns the number of bytes it takes; the length of a string in scala returns the number of unicode points it takes. Thus, with particular unicode strings, you may observe a difference (e.g., `len("世界")` returns 6 in Go, `"世界".length` returns 2 in scala)

jcp19 (changes requested):

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
jcp19: (I guess this error was already there, but we might as well fix it now)

src/main/scala/viper/gobra/frontend/info/implementation/typing/ExprTyping.scala
```diff
@@ -712,6 +688,31 @@ trait ExprTyping extends BaseTyping { this: TypeInfoImpl =>
...
+
+    error(low, s"index $low is out of bounds", lowConstant.exists(i => !withinBounds(i))) ++
+      error(high, s"index $high is out of bounds", highConstant.exists(i => !withinBounds(i))) ++
+      error(cap, s"index $cap is out of bounds", capConstant.exists(i => !withinBounds(i))) ++
+      // in the following, we report an error only if the components are within bounds to avoid
+      // reporting a lot of errors (like the Go compiler)
+      error(slice, "invalid slice indices: low > high",
+        lowConstant.zip(highConstant).exists { case (l, h) => if (withinBounds(l) && withinBounds(h)) l > h else false }) ++
+      error(slice, "invalid slice indices: high > cap",
+        highConstant.zip(capConstant).exists { case (h, c) => if (withinBounds(h) && withinBounds(c)) h > c else false })
```
jcp19:
```suggestion
        lowConstant.zip(highConstant).exists { case (l, h) => withinBounds(l) && withinBounds(h) && l > h }) ++
      error(slice, "invalid slice indices: high > cap",
        highConstant.zip(capConstant).exists { case (h, c) => withinBounds(h) && withinBounds(c) && h > c })
```

jcp19 (approved):
LGTM

To make sure we are using Go's utf8 rather than scala's utf16 (for strings), can you add a test checking that `len(👍) == 4` rather than 2?

PR #1064, opened 2026-08-03
Fixes an incompleteness of append and copy

### LINKED ISSUE #1061: append(/*@ perm(1), @*/ s, v) does not expose that the result's last element equals v
### Summary

`append(/*@ perm(1), @*/ s, v)` grants permission to the resulting slice but exposes no fact connecting the appended argument `v` to the resulting slice's last element — even in the most minimal possible case: plain `[]string`, no predicates, no `unfold`/`fold`, no interfaces.

### Minimal reproduction

Gobra 1.1-SNAPSHOT (`b4cdf753`).

```go
package reproducer

//@ requires forall i int :: {&s[i]} 0 <= i && i < len(s) ==> acc(&s[i], 1)
//@ ensures  len(result) == len(s)+1
//@ ensures  forall i int :: {&result[i]} 0 <= i && i < len(result) ==> acc(&result[i], 1)
//@ ensures  result[len(result)-1] == v
func appendOne(s []string, v string) (result []string) {
	result = append( /*@ perm(1), @*/ s, v)
	//@ assert result[len(result)-1] == v
	return result
}
```

Output:
```
Error at: <reproducer.go:15:2> Assert might fail.
Assertion result[len(result)-1] == v might not hold.
```

The permission facts (`len(result) == len(s)+1`, full access to every `result[i]`) verify fine — it is specifically the *value* connection between the appended argument and the resulting slice's last slot that Gobra's `append` encoding doesn't expose.

### Workaround

Pre-sized allocation plus explicit index assignment instead of `append`, whenever a slice-returning function's postcondition needs to name the value of a specific (usually last) element:

```go
//@ requires forall i int :: {&s[i]} 0 <= i && i < len(s) ==> acc(&s[i], 1)
//@ ensures  forall i int :: {&result[i]} 0 <= i && i < len(result) ==> acc(&result[i], 1)
//@ ensures  result[len(result)-1] == v
func appendOneByIndex(s []string, v string) (result []string) {
	n := len(s)
	result = make([]string, n+1)
	//@ invariant forall i int :: {&s[i]} 0 <= i && i < len(s) ==> acc(&s[i], 1)
	//@ invariant len(s) == n
	//@ invariant len(result) == n + 1
	//@ invariant forall i int :: {&result[i]} 0 <= i && i < len(result) ==> acc(&result[i], 1) && (i < j && i < len(s) ==> result[i] == s[i])
	//@ invariant 0 <= j && j <= n
	for j := 0; j < n; j++ {
		result[j] = s[j]
	}
	result[n] = v
	//@ assert result[len(result)-1] == v   // verifies
	return result
}
```

`make` + explicit index writes are ordinary field writes Gobra tracks precisely; `append`'s built-in encoding just doesn't carry the analogous fact. This workaround isn't always applicable without restructuring the caller's own loop shape (e.g. it doesn't directly transfer to a *filtering* loop whose final length isn't known upfront without a two-pass count-then-fill restructuring).

### Why this matters

This blocks a common and otherwise-natural pattern: accumulating a growing result via `append` inside a loop, then wanting to state a postcondition about which value ended up at which resulting index (e.g. "the element I just appended is provably present in the result," needed to prove coverage/completeness properties of accumulation loops). The gap forces callers to either restructure away from `append` entirely (not always straightforward) or fall back to `inhale`-ing the fact instead of proving it.

### Environment

- Gobra 1.1-SNAPSHOT, commit `b4cdf753`
- Backend: Silicon (Z3 4.8.7+)

ArquintL:
While the specification so far is complete-enough as one can simply pass a fraction of permissions and, thus, exploit framing, I nonetheless think it's useful to have stronger spec as this is a common pitfall.

Fixes #1061

### DIFF
--- a/src/main/scala/viper/gobra/translator/encodings/typeless/BuiltInEncoding.scala
+++ b/src/main/scala/viper/gobra/translator/encodings/typeless/BuiltInEncoding.scala
@@ -441,6 +441,7 @@ class BuiltInEncoding extends Encoding {
           * ensures len(res) == len(dst) + len(src)
           * ensures forall i int :: { &res[i] } 0 <= i && i < len(res) ==> acc(&res[i])
           * ensures forall i int :: { &src[i] } 0 <= i && i < len(src) ==> acc(&src[i], p)
+          * ensures forall i int :: { &src[i] } 0 <= i && i < len(src) ==> src[i] === old(src[i])
           * ensures forall i int :: { &res[i] } 0 <= i && i < len(dst) ==> res[i] === old(dst[i])
           * ensures forall i int :: { &res[i] } len(dst) <= i && i < len(res) ==> res[i] === src[i - len(dst)]
           */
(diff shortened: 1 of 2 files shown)


jcp19 (approved):
LGTM

PR #1065, opened 2026-08-03
Variadic type identity

### LINKED ISSUE #1063: Interface satisfaction check rejects identical variadic parameter as "different signature"
### Description

Gobra's interface-satisfaction check (`Implements.scala`, `syntaxImplements`) rejects a struct
method as not implementing an interface when both the interface method and the struct method
declare the **same variadic parameter**, textually identical. Replacing the variadic parameter
with an equivalent non-variadic slice parameter (`opts []CallOption` instead of
`opts ...CallOption`) makes the identical code verify with 0 errors, so the mismatch appears to
be in how variadic parameter types are compared, not a real signature difference.

### Steps to reproduce

Verify the following file (no other flags needed):

```go
package iface_probe

type CallOption interface{}

type Client interface {
	Ping(req *int, opts ...CallOption) (*int, error)
}

type client struct{}

func (c *client) Ping(req *int, opts ...CallOption) (*int, error) {
	return req, nil
}

func New() Client {
	return &client{}
}
```

### Actual result

```
error: property error: *client is not assignable to Client
reason: The member Ping has a different signature for implementation and interface
	return &client{}
 ^
```

### Expected result

0 errors — `*client`'s `Ping` method has an identical signature to `Client.Ping`, including the
variadic parameter, so it should satisfy the interface.

### Workaround / confirmation the parameter shape is the cause

Changing **both** declarations from `opts ...CallOption` to `opts []CallOption` (interface and
implementation), with no other change, verifies cleanly:

```go
package iface_probe2

type CallOption interface{}

type Client interface {
	Ping(req *int, opts []CallOption) (*int, error)
}

type client struct{}

func (c *client) Ping(req *int, opts []CallOption) (*int, error) {
	return req, nil
}

func New() Client {
	return &client{}
}
```
→ `Gobra found 0 errors.`

Also note that `//@ trusted` on `New()` does **not** suppress the error — the "is not assignable"
check appears to run during type-checking (structural `implements` checking), before the
verification phase that `trusted` skips.

### Context

Found while Gobra-annotating `protoc-gen-go-grpc`-generated client code (a `grpc.CallOption`
variadic parameter) for the Go-Gobra self-hosting rewrite project — this is a common shape in
real-world Go APIs (variadic options pattern), so the bug affects any generated or
hand-written gRPC-style client interface.

### Version

`Gobra 1.1-SNAPSHOT` (commit `b4cdf753`)

ArquintL:
So far, variadic types were not handled by our type identity logic, such that two variadic types ended up being non-identical.
This PR addresses this issue.

Fixes #1063

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/implementation/property/TypeIdentity.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/property/TypeIdentity.scala
@@ -65,6 +65,11 @@ trait TypeIdentity extends BaseProperty { this: TypeInfoImpl =>
           case (l, r) => identicalTypes(l, r)
         } && identicalTypes(lr, rr)
 
+      // regarding variadic types, the Go spec states in the "Type identity" subsection that "two function
+      // types are identical if ... parameter and result types are identical, and either both functions are
+      // variadic or neither is."
+      case (VariadicT(l), VariadicT(r)) => identicalTypes(l, r)
+
       case (PredT(larg), PredT(rarg)) =>
         larg.size == rarg.size && larg.zip(rarg).forall {
           case (l, r) => identicalTypes(l, r)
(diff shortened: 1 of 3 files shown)


jcp19 (approved):
LGTM

PR #1074, opened 2026-08-06
Optimize openInvariants set usage in loops and methods

jcp19:
I noticed some significatn slow-downs due to the introduction of this feature in `gobra-libs`. Right now, we declare a variable of `set[pred()]` type in every method and add helper invariants automatically to loops, even in methods that do not open invariants and thus, never change the set of open invariants. This PR introduces an optimization that skips the declaration of these variables and automatic generation of invariants if no invariants are open. It also fixes an issue with the positional arguments for imported mayInit functions

### DIFF
--- a/src/main/scala/viper/gobra/ast/frontend/utility/Nodes.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/utility/Nodes.scala
@@ -1,0 +1,29 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+//
+// Copyright (c) 2011-2026 ETH Zurich.
+
+package viper.gobra.ast.frontend.utility
+
+import viper.gobra.ast.frontend.PNode
+
+object Nodes {
+
+  /** Returns a list of all direct sub-nodes of this node. */
+  def subnodes(n: PNode): Seq[PNode] = {
+
+    def extractChildren(a: Any): Seq[PNode] = a match {
+      case t: PNode => Seq(t)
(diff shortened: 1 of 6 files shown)


src/main/scala/viper/gobra/ast/frontend/utility/Nodes.scala
```diff
@@ -0,0 +1,29 @@
...
+package viper.gobra.ast.frontend.utility
+
+import viper.gobra.ast.frontend.PNode
+
+object Nodes {
+
+  /** Returns a list of all direct sub-nodes of this node. */
+  def subnodes(n: PNode): Seq[PNode] = {
+
+    def extractChildren(a: Any): Seq[PNode] = a match {
```
ArquintL, line 17:
```suggestion

    // extracts all reachable `PNode`s including the current node
    def extractChildren(a: Any): Seq[PNode] = a match {
```

src/test/resources/same_package/pkg_init/mayinit_import/pkg/importer.gobra
```diff
@@ -0,0 +1,17 @@
...
+
+package pkg
+
+// ##(-I ../)
+
+import "bar"
+
+// Importing a package with a 'mayInit' member must not produce errors: the 'mayInit'
+// annotation must survive the spec-only parsing of imported packages (and, in
+// particular, must not be turned into an 'atomic' annotation).
```
ArquintL, line 12: does this test case check that it isn't turned into an `atomic` annotation?

jcp19: no, this tests that imported methods annotated with `mayInit` are interpreted as `mayInit`. There was a regression introduced in the previous PR that broke this, and we did not catch it bc we had no test for it.

ArquintL (approved):

PR #1075, opened 2026-08-06
Introduces canonical, memoized type factories for context-carrying types

ArquintL:
Follow-up to the audit @jcp19 asked for in the review of #1054; implements the cache-canonicalization findings (companions: #1071, #1073).

## Background

Kiama's attribute caches are keyed on **reference identity** (weak identity maps), so a freshly constructed `Type` value never hits the cache of a memoized attribute — even when it is structurally equal to the canonical instance and carries the correct context. Several places in the type checker reconstructed `DeclaredT`/`InterfaceT`/`StructT`/`AdtT`/`AdtClauseT` by hand because there was no cycle-safe way to obtain the canonical instance during well-definedness checking: `symbType` is gated on well-definedness (`createTyping`), so calling it from within a `wellDef*` check of the same node creates an attribute cycle. Every such hand-rolled construction paid for the full member-set/underlying-type machinery again on a throwaway cache key — the same class of problem as the #1053 slowdown.

## Changes

**Canonical factories, not gated on well-definedness:**
- `declaredSymbType(PTypeDecl)`, `interfaceSymbType(PInterfaceType)`, `adtClauseSymbType(PAdtClause)` — new on `ExternalTypeInfo`, so other contexts obtain the owning context's canonical instances; memoized per declaration node.
- `structSymbType` (per struct declaration and ghostness) and `adtSymbType` are now memoized internally. `wellDefActualType`/`wellDefStructType`/`wellDefGhostType` previously computed full method sets on throwaway `InterfaceT`/`StructT`/`AdtT` instances once per declaration — the well-definedness check of every interface, struct, and ADT paid the member-set computation at least twice.
- `interfaceSymbType`'s memoization also guarantees the `addDemandedEmbeddedInterfaceImplements` registration fires exactly once per interface.

**Call sites switched to canonical instances** (each previously allocated a fresh type, defeating the identity-keyed caches):
- `NameResolution.symbTableLookup` level 2 — ran per identifier occurrence inside interface bodies and recomputed the interface's member set every time;
- `idSymType` / `actualTypeSymbType` (`PDot`) — one fresh `DeclaredT` per occurrence of a named type, including the qualified `pkg.T` path where cross-package cache reuse matters most;
- the four hand-rolled `AdtClauseT` reconstructions (`TypeTyping`, `IdTyping`, `ExprTyping`, `GhostIdTyping`) — also removes the need to keep four copies of the construction logic in sync;
- `GhostMiscTyping.ghostMemberType`, `Desugar.desugarAllTypeDefVariants`, `Desugar.adtSelectionD` (discriminator branch, now consistent with the destructor branch), `Desugar.domainFunctionProxy`;
- `MethodSpec`/`MPredicateSpec.itfType` is now a `lazy` alias of the canonical interface type instead of minting a second identity per interface (lazy, because the entities are created before typing).

**Memoization repair in `MemberResolution`:** `fieldSuffix`, `adtSuffix`, and the `pastPromotions` closures re-created their inner Kiama attributes on every call (and on every recursive `pastDeref` step), so those memo tables could never produce a hit and were pure allocation overhead. The attributes are now created once per continuation and `pastDeref` polarity (`pastPromotedFields`/`AdtMembers`/`AdtConstructors`/`DomainFunctions`/`Methods`) and shared across all call sites.

## Verification

- Full `sbt test`: failure set identical to the unmodified base on the same machine (Carbon-without-Boogie and local scratch files only); all tracked Silicon-based tests pass.
- Smoke-checked ADT- and interface-heavy examples end-to-end.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
+++ b/src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
@@ -46,7 +46,12 @@ trait MemberResolution { this: TypeInfoImpl =>
 
   private val fieldSuffix: Type => AdvancedMemberSet[StructMember] = {
 
-    def go(pastDeref: Boolean): Type => AdvancedMemberSet[StructMember] = attr[Type, AdvancedMemberSet[StructMember]] {
+    // one attribute per `pastDeref` polarity such that the recursive calls share the attributes' caches
+    def go(pastDeref: Boolean): Type => AdvancedMemberSet[StructMember] = if (pastDeref) goPastDeref else goNoDeref
+    lazy val goNoDeref: Type => AdvancedMemberSet[StructMember] = mk(pastDeref = false)
+    lazy val goPastDeref: Type => AdvancedMemberSet[StructMember] = mk(pastDeref = true)
+
+    def mk(pastDeref: Boolean): Type => AdvancedMemberSet[StructMember] = attr[Type, AdvancedMemberSet[StructMember]] {
 
       case DeclaredT(decl, context) => go(pastDeref)(context.symbType(decl.right)).surface
       case PointerT(t) if !pastDeref => go(pastDeref = true)(t).ref
@@ -63,7 +68,7 @@ trait MemberResolution { this: TypeInfoImpl =>
 
   val structMemberSet: Type => AdvancedMemberSet[StructMember] =
     attr[Type, AdvancedMemberSet[StructMember]] {
-      case Single(t) => fieldSuffix(t) union pastPromotions(fieldSuffix)(t)
+      case Single(t) => fieldSuffix(t) union pastPromotedFields(t)
       case _ => AdvancedMemberSet.empty
     }
 
@@ -82,7 +87,12 @@ trait MemberResolution { this: TypeInfoImpl =>
     **/
   private val adtSuffix: Type => AdvancedMemberSet[AdtMember] = {
 
-    def go(pastDeref: Boolean): Type => AdvancedMemberSet[AdtMember] = attr[Type, AdvancedMemberSet[AdtMember]] {
+    // one attribute per `pastDeref` polarity such that the recursive calls share the attributes' caches
+    def go(pastDeref: Boolean): Type => AdvancedMemberSet[AdtMember] = if (pastDeref) goPastDeref else goNoDeref
+    lazy val goNoDeref: Type => AdvancedMemberSet[AdtMember] = mk(pastDeref = false)
+    lazy val goPastDeref: Type => AdvancedMemberSet[AdtMember] = mk(pastDeref = true)
+
+    def mk(pastDeref: Boolean): Type => AdvancedMemberSet[AdtMember] = attr[Type, AdvancedMemberSet[AdtMember]] {
 
       case DeclaredT(decl, context) => go(pastDeref)(context.symbType(decl.right)).surface
       case PointerT(t) if !pastDeref => go(pastDeref = true)(t).ref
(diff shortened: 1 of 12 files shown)


src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
```diff
@@ -46,7 +46,12 @@ trait MemberResolution { this: TypeInfoImpl =>
 
   private val fieldSuffix: Type => AdvancedMemberSet[StructMember] = {
 
-    def go(pastDeref: Boolean): Type => AdvancedMemberSet[StructMember] = attr[Type, AdvancedMemberSet[StructMember]] {
+    // one attribute per `pastDeref` polarity such that the recursive calls share the attributes' caches
+    def go(pastDeref: Boolean): Type => AdvancedMemberSet[StructMember] = if (pastDeref) goPastDeref else goNoDeref
+    lazy val goNoDeref: Type => AdvancedMemberSet[StructMember] = mk(pastDeref = false)
+    lazy val goPastDeref: Type => AdvancedMemberSet[StructMember] = mk(pastDeref = true)
```
jcp19, line 52: Can you add more documentation on why we need to have these two separate attrs?

src/main/scala/viper/gobra/frontend/info/implementation/typing/TypeTyping.scala
```diff
@@ -208,18 +206,33 @@ trait TypeTyping extends BaseTyping { this: TypeInfoImpl =>
...
 
         case p => violation(s"expected type, but got $n with resolved pattern $p")
       }
   }
 
-  private[typing] def structSymbType(t: PStructType, isGhost: Boolean): StructT = {
+  override def interfaceSymbType(t: PInterfaceType): InterfaceT = interfaceSymbTypeAttr(t)
+
+  /** Canonical [[InterfaceT]] per interface declaration. Not gated on well-definedness (in contrast
+    * to [[typeSymbType]]), so that well-definedness checks can obtain the canonical instance without
```
jcp19, line 221: instead of "gated", I'd probably use "guarded"

src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
```diff
@@ -240,6 +251,13 @@ trait MemberResolution { this: TypeInfoImpl =>
     go(pastDeref = false)
   }
 
+  // one memoized promotion closure per continuation such that repeated evaluations share the caches
+  private lazy val pastPromotedFields: Type => AdvancedMemberSet[StructMember] = makePastPromotions(fieldSuffix)
+  private lazy val pastPromotedAdtMembers: Type => AdvancedMemberSet[AdtMember] = makePastPromotions(adtSuffix)
+  private lazy val pastPromotedAdtConstructors: Type => AdvancedMemberSet[AdtClause] = makePastPromotions(adtConstructorSuffix)
+  private lazy val pastPromotedDomainFunctions: Type => AdvancedMemberSet[DomainFunction] = makePastPromotions(domainFunctionSuffix)
+  private lazy val pastPromotedMethods: Type => AdvancedMemberSet[TypeMember] = makePastPromotions(pastPromotionsMethodSuffix)
```
jcp19, line 259: is it really worth making these values lazy? I guess that, in the course of verifying regular programs, we will always cause these values to be evaluated, so maybe `lazy` is not worth it

jcp19 (commented):
Thank you so much for looking into this! I was aware that there was some degree of redundant computations / technical debt in the type checker, but the amount of opportunities for caching surprised me

PR #1077, opened 2026-08-07
Fix incompleteness with `nil` checking change

jcp19:
PR #531 makes Gobra report a possible panic whenever a program dereferences a pointer that may be nil. The checks it added were too strict, and rejected correct programs (see the comments in that PR for examples). This PR changes what is checked and where.

## Correct programs were rejected
```go
requires  8 <= len(raw)
preserves forall i int :: { &raw[i] } 0 <= i && i < len(raw) ==> acc(&raw[i])
func DecodeFromBytes(raw []byte) {
	assert forall i int :: 0 <= i && i < len(raw[2:4]) ==>
		&raw[2:4][i] == &raw[2 + i]     // The receiver raw[2:4][i] might be nil
}
```
We own every element of raw, and nothing here is even a pointer that could be nil. The old check asked whether the address of the element is non-nil. Gobra computes that address by combining the offset of the sub-slice with the index, and the verifier could not relate the result to the permissions we hold, which are written in terms of raw[i].

The new check asks a simpler question: which pointer is dereferenced here, and can it be nil? In the example, none is — raw is a variable, and a slice already carries its own array — so no proof obligation is generated at all. When a pointer is dereferenced, as in p.f, the obligation is just p != nil, which is normally immediate from the surrounding specification.

## Checks were reported in specifications
```go
pred (s *Raw) Mem(buf []byte) {
	s.Base.Mem() &&                     // The receiver s.Base might be nil
	acc(&s.Raw) && ...
}
```
A predicate definition is not executed, so it cannot panic — and a predicate about a nil receiver can never be folded in the first place. The check nevertheless fired here, before the definition had mentioned any permission. Putting checks into specifications also changed the shape of the formulas the prover sees, which made unrelated proofs fail elsewhere. Checks are now generated only in code that actually runs; bodies of pure functions keep theirs, since a pure function that is not ghost is real code. In the future, we could disable the checks in the bodies of pure ghost functions, **but we don't do it atm**.


src/main/scala/viper/gobra/translator/encodings/defaults/DefaultPureMethodEncoding.scala
```diff
@@ -53,7 +53,7 @@ class DefaultPureMethodEncoding extends Encoding {
       body <- option(meth.body map { b =>
         pure(
           for {
-            results <- ctx.expression(b)
+            results <- ctx.withoutNilChecks.expression(b)
```
jcp19: shouldn't this have nil checks in the case of non-ghost pure functions? If so, add a test for this case

src/main/scala/viper/gobra/translator/encodings/defaults/DefaultPureMethodEncoding.scala
```diff
@@ -99,7 +99,7 @@ class DefaultPureMethodEncoding extends Encoding {
       body <- option(func.body map { b =>
         pure(
           for {
-            results <- ctx.expression(b)
+            results <- ctx.withoutNilChecks.expression(b)
```
jcp19: shouldn't this have nil checks when `func` is not ghost?

src/test/resources/regressions/issues/000491-4.gobra
```diff
@@ -0,0 +1,41 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package issue491_4
+
+// Tests that usages of L-values that cannot panic due to a nil dereference
+// do not generate nil-ness proof obligations. Earlier versions of the nil
+// checks asserted non-nilness of the address of the entire L-value, which
+// the verifier often cannot prove for compound locations such as elements
+// of a resliced slice (see the discussion in PR #531).
```
jcp19:
```suggestion
// of a resliced slice (see the comments in PR #531).
```

src/test/resources/regressions/issues/000491-4.gobra
```diff
@@ -0,0 +1,41 @@
...
+	x int
+	y int
+}
+
+// Taking the address of a field only requires the root pointer to be non-nil.
+requires p != nil
+func addrOfField(p *Pair) (r *int) {
+	r = &p.x
+	return
+}
```
jcp19: add the same test without the precondition to make sure the error is reported

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
```diff
@@ -506,23 +506,71 @@ object TypeEncoding {
...
+    *
+    * Using an L-value panics if any pointer dereferenced along its chain is nil, but it suffices to
+    * check the outermost dereference, i.e. the last one performed when evaluating the L-value. The
+    * pointers dereferenced before it are read out of memory to form the chain, and reading them
+    * requires permission to the corresponding fields, which already entails that they are non-nil.
+    * The outermost dereference is the only one without such a witness, as the memory it refers to
+    * is not necessarily read. Panics that are not caused by dereferencing nil (e.g. an out-of-bounds
+    * index) are covered by separate checks. Thus, an obligation is generated only if `loc` contains
+    * a dereference, in which case we check the pointer of the outermost one:
     *
```
jcp19: Comments are too large. Make them clearer while more concise.

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
```diff
@@ -506,23 +506,71 @@ object TypeEncoding {
...
+    }
+    ctx.expression(in.UneqCmp(
+      d.exp,
+      in.NilLit(in.PointerT(d.typ, Exclusive))(annotatedInfo)
+    )(annotatedInfo))
+  }
+
+  /**
+    * Returns the outermost dereference of an L-value chain, if any, i.e. the last dereference that
+    * is performed when the L-value is evaluated.
```
jcp19: what do you mean by "last dereference"? Isn't this actually the first dereference that would be evaluated when evaluating the entire L-value?

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
```diff
@@ -506,23 +506,71 @@ object TypeEncoding {
...
+    * L-values without a dereference cannot cause a nil dereference: they are rooted in a variable,
+    * or in an exclusive value (e.g. a slice value as the base of an index expression), for which an
+    * in-bounds index implies that the accessed element exists.
+    */
+  @tailrec
+  private def nilCheckSource(l: in.Expr): Option[in.Deref] = l match {
+    case d: in.Deref => Some(d)
+    case l: in.FieldRef => nilCheckSource(l.recv)
+    case l: in.IndexedExp => nilCheckSource(l.base)
+    case _ => None
```
jcp19: enumerate all cases instead of a catch-all

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
```diff
@@ -506,23 +506,71 @@ object TypeEncoding {
...
 
   /**
-    * Checks whether an L-value is safe, i.e. does not cause a runtime panic due to dereferencing nil.
+    * Encodes the condition that the pointer of a dereference is not nil.
+    *
+    * [p != nil]    where d = *p
+    *
+    * The annotation is attached to the information of `p` and not of `d`, such that the error
+    * message names the pointer that might be nil. The position of `d` spans the entire selector
+    * when the dereference is implicit, as in `p.f`.
```
jcp19: this equality is confusing, `d` is already a Deref

ArquintL:
> predicate about a nil receiver can never be folded in the first place

Why is that? Is this something that Gobra simply enforces (even though we technically won't have to)?

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
```diff
@@ -341,7 +341,7 @@ trait TypeEncoding extends Generator {
     * Instead of checking that a dereference is safe immediately, the encoding checks that usages of l-values are safe.
     * Usages of L-values are: (1) taking a reference, (2) taking a slice, (3) converting to R-value
     *
-    * SafeRef[loc: T@] => assert [&loc != nil: *T°]; Ref[loc]
+    * SafeRef[loc: T@] => assert [p != nil]; Ref[loc]    where *p is the root of loc (if any, see checkNotNil)
```
ArquintL, line 344: what's the "root"?

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
```diff
@@ -505,23 +505,65 @@ object TypeEncoding {
...
+    // The receiver of a field access and the base of an index expression are exclusive values,
+    // which are not dereferenced. An in-bounds index implies that an element exists.
+    case _: in.FieldRef | _: in.IndexedExp => None
+    // Variables are not dereferenced.
+    case _: in.Var => None
+  }
+
+  /**
+    * Encodes the non-nilness of the address of an L-value. Used as the footprint of zero-sized types
+    * (which have no permission footprint), not as a runtime-panic check.
```
ArquintL, line 566: What does it mean to use this function as the footprint of zeros-zed types and not as a runtime-panic check?

src/test/resources/regressions/issues/000491-4.gobra
```diff
@@ -0,0 +1,48 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package issue491_4
+
+// Tests that usages of L-values that cannot panic due to a nil dereference
+// do not generate nil-ness proof obligations. Earlier versions of the nil
+// checks asserted non-nilness of the address of the entire L-value, which
+// the verifier often cannot prove for compound locations such as elements
+// of a resliced slice (see the comments in PR #531).
```
ArquintL, line 10: Since `*[N]byte` behaves differently from `[]byte`, I'd like to see this being covered in the test cases

ArquintL (commented):
First batch of comments

jcp19:
> > predicate about a nil receiver can never be folded in the first place
> 
> Why is that? Is this something that Gobra simply enforces (even though we technically won't have to)?

It is a convention; we could definitely consider alternative designs.
As it stands, it is something we must enforce (and I believe we do). If you have a `nil` of an interface value `i`, you can never fold `i.Mem()` for the same reason you can never call a method `i.M()` on it: you do not have a concrete implementation of `Mem` to resolve to.

src/test/resources/regressions/issues/000491-4.gobra
ArquintL: --> #1094

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
jcp19: This was a stale comment; It's the outermost dereference, i.e. the last one evaluated (for l.next.val, the dereference of l.next) — what outermostDeref returns. Reworded.

src/main/scala/viper/gobra/translator/encodings/combinators/TypeEncoding.scala
jcp19: A location of a zero-sized type (struct{}, [0]int, …) occupies no memory, so no permission can constitute its footprint — acc(x) would degenerate to true and an allocated *struct{} would be indistinguishable from a nil one. Gobra therefore defines the footprint of such a location to be the non-nilness of its address. So this is not an obligation imposed on a usage of the L-value (that is the other checkNotNil, which asserts before a use); it is inhaled and exhaled where the permissions of a non-zero-sized type would be: inhaled by initialization/allocate and the New encoding, exhaled and re-inhaled by assignment, and exhaled wherever acc(loc) is exhaled via ctx.footprint. The two functions used to be one, because the panic check was previously stated over &loc; now that it checks the dereferenced pointer instead, they share nothing but the name, so this one is renamed to addressNotNil and documented as above.

jcp19:
@ArquintL do you oppose to merging this?

PR #1085, opened 2026-08-10
Desugar map keys to the key type of the map

### LINKED ISSUE #1081: Granting acc() to a package-level map[Interface]*Struct var crashes Silicon (assertNotNegative sort mismatch)
Granting `acc()` permission to a package-level `map[Interface]*Struct` var crashes Silicon with an internal `AssertionError` rather than either verifying or failing cleanly, when the map's key type is a plain (non-ghost) interface — one with real methods but no interface-level ghost predicate of its own.

## Symptoms

A single-key comma-ok read of a package-level map var fails cleanly with a normal permission error when no `acc` precondition is given (expected). But once the caller supplies `//@ requires acc(mapVar, _)` so the read itself has permission, verification does not produce a normal proof failure — it crashes the verifier:

```
java.lang.AssertionError: assertion failed: Expected first operand b_V0@3@17 to be of sort Tuple2[Ref, Types], but found ShStruct1[Ref].
	at scala.Predef$.assert(Predef.scala:279)
	at viper.silicon.state.terms.utils$.assertSort(Terms.scala:2594)
	at viper.silicon.decider.DefaultDeciderProvider$decider$.assert(Decider.scala:386)
	at viper.silicon.rules.permissionSupporter$.assertNotNegative(PermissionSupporter.scala:24)
```

The crash is reported against the *caller's* precondition check (`Precondition of call lookup(b) might not hold`) and cascades: other, unrelated members being verified in the same batch subsequently report "did not terminate" or fail with the same internal exception, so it isn't confined to the one method.

## Minimal reproduction

```go
package repro

// Node mirrors go/ast.Node: a plain (non-ghost) interface with real methods,
// no interface-level ghost predicate. Concrete implementers get their own
// mem() predicate separately.
type Node interface {
	//@ decreases
	//@ pure
	Pos() int
	End() int
}

type BlockStmt struct {
	List []int
}

//@ pred (b *BlockStmt) mem() { acc(b) && acc(&b.List) }

//@ decreases
//@ pure
func (b *BlockStmt) Pos() int { return 0 }

func (b *BlockStmt) End() int { return 0 }

type Metadata struct {
	Block *int
}

var ctxMetadata map[Node]*Metadata

//@ requires acc(ctxMetadata, _)
//@ ensures ok ==> md != nil
//@ ensures ok && md != nil ==> acc(&md.Block, _)
func lookup(b *BlockStmt) (md *Metadata, ok bool) {
	if ctxMetadata == nil {
		return nil, false
	}
	md, ok = ctxMetadata[b]
	return md, ok
}

//@ requires b != nil ==> (acc(b, _) && b.mem())
func caller(b *BlockStmt) {
	_, _ = lookup(b)
}
```

Verifying this crashes with the AssertionError above at `lookup`'s call site inside `caller`.

### What was tried

- Removing the `//@ requires acc(ctxMetadata, _)` precondition entirely: no crash, but a normal (expected) `Permission to ctxMetadata might not suffice` error instead — confirms the crash is specifically triggered by granting the map-var permission, not by the map read syntax itself.
- Adding an extra interface-typed field to `Metadata` (the map's value struct): still crashes identically — the map value's shape does not appear to matter.
- An earlier, simpler attempt where the map's key interface declared its own `//@ pred mem()` directly (rather than plain, non-ghost `Pos()`/`End()` methods, with `mem()` living only on the concrete implementer as in the reproduction above) did **not** crash — it either verified or failed with an ordinary `isComparable` error depending on the precondition given. I did not fully bisect which specific difference between that interface shape and `go/ast.Node`'s real shape (used above) matters, but the reproduction above is confirmed minimal in the sense that removing the `acc` precondition avoids the crash, and it reproduces with no external dependencies.

## Environment

- Gobra 1.1-SNAPSHOT (`b4cdf753`)
- Discovered while removing `@trusted` from a native-map-reading helper function in a larger Go-Gobra self-hosting rewrite project (`internal/ast/desugar/desugar.go`'s `lookupBlockMetadata`, reading `map[go_ast.Node]*GobraMetadata`) — the reproduction above reduces that real case to a self-contained ~40-line snippet with no external stub dependencies, confirmed to hit the identical assertion message and stack trace as the original.

## Possibly related

- #893 (crashes involving maps with reference-typed values/comma-ok) — different exact crash signature (`NoSuchElementException` / `LogicException: got unexpected type InternalSingleMulti`), but same general area: map encoding interacting badly with reference-typed map values.
- #278 (maps whose key type has ghost fields) — different mechanism (ghost fields specifically), but same `mapencoding` territory; our key type has no ghost fields, just a plain-Go-interface type with real (non-ghost) methods and no interface-level predicate.

Neither issue's reproduction matches this crash's signature, so filing as a new report rather than a duplicate.

jcp19:
Fixes #1081.

## Problem

Gobra never inserted the implicit conversion to an interface type for a map key. When the key type of a map is an interface type and the key expression has a concrete type, the desugared program keeps the key at its concrete type, and the encoding then compares it against values of the encoded interface type. The resulting Viper program is ill-typed, and Silicon crashes with an internal `AssertionError` instead of verifying the program or reporting a verification error. Because the exception aborts the run, all remaining members are reported as "did not terminate".

Minimal reproduction (from #1081): a `map[Node]*Metadata` read with a `*BlockStmt` key gives

```
java.lang.AssertionError: assertion failed: Expected first operand b_V0@3@03 to be of sort Tuple2[Ref, Types], but found ShStruct1[Ref].
	at viper.silicon.state.terms.utils$.assertSort(Terms.scala:2594)
	at viper.silicon.state.terms.SetIn$.apply(Terms.scala:1843)
	at viper.silicon.rules.evaluator$.$anonfun$evalBinOp$1(Evaluator.scala:1207)
	...
```

`Tuple2[Ref, Types]` is the encoding of an interface value and `ShStruct1[Ref]` the encoding of `*BlockStmt`. The `SetIn` term comes from `MapEncoding`, which encodes `m[k]` as `[k] in domain([m].underlyingMapField) ? ... : [dflt]`: the domain of the underlying Viper map has interface-typed elements, while the encoded key has the type of the concrete key.

The crash only shows up once the caller has permission to the map, because otherwise the field read fails before the offending term is built — which matches the bisection reported in the issue.

## Changes

`Desugar` already has `implicitConversion`, which is applied to assignments, call arguments, return values and composite literal elements. This PR additionally applies it to:

* the index of a map index expression (`indexedExprD`), covering reads, comma-ok reads and map updates, for both `map` and `dict`;
* the left-hand side of a membership expression (`PElem`), covering `k elem m` as well as sequences, sets and multisets whose elements have an interface type;
* keys of map literals that are not composite values (`handleMapEntries`) — composite-value keys already went through `compositeValD`.

Two small helpers, `indexType` and `elementType`, return the relevant type of a collection. Where the target type is not an interface type, `implicitConversion` is the identity, so nothing changes for existing programs.

## Testing

* The reproduction from the issue no longer crashes; it now reports a regular verification error for the postcondition that does not hold. The crash reproduces on the parent commit with the exact stack trace above.
* New regression test `src/test/resources/regressions/issues/001081.gobra` covers lookups, comma-ok lookups, membership, updates and literals with keys of a concrete type, for both `map` and `dict`, and checks that a concrete key denotes the same key as the corresponding interface value.
* `testOnly viper.gobra.GobraTests` is green on CI for both commits of this PR.

### DIFF
--- a/src/main/scala/viper/gobra/frontend/Desugar.scala
+++ b/src/main/scala/viper/gobra/frontend/Desugar.scala
@@ -2481,6 +2481,26 @@ object Desugar extends LazyLogging {
       }
     }
 
+    /**
+      * Returns the type of the elements that `typ` is indexed with, i.e., the key type of a (mathematical) map.
+      * Returns None for all other types (in particular for types that are indexed with integers).
+      */
+    def indexType(typ: in.Type): Option[in.Type] = underlyingType(typ) match {
+      case t: in.MapT => Some(t.keys)
+      case t: in.MathMapT => Some(t.keys)
+      case _ => None
+    }
+
+    /** Returns the type of the elements contained in `typ`, if `typ` is a collection type. */
+    def elementType(typ: in.Type): Option[in.Type] = underlyingType(typ) match {
+      case t: in.SequenceT => Some(t.t)
+      case t: in.SetT => Some(t.t)
+      case t: in.MultisetT => Some(t.t)
+      case t: in.MapT => Some(t.keys)
+      case t: in.MathMapT => Some(t.keys)
+      case _ => None
+    }
+
     def singleAss(left: in.Assignee, right: in.Expr)(info: Source.Parser.Info): in.SingleAss = {
       in.SingleAss(left, implicitConversion(right.typ, left.op.typ, right))(info)
     }
@@ -2561,7 +2581,10 @@ object Desugar extends LazyLogging {
         dbase <- exprD(ctx, info)(base)
         dindex <- exprD(ctx, info)(index)
         baseUnderlyingType = underlyingType(dbase.typ)
-      } yield in.IndexedExp(dbase, dindex, baseUnderlyingType)(src)
+        // the index of a map may require an implicit conversion, e.g., when the key type of the map is an
+        // interface type and `index` has a concrete type
+        dkey = indexType(baseUnderlyingType).fold(dindex)(implicitConversion(dindex.typ, _, dindex))
+      } yield in.IndexedExp(dbase, dkey, baseUnderlyingType)(src)
     }
 
     def indexedExprD(expr : PIndexedExp)(ctx : FunctionContext, info : TypeInfo) : Writer[in.IndexedExp] =
@@ -3282,7 +3305,7 @@ object Desugar extends LazyLogging {
       sequence(
         lit.elems map {
           case PKeyedElement(Some(key), value) => for {
-            entryKey <- key match {
+            dKey <- key match {
               case v: PCompositeVal => compositeValD(ctx, info)(v, keys)
               case k: PIdentifierKey => info.regular(k.id) match {
                 case _: st.Variable => unit(varD(ctx, info)(k.id))
@@ -3290,6 +3313,9 @@ object Desugar extends LazyLogging {
                 case _ => violation(s"unexpected key $key")
               }
             }
+            // keys that are not composite values do not go through `compositeValD` and, thus, may still require
+            // an implicit conversion, e.g., when the key type is an interface type
+            entryKey = implicitConversion(dKey.typ, keys, dKey)
             entryVal <- compositeValD(ctx, info)(value, values)
           } yield (entryKey, entryVal)
 
@@ -4543,10 +4569,13 @@ object Desugar extends LazyLogging {
         case PElem(left, right) => for {
           dleft <- go(left)
           dright <- go(right)
+          // the left-hand side may require an implicit conversion, e.g., when the elements of the collection have
+          // an interface type and `left` has a concrete type
+          delem = elementType(dright.typ).fold(dleft)(implicitConversion(dleft.typ, _, dleft))
         } yield underlyingType(dright.typ) match {
-          case _: in.SequenceT | _: in.SetT => in.Contains(dleft, dright)(src)
-          case _: in.MultisetT => in.LessCmp(in.IntLit(0)(src), in.Contains(dleft, dright)(src))(src)
-          case _: in.MapT => in.Contains(dleft, dright)(src)
+          case _: in.SequenceT | _: in.SetT => in.Contains(delem, dright)(src)
+          case _: in.MultisetT => in.LessCmp(in.IntLit(0)(src), in.Contains(delem, dright)(src))(src)
+          case _: in.MapT => in.Contains(delem, dright)(src)
           case t => violation(s"expected a sequence or (multi)set type, but got $t")
         }
 
--- a/src/main/scala/viper/gobra/frontend/Desugar.scala
+++ b/src/main/scala/viper/gobra/frontend/Desugar.scala
@@ -2482,23 +2482,37 @@ object Desugar extends LazyLogging {
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/test/resources/regressions/features/maps/maps-interface-key1.gobra
```diff
@@ -0,0 +1,110 @@
...
+	return 0
+}
+
+type Metadata struct {
+	block int
+}
+
+// lookups with a key of a concrete type
+
+requires acc(m, _)
```
jcp19: The permission amount in specs of pure functions is now inconsequential. Drop `_` altogehter. Apply the same transformation throughout

src/test/resources/regressions/issues/001081.gobra
```diff
@@ -0,0 +1,110 @@
+// Any copyright is dedicated to the Public Domain.
```
jcp19, line 1: rename this file after the issue that this PR is adressing

src/test/resources/regressions/features/maps/maps-interface-key1.gobra
jcp19: Done in 34ebe49 — dropped `_` from all four `acc(m, _)` in the file.

One thing to flag, since your rationale mentions pure functions but the comment was anchored on `lookup1`: of the four, only `lookup3` is pure. I read "apply the same transformation throughout" literally and changed the three non-pure ones (`lookup1`, `lookup2`, `lookup4`) too, which does strengthen their preconditions from a wildcard to full permission. They still verify, and the regression only needs *some* permission to the map for the crash to be reachable, so the test is unaffected either way — but it no longer exercises the read-under-wildcard case. Happy to put `_` back on the non-pure ones if you'd rather keep that coverage.

---

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -2481,6 +2481,26 @@ object Desugar extends LazyLogging {
       }
     }
 
+    /**
+      * Returns the type of the elements that `typ` is indexed with, i.e., the key type of a (mathematical) map.
+      * Returns None for all other types (in particular for types that are indexed with integers).
+      */
```
ArquintL, line 2487: Why not return the index type for any indexable collection? I.e., include slices, arrays, sequences, ...

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -2481,6 +2481,26 @@ object Desugar extends LazyLogging {
...
+    def indexType(typ: in.Type): Option[in.Type] = underlyingType(typ) match {
+      case t: in.MapT => Some(t.keys)
+      case t: in.MathMapT => Some(t.keys)
+      case _ => None
+    }
+
+    /** Returns the type of the elements contained in `typ`, if `typ` is a collection type. */
```
ArquintL, line 2496: I think it would be cleaner if this function operates on any collection type, i.e., include, e.g., slices and arrays

jcp19:
@ArquintL incorporated your feedback, should be ready to be merged

ArquintL (approved):

Issue #893, opened 2025-03-14
Gobra crashes in some situations involving maps


henriman:
I am working on verifying SIF for the VerifiedSCION router (using hyperGobra) with @jcp19, however this bug is unrelated to hyperGobra and also occurs using "normal" Gobra.

I believe these are two separate, but related bugs.

## 1. Using "comma ok" idiom with `let` crashes when trying to subsequently access `ok`

```go
requires acc(inputMap)
requires let x, ok := inputMap[0] in ok
ensures res
func test2(inputMap map[int]int) (res bool) {
    x, ok := inputMap[0]
    return ok
}
```

Trying to verify the above code crashes with `java.util.NoSuchElementException: key not found: ok_V1`, as there is no definition for `ok` in the Viper encoding. Note that accessing `x` in `let` works.

This can be worked around by instead using `0 in domain(inputMap)`, of course.

## 2. Comparing result of map access of a map with pointer-type elements crashes

```go
requires acc(inputMap) && (0 in domain(inputMap) ==> inputMap[0] != nil)
func test3(inputMap map[int](*int)) int

requires acc(inputMap) && (0 in domain(inputMap) ==> inputMap[0] != 0)
func test4(inputMap map[int](int)) int
```

While verifying `test4` succeeds, verifying `test3` crashes with:

```
15:10:05.519 [main] ERROR viper.gobra.GobraRunner$ - An assumption was violated during execution.
15:10:05.521 [main] ERROR viper.gobra.GobraRunner$ - Logic error: got unexpected type InternalSingleMulti(*int,(*int,bool))
viper.gobra.util.Violation$LogicException: Logic error: got unexpected type InternalSingleMulti(*int,(*int,bool))
        at viper.gobra.util.Violation$.violation(Violation.scala:27)
        at viper.gobra.frontend.Desugar$Desugarer.typeD(Desugar.scala:3833)
        at viper.gobra.frontend.Desugar$Desugarer.litD(Desugar.scala:2997)
        at viper.gobra.frontend.Desugar$Desugarer.exprD(Desugar.scala:2723)
        at viper.gobra.frontend.Desugar$Desugarer.exprAndTypeAsExpr(Desugar.scala:2956)
        ...
```

This can be worked around by first assigning the result of `inputMap[0]` to a variable using `let`, and then comparing on that variable.

(This has no urgency for me.)

PR #1087, opened 2026-08-10
Fix crashes on comma-ok lookups and on comparing map lookups to nil (#893)

jcp19:
Fixes both crashes reported in #893, plus two closely related crashes on the "comma-ok" form of a lookup that surfaced while fixing them.

## 1. `let v, ok := m[k] in ...` crashed with `key not found: ok_V1`

A map lookup is typed as `InternalSingleMulti(elem, (elem, bool))`, so the type checker accepts the "comma-ok" form of a short variable declaration inside a `let`, and gives `ok` type `bool`. The desugarer, however, only implemented the single-assignment form: both `PLet` cases zipped `ass.left` with `ass.right`, which silently truncates `[v, ok]` against `[m[0]]`. `ok` was therefore never bound, `varD` fell through to `localVarContextFreeD`, and the free variable `ok_V1` ended up in the Viper expression, where the backend failed to look it up.

`let v, ok := m[k] in ...` now binds `ok` to `k in m` and `v` to the value stored in the map, or to the zero value of the map's value type if the key is not contained in the map. Mathematical maps are handled as well: since a lookup in a mathematical map is only well-defined for keys in its domain, `v` is bound to `k in domain(d) ? d[k] : dflt`, which gives the same behavior as for maps.

The remaining unsupported assignments of a `let` — binding the results of a multi-valued expression, e.g. `let v, ok := x.(T) in ...` — crashed the same way. They are now reported as type errors instead.

## 2. `m[k] != nil` crashed with `got unexpected type InternalSingleMulti(...)`

`nilType` derives the type of an untyped `nil` from the other operand of the comparison and used its raw type. For a map lookup that is `InternalSingleMulti`, which `litD` passed on to `typeD`, hitting its catch-all violation. The trigger is the `nil` literal rather than the pointer element type of the map, so the crash also occurred for maps of slices, interfaces, funcs, ..., and for the other expressions typed as `InternalSingleMulti`: type assertions and receives.

`nilType` now projects the type of the compared operand to its single-valued type, which matches the workaround of first binding the lookup to a `let` variable.

## 3. `v, ok := d[k]` and `v, ok := m[k]` on a declared type crashed the desugarer

Not reported in #893, but the same family of bugs, in the statement counterpart of 1.: `multiassD` only handled lookups whose base has type `in.MapT`, so everything else fell through to `Violation.violation("Multi assignment of ... is not supported")`.

- Lookups in mathematical maps are now supported, with the same semantics as the corresponding `let` expression.
- The case for maps now matches on the underlying type of the base expression, which fixes the crash for lookups in a declared map type such as `type intMap map[int]int`. The encoding already unfolds declared types, so no change was needed there.

Both cases compute the results into temporaries before assigning them, since a target may occur in the lookup, e.g. `k, ok := m[k]`.

## Testing

- `src/test/resources/regressions/issues/000893.gobra` covers the "comma-ok" form of map and `dict` lookups in pure functions and in assertions (including lookups of keys that are not contained in the map, and of a key of a nil map), and comparisons of map lookups against `nil` for maps of pointers, maps of slices, and dictionaries.
- `src/test/resources/regressions/features/let/let_fail2.gobra` covers the assignments of a `let` that are rejected by the type checker.
- `features/maps/math-maps-simple1.gobra` and `features/maps/maps-simple1.gobra` cover the statement form for mathematical maps and for declared map types.
- `sbt "testOnly viper.gobra.GobraTests"`: 1184 passed; the only 2 failures are `features/backends/carbon.gobra` and `features/backends/vs-with-carbon.gobra`, which fail because Boogie is not installed in the environment the tests were run in.

### DIFF
--- a/src/test/resources/regressions/features/let/let_fail2.gobra
+++ b/src/test/resources/regressions/features/let/let_fail2.gobra
@@ -1,0 +1,19 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package pkg
+
+ghost
+pure func f1(x interface{}) bool {
+	// the "comma-ok" form is only supported for map lookups
+	//:: ExpectedOutput(type_error)
+	return let v, ok := x.(int) in ok
+}
+
+ghost
+requires acc(m, _)
+pure func f2(m map[int]int) int {
+	// the "comma-ok" form of a map lookup binds exactly two variables
+	//:: ExpectedOutput(type_error)
+	return let x, y, z := m[0] in x
+}
--- a/src/test/resources/regressions/features/let/let_fail2.gobra
+++ b/src/test/resources/regressions/features/let/let_fail2.gobra
@@ -5,7 +5,7 @@ package pkg
 
 ghost
 pure func f1(x interface{}) bool {
-	// the "comma-ok" form is only supported for map lookups
+	// the "comma-ok" form is only supported for map lookups for now
 	//:: ExpectedOutput(type_error)
 	return let v, ok := x.(int) in ok
 }
--- a/src/main/scala/viper/gobra/frontend/Desugar.scala
+++ b/src/main/scala/viper/gobra/frontend/Desugar.scala
@@ -2103,13 +2103,30 @@ object Desugar extends LazyLogging {
             )
           )(src)
 
-        case l@ in.IndexedExp(base, _, _) if base.typ.isInstanceOf[in.MapT] && lefts.size == 2 =>
+        case l@ in.IndexedExp(_, _, _: in.MapT) if lefts.size == 2 =>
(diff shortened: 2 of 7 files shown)


src/test/resources/regressions/features/let/let_fail2.gobra
```diff
@@ -0,0 +1,19 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package pkg
+
+ghost
+pure func f1(x interface{}) bool {
+	// the "comma-ok" form is only supported for map lookups
```
jcp19:
```suggestion
	// the "comma-ok" form is only supported for map lookups for now
```

ArquintL (approved):
LGTM

PR #1088, opened 2026-08-10
Add support for converting slices to sequences

jcp19:
Finally, we can convert directly from slices and shared arrays to sequences without having to introduce the same `toSeq` function over and over again.


src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -875,14 +875,19 @@ case class SequenceTake(left : Expr, right : Expr)(val info: Source.Parser.Info)
 /**
   * Represents the conversion of a collection of type 't',
   * represented by `expr`, to a (mathematical) sequence of type 't'.
-  * Here `expr` is assumed to be either a sequence or an exclusive array.
+  * Here `expr` is assumed to be a sequence, an array, a slice, or an option.
+  * Notice that converting a slice or a shared array requires (read) access
+  * to the elements of the converted collection.
```
jcp19:
```suggestion
```

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -875,14 +875,19 @@ case class SequenceTake(left : Expr, right : Expr)(val info: Source.Parser.Info)
...
   */
 case class SequenceConversion(expr: Expr)(val info: Source.Parser.Info) extends Expr {
   override val typ : Type = expr.typ match {
     case t: SequenceT => t
-    case t: ArrayT => t.sequence
+    // the elements of a shared array are shared as well, whereas the elements of the resulting
+    // sequence are always exclusive; `withAddressability` performs that adjustment.
+    case t: ArrayT => t.sequence.withAddressability(Addressability.conversionResult)
+    case t: SliceT => t.sequence
```
jcp19, line 884: shouldn't the code for the SliceT case be the same as the one for the ArrayT case?

src/main/scala/viper/gobra/translator/encodings/arrays/ArrayEncoding.scala
```diff
@@ -196,8 +197,14 @@ class ArrayEncoding extends TypeEncoding with SharedArrayEmbedding {
...
 
-    case n@ in.SequenceConversion(e :: ctx.Array(len, t) / Exclusive) =>
-      ctx.expression(e).map(vE => ex.toSeq(vE, cptParam(len, t)(ctx))(n)(ctx))
+    case n@ in.SequenceConversion(e :: ctx.Array(len, t) / m) =>
+      // for a shared array, `ctx.expression` already yields the corresponding exclusive array
+      // (which requires read access to all of its elements), whose elements are exclusive as well.
+      val elemT = m match {
+        case Exclusive => t
+        case Shared => t.withAddressability(Addressability.arrayElement(Exclusive))
+      }
```
jcp19: why not make this `t.withAddressability(Addressability.arrayElement(Exclusive))` without the case analsysis?

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
```diff
@@ -621,6 +637,70 @@ class SliceEncoding(arrayEmb : SharedArrayEmbedding) extends LeafTypeEncoding {
...
+    * (mathematical) sequence of the values that are stored in the first `slen(s)` cells
+    * of the slice. This function is used in the Gobra translation of the conversion
+    * "seq(s)", where the base "s" is a slice. Note that read access to the cells of the
+    * slice is required, as the result of the conversion depends on the heap.
+    *
+    * {{{
+    * function ssliceToSeq(s : Slice[T]) : Seq[T]
+    *   requires forall i : Int :: { sloc(s, i) } 0 <= i && i < slen(s) ==> Footprint[ s[i], wildcard ]
+    *   ensures |result| == slen(s)
+    *   ensures forall i : Int :: { result[i] } { sloc(s, i) } 0 <= i && i < slen(s) ==> [ result[i] == s[i] ]
```
jcp19: I wonder if it should be the encoding of `===` instead of `==`, as we may run into problems with slices of floats where `==` between NaN always yields false

src/test/resources/regressions/features/slices/slice-to-seq-simple1.gobra
```diff
@@ -0,0 +1,65 @@
...
+  assert len(seq(s)) == len(s)
+}
+
+requires forall i int :: 0 <= i && i < len(xs) ==> acc(&xs[i])
+func test9(xs ...int) {
+  assert len(seq(xs)) == len(xs)
+}
+
+ghost
+requires acc(s, _)
```
jcp19: still has a wildcard here; delete permission amounts in all pure funcs preconditions

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
jcp19: I believe this is already the `===` encoding, so no change should be needed here.

In `MemoryEncoding`, `in.EqCmp` (Go `==`) dispatches to `ctx.goEqual`, while `in.GhostEqCmp` (`===`) dispatches to `ctx.equal`:

```scala
case x@ in.EqCmp(l, r)      => ctx.goEqual(l, r)(x)
case x@ in.GhostEqCmp(l, r) => ctx.equal(l, r)(x)
```

`TypeEncoding.goEqual` merely defaults to `equal`, precisely so an encoding can override only the Go-semantics side (as `StructEncoding` and `InterfaceEncoding` do). Since this postcondition calls `ctx.equal`, it already carries `===` semantics, so a float slice holding a `NaN` would not make us inhale a false postcondition.

This also matches the surrounding precedent: `ArrayEncoding.conversionFunc` (shared → exclusive array) uses `equal` for the same purpose, and `BuiltInEncoding` builds `copy`'s element-wise postconditions with `in.GhostEqCmp`.

Worth noting there is currently no `NaN` handling anywhere in the encodings (`goEqual` for floats is still just `equal`), so the two coincide today regardless — but `ctx.equal` is the one that stays correct once float `==` gets its own semantics.

---

src/main/scala/viper/gobra/ast/internal/Program.scala
jcp19: Removed. I dropped the whole two-line sentence ("Notice that converting a slice or a shared array requires (read) access to the elements of the converted collection.") rather than only the anchored line, since deleting just that line would have left the sentence dangling. Happy to restore part of it if you meant something narrower.

---

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -1432,11 +1433,15 @@ case object SortT extends PrettyType("sort") {
   * The type of `length`-sized arrays of elements of type `typ`.
   */
 case class ArrayT(length: BigInt, elems: Type, addressability: Addressability) extends PrettyType(s"[$length]$elems") {
-  /** (Deeply) converts the current type to a `SequenceT`. */
+  /**
+    * (Deeply) converts the current type to a `SequenceT`.
+    * Notice that the elements of a shared array are shared as well, whereas the elements
+    * of a mathematical sequence are always exclusive.
```
ArquintL: How is this comment related to the implementation?

src/main/scala/viper/gobra/ast/internal/Program.scala
```diff
@@ -1451,6 +1456,14 @@ case class ArrayT(length: BigInt, elems: Type, addressability: Addressability) e
   * The (composite) type of slices of type `elems`.
   */
 case class SliceT(elems : Type, addressability: Addressability) extends PrettyType(s"[]$elems") {
+  /**
+    * Converts the current type to the `SequenceT` of the values stored in the slice.
+    * Notice that the elements of a slice are always shared, whereas the elements
+    * of a mathematical sequence are always exclusive.
```
ArquintL: How is this comment related to the implementation?
Also, what happens if we have, e.g., a sequence of slices? In this case, we would only do a shallow to sequence conversion, right?

src/main/scala/viper/gobra/translator/encodings/slices/SliceEncoding.scala
ArquintL: Then please adapt the comment to use `===` and also add a comment to line 682 saying that we use ghost equality

ArquintL (approved):
LGTM except for a few minor comments

PR #1089, opened 2026-08-11
Adds an API to cancel a running verification

ArquintL:
`GoVerifier.verifyCancellable` returns a handle with which the caller can cancel the running verification:

- Gobra's compilation stages (parsing, type-checking, desugaring, internal transformations, and encoding) check an abort signal at every stage boundary and abort cooperatively (the JVM does not support preempting the stages themselves, so cancellation latency is bounded by one stage).
- A running backend verification is interrupted via ViperServer's per-job stop (`ViperServerService.stopVerification`), for which the previously discarded `VerJobId` is retained in the `ViperServer` backend wrapper. If the verification is cancelled before the backend job is submitted, the submission is prevented (the abort remains "pending" until it can be executed).
- A cancelled verification completes with the new distinguished result `VerifierResult.Aborted` — also when the interruption of the backend's job surfaces as a failed future — i.e., callers never observe interrupt fallout.
- The abort signal lives in the (per-verification) `Config`; the existing `verify` API is unchanged, so the CLI and tests are unaffected.

The motivation is a "stop verification" button in Gobra-IDE (a corresponding branch exists in the gobra-ide repository and will be submitted once this PR lands): the IDE implements stopping via LSP's native request cancellation, which forwards to `VerificationHandle.cancel()`.

Verified: `sbt compile` and the Gobra-IDE test suite plus a wire-level end-to-end test against a server embedding this branch (verify → cancel → `RequestCancelled`, backend job interrupted, subsequent verifications unaffected — no server restart involved).


src/main/scala/viper/gobra/util/AbortSignal.scala
```diff
@@ -0,0 +1,61 @@
...
+    while (listener != null) {
+      try {
+        listener()
+      } catch {
+        // a throwing listener must neither escape into the (unrelated) caller that happens to
+        // drain it nor prevent the remaining listeners from being invoked:
+        case e: Exception => e.printStackTrace()
+      }
+      listener = listeners.poll()
+    }
```
jcp19, line 58: is there a `foreach` method we could have used here?

src/main/scala/viper/gobra/util/AbortSignal.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  /** invokes `listener` as soon as this signal is aborted (immediately if it already is) */
+  def onAbort(listener: () => Unit): Unit = {
+    listeners.add(listener)
+    if (isAborted) {
+      drainListeners()
+    }
+  }
+
+  private def drainListeners(): Unit = {
```
jcp19: `drainListeners` is a weird terminology for me. Is it standard? I would probably rename it to `invokeListeners` or sth like that

src/main/scala/viper/gobra/frontend/Config.scala
```diff
@@ -191,6 +192,8 @@ case class Config(
                    includeDirs: Vector[Path] = ConfigDefaults.DefaultIncludeDirs.map(_.toPath).toVector,
                    projectRoot: Path = ConfigDefaults.DefaultProjectRoot.toPath,
                    reporter: GobraReporter = ConfigDefaults.DefaultReporter,
+                   // signal with which a running verification can be cooperatively aborted:
+                   abortSignal: AbortSignal = new AbortSignal(),
```
jcp19: This object only makes sense when runing gobra through the GobraServer, right? Right now, we need to have a signal in every config object. Could we move it to a more appropriate place?

jcp19: Maybe it does not need to exist in the Config object, maybe only in the `ViperServer` object or we can create a new one on every call to `verifyCancelable`

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -172,18 +194,37 @@ class Gobra extends GoVerifier with GoIdeVerifier {
...
+      // particular, backend failures caused by interrupting the backend's job are mapped, too):
+      case _ if config.abortSignal.isAborted => VerifierResult.Aborted
+      case _: AbortedException => VerifierResult.Aborted
+    }
+  }
+
+  /** fails the verification with an [[AbortedException]] if the verification has been aborted */
+  private def abortCheckpoint(config: Config)(implicit executor: GobraExecutionContext): EitherT[Vector[VerifierError], Future, Unit] = {
+    if (config.abortSignal.isAborted) {
+      EitherT.fromEither(Future.failed[Either[Vector[VerifierError], Unit]](new AbortedException()))
```
jcp19: this is a complicated return type where we never care about the value in the Right case. I don't see a better alternative right now, it would be cool if we could avoid having to introduce exceptions for the sake ok exiting the monad early, but all alternatives seem equally clumsy

src/main/scala/viper/gobra/util/AbortSignal.scala
ArquintL: I change it to `notifyAndDeregisterListeners` to emphasize that this removes the invoked listeners from the queue

src/main/scala/viper/gobra/util/AbortSignal.scala
ArquintL: I just checked the API and there does not seem to be a more elegant way of iterating & removing elements in a thread-safe way

src/main/scala/viper/gobra/Gobra.scala
ArquintL: I just introduced an "intermediate verifier result" such that we cleanly distinguish between success, error, skipped, and abort

src/main/scala/viper/gobra/frontend/Config.scala
ArquintL: Are you worried about runtime overhead, i.e., would making it an `Option` be a solution? Imho, putting it into `Config` seems quite elegant to avoid the extra parameter left and right

jcp19: No, it was more about people reading the code. Adding this field may lead people to believe that this field is meaningful for every configuration / run of Gobra when that is seldom the case. At any rate, I understand the point of keeping it here. Could we create it only for the configs for the ViperServer / IDE?

jcp19: I do not insist on this point, if you think that is too cumbersome, I am happy to keep it like this

ArquintL: It would be easy to make it an `Option`. Not sure what other alternatives would be... introducing an IDE specific Config subclass?

jcp19: yeah, I think neither of these two options is particularly appealing. It's fine to keep it like this

ArquintL: Okay. I've added an explicit comment to the Config field to point out that this for Gobra-IDE

src/main/scala/viper/gobra/backend/BackendVerifier.scala
```diff
@@ -62,19 +65,26 @@ object BackendVerifier {
...
-          partialRes.foldLeft[VerificationResult](silver.verifier.Success) {
-            case (acc, silver.verifier.Success) => acc
-            case (silver.verifier.Success, res) => res
-            case (silver.verifier.Failure(l), silver.verifier.Failure(r)) => silver.verifier.Failure(l ++ r)
+        val fut = partialVerificationResults map { partialRes =>
+          partialRes.foldLeft[Either[LeftIntermediateVerifierResult, VerificationResult]](Right(silver.verifier.Success)) {
+            // merge logic: we accumulate (right) verification errors until we hit the first left result at which point
+            // we stop accumulating. While this models the monadic bind of Either, one could consider improving this merge
+            // logic in the future: E.g., one could accumulate Left(Errored(errs)) or define a precedence for different
+            // left results.
```
jcp19: This comment seems pretty obvious and verbose for the most part. Maybe keep the idea for future improvements, but drop the rest.

src/main/scala/viper/gobra/reporting/VerifierResult.scala
```diff
@@ -5,13 +5,57 @@
...
+  }
+
+  /** the verification has been aborted, e.g., via [[viper.gobra.util.AbortSignal]] */
+  case object Aborted extends LeftIntermediateVerifierResult {
+    override def toVerifierResult: VerifierResult = VerifierResult.Aborted
+  }
+  /** the step has been skipped due to a configuration option, which ends the pipeline successfully */
+  case object Skipped extends LeftIntermediateVerifierResult {
+    override def toVerifierResult: VerifierResult = VerifierResult.Success
+  }
```
jcp19: Do I see it correctly that if we extend `VerifierResult` with an extra constructor (Skipped), then we could drop the `LeftIntermediateVerifierResult` type entirely? That would probably be preferrable, bc then we could produce better error messages (e.g., rather than saying we got 0 errors bc we convert a skip to a success, we could write messages like "Verification skipped" -- this would have helped me a few times to realize that I passed --noVerify by mistake from copying and pasting commands).

jcp19 (approved):
LGTM. However, please consider dropping the type `LeftIntermediateVerifierResult` and extending `VerifierResult` instead, I think that leads to much clearer code and better messages.

src/main/scala/viper/gobra/reporting/VerifierResult.scala
ArquintL: Sounds good, I did exactly this but kept another trait `NegativeVerifierResult` to distinguish `Success` from all other such that we get type guarantees that it's impossible to have `Left(Success)`

PR #1092, opened 2026-08-13
[Experiment] Enable Go's name visibility rules

jcp19:
This PR introduces Go's visibility rules in Gobra, in a way described in file `docs/design/access-modifiers.md` (this design doc was written by me, not by an LLM). There, I describe the basic rules (besides the rules from Go, I also require that the specifications of public members reference only public symbols, as private members are not known to clients of the package and thus, clients of public members would not be able to interpret the specs if they contained private symbols).

In addition, I introduce the concept of closed pure functions and predicates, i.e., pure functions and predicates may never be known to clients.

The idea is that enforcing these restrictions would allow us to reduce the context passed to the desugarer and encoding, severely reducing the size and proof context of the generated files (ideally, we would be able to omit (1) the definitions of imported private functions, predicates, and methods, (2) the bodies of imported closed predicates and pure functions. Unfortunately, this PR stops short of that: it only introduces the type checking restrictions, plus one change to the desugarer: closed pure functions and methods of imported packages are encoded as opaque functions, which importing packages may never reveal. Closed predicates are encoded as usual; what hides their bodies from clients is the type checker, which rejects folding and unfolding an imported closed predicate.

The reason for this is that removing bodies of functions may lead to unsoundenesses with termination-checking due to the non-modular nature of our checks. I will describe these issues in a separate issue.
Despite the changes to the desugarer and encoding not being implemented yet, I think it is worthwile merging the changes to the type-system, as this leads to generally better specifications and more accurate language support.

@ArquintL would this PR break a lot of stuff on your side?


src/main/resources/stubs/strconv/itoa.gobra
ArquintL, line 107:
```suggestion
const digits = "0123456789abcdefghijklmnopqrstuvwxyz"
const DigitsLen = 36

// since contracts cannot reference `digits` and its length, we introduce the (public)
// constant `DigitsLen` and make sure that it correctly reflects `digits`'s length:
ghost
decreases
func digitsProofObligation() {
    assert len(digits) == DigitsLen
}
```

src/main/resources/stubs/strconv/itoa.gobra
```diff
@@ -46,7 +46,7 @@ func Itoa(i int) string {
 
 // AppendInt appends the string form of the integer i,
 // as generated by FormatInt, to dst and returns the extended buffer.
-requires 2 <= base && base <= len(digits)
+requires 2 <= base && base <= 36 // 36 == len(digits), which cannot be referenced in the contract of an exported member
```
ArquintL, line 49:
```suggestion
requires 2 <= base && base <= DigitsLen
```

src/main/resources/stubs/strconv/itoa.gobra
```diff
@@ -62,7 +62,7 @@ func AppendInt(dst []byte, i int64, base int) (res []byte) {
 
 // AppendUint appends the string form of the unsigned integer i,
 // as generated by FormatUint, to dst and returns the extended buffer.
-requires 2 <= base && base <= len(digits)
+requires 2 <= base && base <= 36 // 36 == len(digits), which cannot be referenced in the contract of an exported member
```
ArquintL, line 65:
```suggestion
requires 2 <= base && base <= DigitsLen
```

src/main/resources/stubs/strconv/itoa.gobra
```diff
@@ -113,7 +113,7 @@ const digits = "0123456789abcdefghijklmnopqrstuvwxyz"
 // as the second result value.
 //
 requires append_ ==> forall i int :: 0 <= i && i < len(dst) ==> acc(&dst[i])
-requires 2 <= base && base <= len(digits)
+requires 2 <= base && base <= 36 // 36 == len(digits), which cannot be referenced in the contract of an exported member
```
ArquintL, line 116:
```suggestion
requires 2 <= base && base <= DigitsLen
```

src/main/resources/stubs/time/time.gobra
```diff
@@ -244,7 +247,7 @@ func (t Time) In(loc *Location) Time
 
 // Location returns the time zone information associated with t.
 // ensures t.loc == nil ==> res == UTC // UTC is currently commented because it is a global var
-ensures t.loc != nil ==> res == t.loc
+// ensures t.loc != nil ==> res == t.loc // commented because contracts of exported members cannot reference the non-exported field loc
 decreases
 func (t Time) Location() (res *Location)
```
ArquintL, line 252: That's an interesting spec example. Clearly it doesn't make sense to reference private fields but on the other hand the now commented out spec specifies determinism, which we should retain by marking the function as `pure`: cf. https://cs.opensource.google/go/go/+/refs/tags/go1.27.1:src/time/time.go;l=1393

src/main/scala/viper/gobra/ast/frontend/PrettyPrinter.scala
```diff
@@ -815,9 +817,9 @@ class ShortPrettyPrinter extends DefaultPrettyPrinter {
     }
     case member: PGhostMember => member match {
       case PExplicitGhostMember(m) => "ghost" <+> showMember(m)
-      case PFPredicateDecl(id, args, _) =>
+      case PFPredicateDecl(id, args, _, _) =>
         "pred" <+> showId(id) <> parens(showParameterList(args))
-      case PMPredicateDecl(id, recv, args, _) =>
+      case PMPredicateDecl(id, recv, args, _, _) =>
         "pred" <+> showReceiver(recv) <+> showId(id) <> parens(showParameterList(args))
```
ArquintL, line 823: I'm not entirely sure where exactly we use the ShortPrettyPrinter but I'm afraid it might be confusing if `closed` doesn't show up despite the member being `closed`

```suggestion
      case PFPredicateDecl(id, args, _, isClosed) =>
(if (isClosed) showClosed else emptyDoc) <> "pred" <+> showId(id) <> parens(showParameterList(args))
      case PMPredicateDecl(id, recv, args, _, isClosed) =>
(if (isClosed) showClosed else emptyDoc) <> "pred" <+> showReceiver(recv) <+> showId(id) <> parens(showParameterList(args))
```

src/main/scala/viper/gobra/frontend/info/implementation/property/Assignability.scala
```diff
@@ -195,6 +195,12 @@ trait Assignability extends BaseProperty { this: TypeInfoImpl =>
                 }.getOrElse(successProp)
               })
           } else if (elems.size == s.embedded.size + s.fields.size) {
+            // a positional literal assigns to all fields, including non-exported ones,
+            // which is not allowed for struct types declared in other packages
+            val hasNonExportedFields = (s.embedded.keys ++ s.fields.keys).exists(!isExportedName(_))
+            failedProp(s"cannot use a positional composite literal for type $typ declared in another package, " +
+              s"as it would assign to non-exported fields",
+              isImportedContextualType(typ) && hasNonExportedFields) and
```
ArquintL, line 203: So this basically leaks whether a type is private fields. Is this fine / handled like this by Go?

src/main/scala/viper/gobra/frontend/info/implementation/property/Visibility.scala
```diff
@@ -0,0 +1,116 @@
...
+    * (e.g., `error`) are accessible from every package despite having non-exported names.
+    */
+  def isBuiltinContext(context: ExternalTypeInfo): Boolean = context.pkgName.name == "builtin"
+
+  /**
+    * True iff member `m` with name `name` is exported. ADT discriminators are generated
+    * members named `is<Clause>`; they are considered exported iff their clause is exported.
+    */
+  def isEffectivelyExported(m: Regular, name: String): Boolean = m match {
+    case _: AdtDiscriminator => isExportedName(name.stripPrefix("is"))
```
ArquintL, line 35: Can we introduce a constant for `"is"`?

src/main/scala/viper/gobra/frontend/info/implementation/property/Visibility.scala
```diff
@@ -0,0 +1,116 @@
...
+    case _ => isExportedName(name)
+  }
+
+  /**
+    * True iff accessing member `m` under name `name` from this package violates Go's visibility
+    * rules, i.e., `m` is declared in another (non-builtin) package and `name` is not exported.
+    */
+  def isForeignPrivateAccess(m: Regular, name: String): Boolean =
+    m.context.getTypeInfo != this.getTypeInfo && !isBuiltinContext(m.context) && !isEffectivelyExported(m, name)
```
ArquintL, line 44:
```suggestion
    isImportedMember(m) && !isBuiltinContext(m.context) && !isEffectivelyExported(m, name)
```

src/main/scala/viper/gobra/frontend/info/implementation/property/Visibility.scala
```diff
@@ -0,0 +1,116 @@
...
+
+  /** True iff `m` is declared in another package than the one this type checker belongs to. */
+  def isImportedMember(m: Regular): Boolean = m.context.getTypeInfo != this.getTypeInfo
+
+  /**
+    * Replaces `e` by an error entity if accessing it under name `id` from this package
+    * violates Go's visibility rules; returns `e` unchanged otherwise.
+    */
+  def filterForeignPrivate(e: Entity, id: PIdnUse): Entity = e match {
```
ArquintL, line 53: Hmm I don't like it that much that we have to carry around an entity _and_ an identifier, despite them being tightly correlated. I'm wondering whether it wouldn't be nicer for the symbol table (i.e., each regular) to keep track of whether it is exported or not. I feel like this is a property of a symbol table entry and less so of the particular identifier that is used to access it)

src/main/scala/viper/gobra/frontend/info/implementation/resolution/MemberResolution.scala
```diff
@@ -453,6 +459,12 @@ trait MemberResolution { this: TypeInfoImpl =>
     Violation.violation(dependentTypeInfo.contains(importTarget), s"Expected that package ${tree.root.info.id} has access to the type information of package $importTarget")
     val foreignTypeChecker = dependentTypeInfo(importTarget)
     val foreignPkgResult = foreignTypeChecker.externalRegular(id)
-    foreignPkgResult.flatMap(m => Some((m, Vector())))
+    foreignPkgResult.map { m =>
+      // enforce Go's visibility rules: only exported members of imported packages may be
+      // accessed (the `builtin` package being the exception).
+      if (importTarget != BuiltInImport && !isBuiltinContext(foreignTypeChecker) && !isExportedName(id.name)) {
+        (ErrorMsgEntity(error(id, s"${id.name} is not exported by package ${foreignTypeChecker.pkgName.name}")), Vector())
+      } else (m, Vector())
```
ArquintL, line 467: Isn't this very similar to what's happening on line 349? Why does line 349 use `filterForeignPrivate` while we use a different mechanism here?

src/main/scala/viper/gobra/frontend/info/implementation/typing/ghost/GhostMiscTyping.scala
```diff
@@ -172,6 +172,28 @@ trait GhostMiscTyping extends BaseTyping { this: TypeInfoImpl =>
...
+        })
+        error(n, "Only members with exported names can be marked as closed.", closedOnNonExportedName)
+      } ++
+      {
+        // Contracts of exported members are part of the package's interface and must be
+        // interpretable by importing packages: they may only reference exported members.
+        val ownerIsClientFacing = tree.parent(n).head match {
+          case d: PFunctionDecl => isExportedName(d.id.name)
+          case d: PMethodDecl => isExportedName(d.id.name) && hasExportedReceiver(d.receiver)
+          case sig: PMethodSig => isClientFacingInterfaceMember(sig)
```
ArquintL, line 192: why do the checks here differ from the checks on lines 180f?

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -183,6 +183,14 @@ object Desugar extends LazyLogging {
 
   private class Desugarer(@unused pom: PositionManager, info: viper.gobra.frontend.info.TypeInfo, initSpecs: PackageInitSpecCollector) {
 
+    /**
+      * True iff this desugarer processes an imported package. In that case, `closed` pure functions
+      * and methods are desugared as opaque members, i.e., their bodies are encoded (they are needed
+      * to obtain a complete call graph) but hidden from the package under verification.
+      * Set by [[packageD]].
+      */
+    private var isDesugaringImportedPackage: Boolean = false
```
ArquintL, line 192: I'd make this a class parameter and then drop it from `packageD` as we use a different `Desugarer` instance per package

src/main/scala/viper/gobra/frontend/info/implementation/property/Visibility.scala
```diff
@@ -0,0 +1,116 @@
...
+    case m: Regular if isForeignPrivateAccess(m, id.name) =>
+      ErrorMsgEntity(error(id, s"${id.name} is not exported by package ${m.context.pkgName.name}"))
+    case _ => e
+  }
+
+  /** True iff the receiver's declared type has an exported name. */
+  def hasExportedReceiver(recv: PReceiver): Boolean = isExportedName(recv.typ.typ.id.name)
```
ArquintL, line 60: Doesn't this contradict the example of a public constructor returning a private type? So you cannot call public methods on such a type despite holding a value of this type?

src/main/scala/viper/gobra/frontend/info/implementation/typing/MemberTyping.scala
```diff
@@ -28,14 +28,23 @@ trait MemberTyping extends BaseTyping { this: TypeInfoImpl =>
...
     case m: PMethodDecl =>
       wellDefVariadicArgs(m.args) ++
         isReceiverType.errors(miscType(m.receiver))(member) ++
         wellDefIfPureMethod(m) ++
         wellFoundedIfNeeded(m) ++
         atomicMemberIsWellFormed(m) ++
-        noConditionalMeasureIfGhostOrPure(m)
+        noConditionalMeasureIfGhostOrPure(m) ++
+        // see the analogous check for pure functions above
+        (if (m.spec.isPure && !m.spec.isClosed && isExportedName(m.id.name) && hasExportedReceiver(m.receiver))
```
ArquintL, line 45: this is related to my question above about exported receivers: is it really necessary that the method and receiver are public? Should a capitalized method with a private receiver be rejected by the type checker or is there a use-case for this?

ArquintL (commented):
I've skipped all changes in `test/`

ArquintL:
> @ArquintL would this PR break a lot of stuff on your side?

I'm checking right now how big the changes would be for my most important project

ArquintL:
@jcp19 Claude identified the following issue with the current design:
`lib/lib.gobra`:
```
package lib

type Ctx interface {
	ghost
	decreases
	pure Props() bool
}

ghost type Searcher ghost struct {
	ctx Ctx
}

ghost
decreases
closed
pure func (s Searcher) Props() bool {
	return s.ctx != nil && s.ctx.Props()
}

ghost
decreases
requires s.Props()
ensures  res != nil && res.Props()
closed
pure func (s Searcher) Get() (res Ctx) {
	return s.ctx
}
```

`main.gobra`:
```
package main

import "lib"

ghost
decreases
requires s.Props()
func client(s lib.Searcher) {
	c := s.Get()
	assert c != nil
}
```

Commands and output:

$ java -Xss128m -jar gobra.jar -i lib/lib.gobra -I .
Gobra found 0 errors.

$ java -Xss128m -jar gobra.jar -i main.gobra -I .
Error at: <./lib/lib.gobra:23:10> Postcondition might not hold.
Assertion res != nil might not hold.
Gobra found 1 error.
What happens: lib verifies on its own because Get's postcondition follows from its precondition s.Props(), whose body says s.ctx != nil. When main imports lib, the desugarer marks both closed members opaque (hidesBodyFromClients in Desugar.scala), yet their bodies are still emitted and Viper re-verifies Get inside main's program. There, Props() is opaque, so s.ctx != nil is no longer derivable and the re-check of Get's own postcondition fails, even though the importer never looks at either body. The error location is inside the imported package, which makes it look like a library bug to the client.

Two possible fixes on the Gobra side: keep sibling calls within the same imported package non-opaque (or reveal them) when re-encoding that package's members, or skip re-verifying imported members' contracts altogether since they were already checked in their declaring package. The design doc keeps bodies for the termination-checking call graph, so the first option is likely the smaller change.

ArquintL:
It seems to be that what we would need here is `assume false` or a more semantic per-member flag to instruct Viper backends to not verify a member. So far, we always operated under the assumption that one would simply omit a body if one wants this behavior (that's what ViperServer, e.g., does for cached members) but the termination plugin's analysis is a compelling example why such a seemingly trivial code transformation has disadvantages.
(We could even consider implementing such a language extension as a Viper plugin.)

jcp19:
> It seems to be that what we would need here is `assume false` or a more semantic per-member flag to instruct Viper backends to not verify a member. So far, we always operated under the assumption that one would simply omit a body if one wants this behavior (that's what ViperServer, e.g., does for cached members) but the termination plugin's analysis is a compelling example why such a seemingly trivial code transformation has disadvantages.
> 
> (We could even consider implementing such a language extension as a Viper plugin.)

I wish that were true, and in principle, such a fix would be sound. Unfortunately, it would not be modular, in that we would need to know the bodies of imported functions (or at least, a superset of all its callees) in order to understand error messages generated by this approach (calls to two different imported functions with exactly the same contract could have different verification results, depending on their functions called in their bodies). Besides these concerns with the error messages, there are other important aspects: with such an approach, we are never able to fully erase private members from imported packages from the encoding, which is wasteful and polutes the verification context, which further increases stability and performance issues (I discuss some of these issues in my thesis). I'd rather invest in a modular termination approach like that of Bart Jacobs, or the one @marcoeilers used in Nagini, though as far as I know, that would not work yet for pure functions.

PR #1094, opened 2026-08-17
Fix nil intf receiver for mpred

ArquintL:
Experimenting with nil mpred receivers revealed a bug where Gobra reports a Violation if there are no implementations for an interface and a nil receiver of this interface is used to fold a predicate.
While Gobra supports folding a predicate using an interface receiver, the encoding branches over the interface's implementation to perform the fold according to the dynamic type. However, if there are no implementations, Gobra uses an abstract predicate instead, whose fold operation fails. This PR changes Gobra to emit empty predicate families for interfaces without any implementations. Thus, the interface predicate will be non-abstract, branch over the empty set of implementations, and, thus, land in the unknown case which results in a fold error.

### DIFF
--- a/src/main/scala/viper/gobra/translator/encodings/interfaces/InterfaceEncoding.scala
+++ b/src/main/scala/viper/gobra/translator/encodings/interfaces/InterfaceEncoding.scala
@@ -621,12 +621,13 @@ class InterfaceEncoding extends LeafTypeEncoding {
       // which, e.g., cannot be folded without rendering the Viper program inconsistent:
       val implementedItfs = implementedItfNodes.map(_._2.withAddressability(Addressability.Exclusive))
       val unimplementedItfNodes = for {
-        m <- ctx.table.getMPredicates.toSet[in.MPredicateLikeMember].collect{ case m: in.MPredicate => m }
-        itf <- m.receiver.typ match {
-          case itf: in.InterfaceT if !implementedItfs.contains(itf.withAddressability(Addressability.Exclusive)) =>
-            Set(itf.withAddressability(Addressability.Exclusive))
-          case _ => Set.empty[in.InterfaceT]
-        }
+        (m, itf) <- ctx.table.getMPredicates.toSet[in.MPredicateLikeMember]
+          .collect{ case m: in.MPredicate => m }
+          // restrict the predicates to those that have an unimplemented interface receiver:
+          .collect{ m => m.receiver.typ match {
+            case itf: in.InterfaceT if !implementedItfs.contains(itf.withAddressability(Addressability.Exclusive)) =>
+              (m, itf.withAddressability(Addressability.Exclusive))
+          }}
       } yield (m.name, itf, SortedSet.empty[in.Type])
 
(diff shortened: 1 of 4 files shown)


src/main/scala/viper/gobra/translator/encodings/interfaces/InterfaceEncoding.scala
```diff
@@ -610,11 +610,27 @@ class InterfaceEncoding extends LeafTypeEncoding {
...
+      // which, e.g., cannot be folded without rendering the Viper program inconsistent:
+      val implementedItfs = implementedItfNodes.map(_._2.withAddressability(Addressability.Exclusive))
+      val unimplementedItfNodes = for {
+        m <- ctx.table.getMPredicates.toSet[in.MPredicateLikeMember].collect{ case m: in.MPredicate => m }
+        itf <- m.receiver.typ match {
+          case itf: in.InterfaceT if !implementedItfs.contains(itf.withAddressability(Addressability.Exclusive)) =>
+            Set(itf.withAddressability(Addressability.Exclusive))
+          case _ => Set.empty[in.InterfaceT]
+        }
+      } yield (m.name, itf, SortedSet.empty[in.Type])
```
jcp19, line 629: Do we always need to add this triple even when itf is computed in the second case? Couldn't we use a call to `collect` instead and ignore the second case of the match entirely?

jcp19 (approved):
LGTM, just one question

src/main/scala/viper/gobra/translator/encodings/interfaces/InterfaceEncoding.scala
ArquintL: good point!

src/main/scala/viper/gobra/translator/encodings/interfaces/InterfaceEncoding.scala
```diff
@@ -610,11 +610,28 @@ class InterfaceEncoding extends LeafTypeEncoding {
...
+      // predicates of interfaces without any known implementation form a singleton family such
+      // that their instances are also encoded via the (dispatching) family predicate. Otherwise,
+      // such instances would refer to the abstract predicate emitted for the interface member,
+      // which, e.g., cannot be folded without rendering the Viper program inconsistent:
+      val implementedItfs = implementedItfNodes.map(_._2.withAddressability(Addressability.Exclusive))
+      val unimplementedItfNodes = for {
+        (m, itf) <- ctx.table.getMPredicates.toSet[in.MPredicateLikeMember]
+          .collect{ case m: in.MPredicate => m }
+          // restrict the predicates to those that have an unimplemented interface receiver:
+          .collect{ m => m.receiver.typ match {
```
jcp19:
```suggestion
          // restrict the predicates to those that have an unimplemented interface receiver:
          .collect{ m: in.MPredicate => m.receiver.typ match {
```

jcp19: I think this change is worth doing too

PR #1095, opened 2026-08-17
Checks non-nilness when indexing a pointer to an array

ArquintL:
Indexing a pointer to an array implicitly dereferences the pointer ('p[i]' is shorthand for '(*p)[i]'), but the desugarer does not introduce an explicit dereference node for it. Consequently, the structural search for the outermost dereference did not find any dereference and usages such as '&p[0]' generated no nil-ness proof obligation at all, unlike the analogous field accesses.

The added test cases document the different behavior of []byte and *[N]byte, including that permission to the array does not (yet) entail non-nilness of the pointer (cf. `addrOfArrayElemWithPermission`), in contrast to structs, since the shared-array footprint consists of quantified permissions over embedded locations rather than field permissions on the dereferenced pointer.

### DIFF
--- a/src/test/resources/regressions/issues/000491-4.gobra
+++ b/src/test/resources/regressions/issues/000491-4.gobra
@@ -147,4 +147,16 @@ func addrOfReslicedElemOutOfBounds(s []byte) (r *byte) {
 	return
 }
 
+// Triggers are patterns that are never evaluated and, thus, must not generate
+// panic-absence checks, also in expression positions within actual code.
+requires forall i int :: { &s[i] } 0 <= i && i < len(s) ==> acc(&s[i])
+func quantifierWithAddrTriggerInCode(s []byte) {
+	b := forall i int :: { &s[i] } 0 <= i && i < len(s) ==> s[i] >= 0
+	_ = b
+}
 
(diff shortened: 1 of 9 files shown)


src/test/resources/regressions/issues/000491-4.gobra
```diff
@@ -46,3 +46,143 @@ func addrOfFieldOfNil(p *Pair) (r *int) {
...
+// Unlike for structs, permission to the array does not (yet) entail that the pointer is
+// non-nil: the shared-array footprint consists of quantified permissions over embedded
+// locations rather than field permissions on the dereferenced pointer itself, so Viper's
+// permission-implies-non-null reasoning does not apply.
+requires acc(p)
+func addrOfArrayElemWithPermission(p *[8]byte) (r *byte) {
+	//:: ExpectedOutput(load_error:receiver_is_nil_error)
+	r = &p[0]
+	return
+}
```
jcp19, line 99: I don't understand this. What part of the encoding for array pointers would be to blame for this incompletenes?

jcp19:
(This was merged automatically without my approval, we can continue the discussion about this PR in #1077)

PR #1099, opened 2026-08-18
Improve Stats Writing in the Presence of Crashes

ArquintL:
I'm trying to figure out while Gobra crashes in CI for one of my projects due to failed prover interactions, which could be related to OOM situations. While Gobra would provide many insights as part of the produced stats file, this file is not written as the exception bypasses proper shutdown (and skips the verification of all other packages).
This PR addresses this issue by catching non-fatal exceptions while verifying packages and writing the stats file after each package to have at least some data even in the case of fatal exceptions.

### DIFF
--- a/src/main/scala/viper/gobra/Gobra.scala
+++ b/src/main/scala/viper/gobra/Gobra.scala
@@ -67,7 +67,7 @@ trait GoVerifier extends StrictLogging {
     var warningCount: Int = 0
     var allVerifierErrors: Vector[VerifierError] = Vector()
     var allTimeoutErrors: Vector[TimeoutError] = Vector()
-    var abortedPackages: Vector[String] = Vector()
+    var allExceptionErrors: Vector[ExceptionError] = Vector()
     val isVerifyingMultiplePackages = config.packageInfoInputMap.size != 1
 
     def writeStatsReport(): Unit = config.gobraDirectory match {
@@ -135,7 +135,7 @@ trait GoVerifier extends StrictLogging {
           // abort the verification of the remaining packages.
           logger.error(s"The verification of package $pkgId was aborted by an exception.", e)
           statsCollector.report(VerificationTaskFinishedMessage(pkgId))
-          abortedPackages = abortedPackages :+ pkgId
+          allExceptionErrors = allExceptionErrors :+ ExceptionError(pkgId, e)
       }
       // Keep the on-disk report current after every package: a later kill
       // that bypasses shutdown hooks (e.g. SIGKILL) then loses at most the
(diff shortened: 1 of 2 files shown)


src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -125,7 +129,18 @@ trait GoVerifier extends StrictLogging {
...
+          // ProverInteractionFailed when a prover process dies) must not
+          // abort the verification of the remaining packages.
+          logger.error(s"The verification of package $pkgId was aborted by an exception.", e)
+          statsCollector.report(VerificationTaskFinishedMessage(pkgId))
+          abortedPackages = abortedPackages :+ pkgId
       }
+      // Keep the on-disk report current after every package: a later kill
+      // that bypasses shutdown hooks (e.g. SIGKILL) then loses at most the
+      // statistics of the package that was in flight.
+      writeStatsReport()
```
jcp19, line 143: do I get it right that the report might be written twice in one exectution? This is inteded, right?

src/main/scala/viper/gobra/Gobra.scala
```diff
@@ -158,9 +173,14 @@ trait GoVerifier extends StrictLogging {
...
       logger.info(s"The verification of ${allTimeoutErrors.size} member${addPlural(allTimeoutErrors.size)} timed out.")
     }
+    if (abortedPackages.nonEmpty) {
+      logger.error(s"The verification of ${abortedPackages.size} package${addPlural(abortedPackages.size)} was aborted by an exception: ${abortedPackages.mkString(", ")}")
+    }
 
     val allErrors = allVerifierErrors ++ allTimeoutErrors
-    if (allErrors.isEmpty) VerifierResult.Success else VerifierResult.Failure(allErrors)
+    if (allErrors.nonEmpty) VerifierResult.Failure(allErrors)
+    else if (abortedPackages.nonEmpty) VerifierResult.Aborted
```
jcp19: I think we should distinguish between packages that are aborted because the service that started the job requested it, and packages that are aborted due to exceptions in the tools. The former should have `Aborted` as the verification result, whereas the later should be a Failure imo (with a matching error condition). In particular, my expectation when running Gobra in the CI is that, if a package is aborted by an exception, then the CI run fails.

src/main/scala/viper/gobra/Gobra.scala
ArquintL: Yes. It's once per verified package _and_ once at the very end, so at least twice

src/main/scala/viper/gobra/Gobra.scala
ArquintL: Sounds good, then let's add another result, Exception to make it very different from the usual Failure and Abort

src/main/scala/viper/gobra/Gobra.scala
jcp19: Do the stats of each package get overwriten when the next package is being verified??

src/main/scala/viper/gobra/Gobra.scala
jcp19: Sure, either that or Failure are fine by me, the error messages would make it clear which packages had the exception and were unintetionally aborted

src/main/scala/viper/gobra/Gobra.scala
ArquintL: Yes the file get's overwritten but the stats are grouped by package. Thus, the final file will contain all verified packages

jcp19 (approved):

PR #1100, opened 2026-08-27
Provide optional `perm` parameter to range loops over maps

jcp19:
To prevent mutating the map while it is being traversed by a range loop, Gobra exhales a small but fixed permission to the map before iteration. This prevents us from traversing maps in cases where we only have `_` or an amount which is only known to be positive.

This PR extends the syntax for `range` loops over maps, where we can pass an extra optional parameter with a permission amount or `_`, which stands for the permission amount we exhale when traversing the map (instead of using the default)

```
// if p is passed here, gobra exhales acc(m, p), instead of the default perm amount
for k, v := range m, p {

}

// _ is also allowed in the perm amount
for k, v := range m, _ {

}
```

### DIFF
--- a/src/main/scala/viper/gobra/ast/frontend/Ast.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/Ast.scala
@@ -846,6 +846,7 @@ sealed trait PActualMisc extends PMisc
   * Range clause of a for statement, i.e. `range exp(, perm)?( with enumerated)?`.
   * `perm` is the (optional) amount of permission to `exp` that is exhaled while iterating over it.
   * It is only supported for maps, where it defaults to a very small, but positive, amount.
+  * As in an `acc` expression, the blank identifier denotes a wildcard amount.
   */
 case class PRange(exp: PExpression, perm: Option[PRangePerm], enumerated: PUnkLikeId) extends PActualMisc
 
@@ -1367,7 +1368,8 @@ case class PTrigger(exps: Vector[PExpression]) extends PGhostMisc
 /**
   * The permission amount of a range clause, i.e. the `p` in `range exp, p`. It is a
   * specification-only annotation, and thus ghost, even though the enclosing for statement
-  * and the range expression itself need not be.
+  * and the range expression itself need not be. A `PWildcardPerm` amount stands for the
+  * `_` written by the user.
   */
 case class PRangePerm(exp: PExpression) extends PGhostMisc
 
--- a/src/main/scala/viper/gobra/ast/frontend/Ast.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/Ast.scala
@@ -842,7 +842,12 @@ sealed trait PMisc extends PNode
 
 sealed trait PActualMisc extends PMisc
 
-case class PRange(exp: PExpression, enumerated: PUnkLikeId) extends PActualMisc
+/**
+  * Range clause of a for statement, i.e. `range exp(, perm)?( with enumerated)?`.
+  * `perm` is the (optional) amount of permission to `exp` that is exhaled while iterating over it.
+  * It is only supported for maps, where it defaults to a very small, but positive, amount.
+  */
+case class PRange(exp: PExpression, perm: Option[PExpression], enumerated: PUnkLikeId) extends PActualMisc
 
 sealed trait PParameter extends PMisc {
   def typ: PType
--- a/src/main/scala/viper/gobra/ast/frontend/Ast.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/Ast.scala
@@ -847,7 +847,7 @@ sealed trait PActualMisc extends PMisc
   * `perm` is the (optional) amount of permission to `exp` that is exhaled while iterating over it.
   * It is only supported for maps, where it defaults to a very small, but positive, amount.
   */
-case class PRange(exp: PExpression, perm: Option[PExpression], enumerated: PUnkLikeId) extends PActualMisc
+case class PRange(exp: PExpression, perm: Option[PRangePerm], enumerated: PUnkLikeId) extends PActualMisc
 
 sealed trait PParameter extends PMisc {
   def typ: PType
@@ -1363,6 +1363,13 @@ sealed trait PGhostMisc extends PMisc with PGhostNode
 case class PBoundVariable(id: PIdnDef, typ: PType) extends PGhostMisc
 
 case class PTrigger(exps: Vector[PExpression]) extends PGhostMisc
+
+/**
+  * The permission amount of a range clause, i.e. the `p` in `range exp, p`. It is a
+  * specification-only annotation, and thus ghost, even though the enclosing for statement
+  * and the range expression itself need not be.
+  */
+case class PRangePerm(exp: PExpression) extends PGhostMisc
 
 case class PExplicitGhostParameter(actual: PActualParameter) extends PParameter with PGhostMisc with PGhostifier[PActualParameter] {
   override def typ: PType = actual.typ
--- a/src/main/scala/viper/gobra/ast/frontend/Ast.scala
+++ b/src/main/scala/viper/gobra/ast/frontend/Ast.scala
@@ -848,7 +848,7 @@ sealed trait PActualMisc extends PMisc
   * It is only supported for maps, where it defaults to a very small, but positive, amount.
   * As in an `acc` expression, the blank identifier denotes a wildcard amount.
   */
-case class PRange(exp: PExpression, perm: Option[PRangePerm], enumerated: PUnkLikeId) extends PActualMisc
+case class PRange(exp: PExpression, perm: Option[PExpression], enumerated: PUnkLikeId) extends PActualMisc
 
 sealed trait PParameter extends PMisc {
   def typ: PType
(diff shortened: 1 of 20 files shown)


src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -1359,6 +1365,14 @@ case class PBoundVariable(id: PIdnDef, typ: PType) extends PGhostMisc
 
 case class PTrigger(exps: Vector[PExpression]) extends PGhostMisc
 
+/**
+  * The permission amount of a range clause, i.e. the `p` in `range exp, p`. It is a
+  * specification-only annotation, and thus ghost, even though the enclosing for statement
+  * and the range expression itself need not be. A `PWildcardPerm` amount stands for the
+  * `_` written by the user.
+  */
+case class PRangePerm(exp: PExpression) extends PGhostMisc
```
jcp19: why do we introduce this intermediate step, rather than having perm in PRange of type `PExpression`?

src/test/resources/regressions/features/loops/range_maps-fail7.gobra
```diff
@@ -0,0 +1,33 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package pkg
+
+// The default amount is a fixed fraction, which is not known to be smaller than a
+// wildcard. This is the case a wildcard amount, `range x, _`, is meant for.
```
jcp19:
```suggestion
// The default amount is a fixed fraction, which is not known to be smaller than an
// existentially quantified permission amout.
// This is the case a wildcard amount, `range x, _`, is meant for.
```

src/main/scala/viper/gobra/ast/frontend/Ast.scala
jcp19: You're right, removed in 2dc4a0f — `perm` is now `Option[PExpression]`.

I introduced it to get the amount into a ghost context: `isEnclosingGhost` walks up the parent chain for a `PGhostNode`, and since `PRange` is a `PActualMisc` inside an actual for statement, the amount otherwise sits in no ghost context at all. But checking what that actually bought, it isn't load-bearing:

- **Purity** — the substantive check. `isPureExpr` works fine from `wellDefActualMisc`; it doesn't need a ghost context.
- **Ghost values in the amount** (`ghost p perm`, ghost pure calls) — allowed either way, because `PShortForRange`/`PAssForRange` are in the blanket `noMessages` case of `stmtGhostSeparation`.
- **Ghost erasure** — handled by the `GhostLessPrinter.showRange` override, independent of the node.

The only thing lost is a second diagnostic on `range m, impure()`: it used to report both "expected pure expression" and "call to non-ghost impure function in ghost code", and now reports only the first. The redundant one is no loss.

And as you imply, this matches how the `with` clause — the range clause's other specification-only annotation — is already modelled, as a plain `PUnkLikeId` on `PRange`.

The checks now live in `wellDefActualMisc` next to the "only when ranging over a map" check, and the `PRangePerm` cases in `wellDefGhostMisc`/`ghostMiscType`/`showMisc` are gone. Unit tests 1127/1127, `features` regressions 947 passing (the two `backends/*carbon*` failures are this machine having no Boogie).

---

src/test/resources/regressions/features/loops/range_maps-fail7.gobra
jcp19: Applied in 2dc4a0f (with `amout` → `amount`). "Existentially quantified permission amount" is the more accurate framing — thanks.

---

ArquintL:
Since the permission amount is optional, what difference does `_` make? It seems no permission amount and `_` both result in the default, fixed amount

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1397,12 +1397,50 @@ object Desugar extends LazyLogging {
...
+        * same amount is exhaled before and inhaled after the loop body, and is checked to be
+        * strictly positive.
+        *
+        * The exception is a wildcard amount, `range exp, _`, which is not stored: as in an `acc`
+        * expression, every occurrence of a wildcard denotes a fresh positive amount that is small
+        * enough to be available. In particular, the amount inhaled after the body need not be the
+        * one exhaled before it, so a wildcard is only useful for loops whose invariant holds a
+        * wildcard amount itself. There is nothing to check, since wildcards are positive by
+        * construction (and cannot be compared to other amounts).
+        */
```
ArquintL, line 1417: this is very verbose and could be condensed. There's no need to reiterate on the syntax or why wildcard amounts are not stored

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1411,6 +1449,8 @@ object Desugar extends LazyLogging {
         * name declared outside. This is also the go behaviour.
         *
         * c := x
+        * p := per
+        * assert none < p // check if permission provided by user is valid; omitted if per is absent
```
ArquintL, line 1453: wouldn't it be sufficient to declare `p` and assert the condition within the else branch?

src/main/scala/viper/gobra/frontend/Desugar.scala
```diff
@@ -1397,12 +1397,50 @@ object Desugar extends LazyLogging {
...
+        * strictly positive.
+        *
+        * The exception is a wildcard amount, `range exp, _`, which is not stored: as in an `acc`
+        * expression, every occurrence of a wildcard denotes a fresh positive amount that is small
+        * enough to be available. In particular, the amount inhaled after the body need not be the
+        * one exhaled before it, so a wildcard is only useful for loops whose invariant holds a
+        * wildcard amount itself. There is nothing to check, since wildcards are positive by
+        * construction (and cannot be compared to other amounts).
+        */
+      def rangePermD(range: PRange, permVar: in.LocalVar)(src: Source.Parser.Info): Writer[(Vector[in.Stmt], in.Expr)] =
```
ArquintL, line 1418: Why is `permVar` a parameter? Couldn't the fresh variable be created in this function?

src/test/resources/regressions/features/loops/range_maps-fail5.gobra
```diff
@@ -0,0 +1,18 @@
+// Any copyright is dedicated to the Public Domain.
+// http://creativecommons.org/publicdomain/zero/1.0/
+
+package pkg
+
+// The permission amount provided by the user is the amount that is exhaled, so it must be
+// available when entering the loop.
+
+requires acc(x, 1/2)
+ensures acc(x, 1/2)
```
ArquintL, line 10: This makes the test slightly more interesting

```suggestion
requires acc(x)
ensures acc(x)
```

src/test/resources/regressions/features/loops/range_maps2.gobra
```diff
@@ -0,0 +1,114 @@
...
+
+requires acc(x)
+ensures acc(x)
+decreases
+func exhalesHalf(x map[int]int) {
+	invariant acc(x)
+	decreases len(domain(x)) - len(visited)
+	for k := range x, 1/2 with visited {
+		// exactly 1/2 was exhaled, so the loop body still holds the other half
+		assert acc(x, 1/2)
```
ArquintL, line 35: This makes me wonder whether it wouldn't be more natural to exhale / inhale `writePerm - p` where p is the amount provided by the user. I feel like exhaling `p` before the body leaks quite a bit of the encoding to the user if the user wants to understand how to use `range` correctly

src/test/resources/regressions/features/loops/range_maps2.gobra
```diff
@@ -0,0 +1,114 @@
...
+	decreases len(domain(x)) - len(visited)
+	for _ = range x, 1/4 with visited {
+	}
+
+	var k int
+	var v int
+	invariant acc(x)
+	decreases len(domain(x)) - len(visited)
+	for k, v = range x, perm(1/4) with visited {
+		assert k elem domain(x)
```
ArquintL, line 56: Related to my comment above:
```suggestion
		assert acc(x, 3/4)
		assert k elem domain(x)
```

src/main/scala/viper/gobra/ast/frontend/Ast.scala
```diff
@@ -842,7 +842,13 @@ sealed trait PMisc extends PNode
...
 sealed trait PActualMisc extends PMisc
 
-case class PRange(exp: PExpression, enumerated: PUnkLikeId) extends PActualMisc
+/**
+  * Range clause of a for statement, i.e. `range exp(, perm)?( with enumerated)?`.
+  * `perm` is the (optional) amount of permission to `exp` that is exhaled while iterating over it.
+  * It is only supported for maps, where it defaults to a very small, but positive, amount.
+  * As in an `acc` expression, the blank identifier denotes a wildcard amount.
+  */
+case class PRange(exp: PExpression, perm: Option[PExpression], enumerated: PUnkLikeId) extends PActualMisc
```
ArquintL, line 851: why is `enumerated` not option? It feels a bit odd that we map an omitted `with` expression to `PWildcard`

jcp19:
> Since the permission amount is optional, what difference does `_` make? It seems no permission amount and `_` both result in the default, fixed amount

`_` is a wildcard amount, by default we get a known small permission amount. Therefore, if you have a wildcard permission to a map, you can never iterate through it so far

